Issue #3, opened 2015-01-13
Add capability to setup lvm thin pool and pass it to docker


rhvgoyal:
Right now we setup a data lvm and metadata lvm and pass it to docker and docker in turn sets up thin pool using libdevmapper library. Now lvm tools support creation of thin pool and one can create a 
thin pool using lvm and pass it to docker. 

Advantage of this approach is that lvm can do all the maintenace of pool like automatic resizing and
doing metadata checks when pool is activated etc.

So this should turn out to be a better approach.

cgwalters:
This was merged.

PR #4, opened 2015-01-13
Lvm thin pool setup

rhvgoyal:
This pull request contains changes for support of setting up lvm thin pool for docker consumption.

I have opened following issue.

https://github.com/projectatomic/docker-storage-setup/issues/3

This is not default right now and one will have to set SETUP_LVM_THIN_POOL=yes in /etc/sysconfig/docker-storage-setup for new code to kick in. 

Please review and pull.

### DIFF
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -26,6 +26,13 @@ VG:   The volume group to use for docker storage.  Defaults to the
       created (which requires that "DEVS" be nonempty, since we don't
       currently support putting a second partition on the root disk).
 
+SETUP_LVM_THIN_POOL:
+      If set to "yes", then an LVM thin pool will be setup which
+      is passed to docker.
+
+      lvm2 version should be same or higher than lvm2-2.02.112 for lvm
+      thin pool functionality to work properly.
+
 The options below should be specified as values acceptable to 'lvextend -L':
 
 ROOT_SIZE: The size to which the root filesystem should be grown.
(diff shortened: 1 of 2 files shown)


cgwalters:
A next step here for Atomic/F22 is to default to this out of the box - it might be nice if we could auto-detect whether we had new enough versions of lvm2/docker in the script, but failing that, we can set the variable by default in the installer.

rhvgoyal:
Agreed. I was hoping that let this get merged first and if there are no issues, then work on making this new option default.

rhvgoyal:
cgwalters, can you please pull in this request.

rhvgoyal:
Found a small bug in my patches. Will fix it and update the pull request soon.

rhvgoyal:
Ok, I pushed two more commits. First commit is to fix a bug in previous patches. And second commit makes lvm pool mode as default mode and by default setups lvm pool to pass to docker.

rhvgoyal:
@cgwalters 

Please have a look and pull the changes.

rhvgoyal:
I ran the script from / and I don't see it expanding to /tmp. Things still work fine.

Instead putting a double quote or single quote around it breaks the things. I want regular express to work and putting a quote around it might break it.

cgwalters:
Ok, I understand `[[` better now from reading the manual.  :+1: from me, merging.  Thanks so much for a well written and split patch series!

PR #8, opened 2015-04-17
Update docker-storage-setup.spec

detiber:
xfsprogs is required, so update the spec.


cgwalters:
http://pkgs.fedoraproject.org/cgit/docker-storage-setup.git/commit/?id=159d12e1dc4cf65a89ed4b5b6d11eb90414dacf4

Issue #13, opened 2015-04-27
Do not auto extend data lv. Let user use lvm facilities for this as per the need


rhvgoyal:
Right now docker-storage-setup keeps on extending the size of data volume
as long as there is free space in the pool. So even if a user has left
some space in the pool free for metadata volume extension, that free space
will be consumed by data volume extension.

Now lvm2 has nice capabilities to automatically extend pool (both data and
metadata) when pool is running short on space. Following are two knobs which
one can use for auto extension.

thin_pool_autoextend_threshold = 100
thin_pool_autoextend_percent = 20

By default extension is disabled. One needs to enable it manually by chaning
the value of thin_pool_autoextend_threshold.

In general it is a good idea to know consume all the free pool space until
and use mechanism like extending pool by certain percentage when it hits
threshold. That allows free space in volume group to be shared by other
entities like root which might have to be grown on need basis too.

elephantfries:
@rhvgoyal  I ran into this indirectly with the latest Centos 7 Atomic installation via kickstart. Anaconda fails on this line:

logvol none --thin --size=100000 --poolname=docker-pool --name=docker-pool --vgname=centos

It complains about 'mount point none is invalid'.  This is probably anaconda problem but leaves me with only implicit configuration which then consumes the entire remaining disk space.  The same logvol configuration works fine on Fedora 22 Atomic.

rhvgoyal:
@elephantfries 

I think issue you are facing has nothing to do with this commit. 

You seem to be trying to create a thin volume using anaconda kickstart commands. And I don't know why that is not working.

In fedora atomic, I think we might not be taking this appraoch. We leave free space in root volume group and later during boot, docker-storage-setup sets up the thin pool. Anaconda does not even come into the picture to setup the docker thin pool.

rhvgoyal:
Patches merged. Closing.

PR #14, opened 2015-04-27
docker-storage-setup: Disable auto extension of data logical volume

rhvgoyal:
Fixes issue #13 

Right now docker-storage-setup keeps on extending the size of data volume
as long as there is free space in the pool. So even if a user has left
some space in the pool free for metadata volume extension, that free space
will be consumed by data volume extension.

Now lvm2 has nice capabilities to automatically extend pool (both data and
metadata) when pool is running short on space. Following are two knobs which
one can use for auto extension.

thin_pool_autoextend_threshold = 100
thin_pool_autoextend_percent = 20

By default extension is disabled. One needs to enable it manually by chaning
the value of thin_pool_autoextend_threshold.

In general it is a good idea to know consume all the free pool space until
and use mechanism like extending pool by certain percentage when it hits
threshold. That allows free space in volume group to be shared by other
entities like root which might have to be grown on need basis too.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -364,8 +364,6 @@ if [ -n "$DATA_LV_SIZE" ]; then
     else
       lvextend -L $DATA_SIZE $VG/$DATA_LV_NAME || true
     fi
-  else
-    lvextend -l "+$DEFAULT_DATA_SIZE_PERCENT%FREE" $VG/$DATA_LV_NAME || true
   fi
 elif [ -n "$DATA_SIZE" ]; then
   # TODO: Error handling when DATA_SIZE > available space.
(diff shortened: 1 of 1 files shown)


rhvgoyal:
Added one more commit to pulll request. We also have some logic to auto extend metadata volume size. Disable that too and leave it to lvm2 the auto extension part.

rhvgoyal:
@cgwalters 

It will be. I am working on a separate pull request for fixing 12. Just that these were important behavior changes which applied to even thin pool mode. So I did not want to club these changes with cleanup part. Instead want to highlight this change and keep in a separate small commit.

Do let me know if you don't like it this way.

rhvgoyal:
IOW, it would not be dead by default even after cleanup as this change applied to thin pool too. So this is more than removing support for old mode.

cgwalters:
Either way is fine by me!  Just checking.

Ok and on the first commit - I just sanity checked that this will continue working for the "first boot" path, because we're still creating the initial pool LV.

So :+1: from me, thanks!

Issue #12, opened 2015-04-24
Delete non-pool code paths


cgwalters:
Since https://github.com/docker/docker/pull/12404 we should just drop the non-pool code paths for distribution targets which don't need them, as users should not use it.

PR #15, opened 2015-04-27
Remove old data and metadata volume mode support

rhvgoyal:
Fixes issue #12 

Docker does not do all the error handing which can be required if thin
 pool gets full or thin pool runs into issues. So passing block devices
 to docker and let it setup thin pool is not a very good idea from error
 handling point of view.
    
 Instead always setup lvm thin pool and pass that to docker so that docker
 does not have to take care of pool management issues and lvm2 commands
 can take care of it. (See man lvmthin).

### DIFF
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -1,8 +1,7 @@
 .TH "DOCKER-STORAGE-SETUP" "1" "NOVEMBER 2014" "Helper Script for Docker Storage Setup" ""
 .SH NAME
 .PP
-docker\-storage\-setup - Grows the root filesystem and sets up LVM volumes for docker 
-metadata and data.
+docker\-storage\-setup - Grows the root filesystem and sets up a LVM thin pool for docker.
 .SH SYNOPSIS
 .PP
 \f[B]docker-storage-setup\f[] 
@@ -12,6 +11,9 @@ None.
 .SH EXAMPLES
 Run \f[B]docker-storage-setup\f[] after setting up your configuration in 
 /etc/sysconfig/docker-storage-setup. 
+
+lvm2 version should be same or higher than lvm2-2.02.112 for lvm thin pool
+functionality to work properly.
 
 \f[B]Supported options for the configuration file\f[]:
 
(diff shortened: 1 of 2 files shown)


rhvgoyal:
@cgwalters 

I don't have setup to run cloud images. I have done testing with normal volume groups. Can you please run this once with cloud images and make sure things are still working.

cgwalters:
Re `DATA_LV_SIZE` - got it, thanks!  I'll give this patch a quick test now.

cgwalters:
Current patch works for me, :+1:

cgwalters:
I'll write down how I tested this quickly, as it may be useful for you or someone else.

First, grab a copy of a cloud VM qcow2 file.  Here we'll call it `foo.qcow2`, but in the case of Fedora it's e.g. `Fedora-Cloud-Atomic-22_Beta-20150407.x86_64.qcow2`.

```
guestfish -a foo.qcow2
run
mount /dev/atomicos /
copy-in /home/walters/src/github/projectatomic/docker-storage-setup/docker-storage-setup.sh.in /ostree/deploy/$osname/deploy/$checksum/usr/bin
mv /ostree/deploy/$osname/deploy/$checksum/usr/bin/docker-storage-setup.sh /ostree/deploy/$osname/deploy/$checksum/usr/bin/docker-storage-setup
sync
(Control-D)
```

Then import that modified qcow2 to a new VM.

You'll need to replace the `$osname` and `$checksum` variables, or just rely on tab completion.  Also having to do `copy-in` plus `mv` is ugly but I didn't find an easier way with libguestfs offhand.

rhvgoyal:
@cgwalters 

Pushed patches again after taking care of review comments. Basically modified the man page and removed the [ $? -ne 0 ] checks. 

Now there is only one place I am still using [ $? -ne 0 ] to capture grep's return code to figure out if we are using old mode or not. And that still works. Looks like bash knows that in case of grep 1, is not an error and it does not exit.

rhvgoyal:
ok, pushed second patch again. This time put the grep in an if statement.

Issue #16, opened 2015-04-28
do not overwrite /etc/sysconfig/docker-storage always


rhvgoyal:
Right now upon each run of script, we overwrite /etc/sysconfig/docker-storage
and this could be bad if user added some of his own storage options. These
will be lost over next run of script.

Write /etc/sysconfig/docker-storage only when pool is created. After that
leave it untouched.

rhvgoyal:
Patch merged. Closing.

PR #17, opened 2015-04-28
docker-storage-setup: Do not overwrite /etc/sysconfig/docker-storage always

rhvgoyal:
This fixes issue #16 

Right now upon each run of script, we overwrite /etc/sysconfig/docker-storage
and this could be bad if user added some of his own storage options. These
will be lost over next run of script.

Write /etc/sysconfig/docker-storage only when pool is created. After that
leave it untouched.


cgwalters:
:+1: 

While reviewing this, I noticed https://github.com/projectatomic/docker-storage-setup/pull/20 - want to review?

Issue #18, opened 2015-04-28
ship a default /etc/sysconfig/doceker-storage-setup file


rhvgoyal:
I am thinking that it might be a good idea to ship a default /etc/sysconfig/docker-storage-setup file which documents supported options. That way it becomes easier for user to find and modify options, especially if man pages are not around.

rhvgoyal:
Patch merged. Closing.

PR #19, opened 2015-04-28
docker-storage-setup: Create a default config file to be shipped with pkg

rhvgoyal:
Fixes issue #18 

I think sending a documented default config file with the package makes it
easy to configure and makes it easy to more config options down the line,
like chunk size of thin pool.

### DIFF
--- a/docker-storage-setup.conf
+++ b/docker-storage-setup.conf
@@ -1,0 +1,25 @@
+# A quoted, space-separated list of devices to be used.  This currently
+# expects the devices to be unpartitioned drives.  If "VG" is not specified,
+# then use of the root disk's extra space is implied.
+#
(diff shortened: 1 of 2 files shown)


rhvgoyal:
I have not been able to do any testing with spec file. Not sure how this spec file is to be tested.

cgwalters:
`docker-storage-setup.conf` is cool with me.

rhvgoyal:
@cgwalters 

Ok, pushed the patch again after taking care of feedback. Please have another look.

cgwalters:
Merging this, but note that no automatic process synchronizes this spec file with the copies in downstream distributions.

rhvgoyal:
ok, I will create separate bug downstream to synchronize the spec file.

PR #20, opened 2015-04-28
Atomically update config file

cgwalters:
Just noticed this while reviewing another PR; previously if we got
e.g. ENOSPC or something on startup, we'd leave a broken partially
written file.  Do the tempfile + `rename()` dance.


cgwalters:
Ah... `/usr/bin/mv` calls out to the `rename(2)` system call, but `/usr/bin/rename` is a totally different thing...

rhvgoyal:
ah, got it. I did not know that. thanks. This patch looks good to me.

Issue #21, opened 2015-04-28
disable auto data volume extension with DATA_SIZE is specified


rhvgoyal:
We are still doing auto data volume extension when user specifies DATA_SIZE. Say DATA_SIZE=+10%FREE, then every reboot we will extend data volume by 10% of free space.

Given that we are moving towards relying on lvm to extend data volume for pool based on threshold, not sure how many people need this functionality.

There are two ways to handle it.

- Get rid of auto extension for pool data volume completely. Rely on lvm. This might be little surprise over upgrade for an existing DATA_SIZE user. Having said that we don't publicize in man page that DATA_SIZE can take arguments which suit to "lvmextend -l". We have maintained that pass in arguments suitable to "lvextend -L". So chances of somebody relying on auto data extension on every reboot are slim.

- Retain auto growth capability but make it conditional. Say AUTO_DATA_LV_GROW=1. By default turn it off in /etc/sysconfig/docker-storage-setup. For users upgrading d-s-s, they will have turned it on by default until they pass in  AUTO_DATA_LV_GROW=0.

Right now I am planning to send patches for first approach. If that's not acceptable, I can work on second approach.

rhvgoyal:
Patch merged. Closing issue.

PR #22, opened 2015-04-28
Disable auto data volume extension when DATA_SIZE is specified

rhvgoyal:
Fixes issue #21   

  If DATA_SIZE is specified we will auto grow data volume if there is free
    space in pool. Say DATA_SIZE=+80%FREE, then we will automatically consume
    80% of free space in volume group upon restart of service.
    
    If user did not specify DATA_SIZE, then by default script determines data
    volume size to be 98% of free space in volume group. And we got rid of that
    logic to auto extend. This is very similar, difference is that was default
    and this is user specified.
    
    Now if we are moving towrads relying on lvm to auto grow pool based on
    /etc/lvm/lvm.conf settings, then we don't require this.
    
    Also it will make life little easier in some cases where this behavior is
    unexected. For example, assume root volume group is 500G in size and I
    want to use only 50% of free space then one can speicfy DATA_SIZE=50%FREE.
    Now one does not expect that if later volume group grows then we will
    extend data volume to make sure it meets 50% of FREE at that time.
    
    So IOW, there could be use cases for both. I am inclined towards not growing
    data volume automatically as lvm2 already has mechanism to do that. So we
    don't have to replicate that thing in docker-storage-setup. If lvm2 does
    not work well for some reason, then we can revisit introducing auto growth
    of data volume based on some config optoin, say AUTO_GROW=1.
    
    For now, get rid of auto growth of data volume logic. This means that
    DATA_SIZE value will take effect only during pool creation time. If pool
    has been created and user changes DATA_SIZE, it will not have any effect
    when service restarts. So this might be a little behavior change for
    existing users of DATA_SIZE. They will have to rely on lvm2 for auto
    data volume extension.

### DIFF
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -36,8 +36,15 @@ The options below should be specified as values acceptable to 'lvextend -L':
 ROOT_SIZE: The size to which the root filesystem should be grown.
 
 DATA_SIZE: The desired size for the docker data LV.  Defaults to using
-           all free space in the VG after the root LV and docker
+           98% free space in the VG after the root LV and docker
            metadata LV have been allocated/grown.
+
+           DATA_SIZE can take values acceptable to "lvcreate -L" as
+           well as some values acceptable to to "lvcreate -l". If user
+           intends to pass values acceptable to "lvcreate -l", then
+           only those values which contains "%" in syntax are
+           acceptable.  If value does not contain "%" it is assumed
+           value is suitable for "lvcreate -L".
 
 
 \f[B]Sample\f[]
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -115,26 +115,10 @@ create_lvm_thin_pool () {
   lvconvert -y --zero n --thinpool $VG/$DATA_LV_NAME --poolmetadata $VG/$META_LV_NAME
 }
 
-# If user specified DATA_SIZE and current pool is smaller, extend it.
-extend_data_lv () {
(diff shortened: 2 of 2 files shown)


docker-storage-setup.1
```diff
@@ -36,9 +36,16 @@ The options below should be specified as values acceptable to 'lvextend -L':
 ROOT_SIZE: The size to which the root filesystem should be grown.
 
 DATA_SIZE: The desired size for the docker data LV.  Defaults to using
-           all free space in the VG after the root LV and docker
+           98% free space in the VG after the root LV and docker
            metadata LV have been allocated/grown.
 
+           DATA_SIZE can take values acceptable to "lvextend -L" as
```
swagiaal: Just a nit but should we say "lvcreate -L" here since the patch is getting rid of the use of lvextend

swagiaal:
One nit; otherwise LGTM

docker-storage-setup.1
rhvgoyal: Ok, makes sense. I will do this change and push the patches again.

rhvgoyal:
@swagiaal Ok, I made the changes and pushed the patches again. Please have a look again.

swagiaal:
LGTM. @cgwalters could you merge this ? I don't have the permissions.

cgwalters:
Merged, thanks!

cgwalters:
I invited both of you to the projectatomic github org - let me know if you didn't get it or it doesn't work.

PR #23, opened 2015-05-18
proposal to update docker-storage-setup.sh to accept option to configure different chunksize for thin lvm

ekuric:
docker-storage-setup.sh will use by default 64K chunksize when creating thin lvm, with this proposal it will be possible to add 
CHUNKSIZE=$value 
in /etc/sysconfig/docker-storage-setup 
and crate thin lvm with desired (supported) value for chunksize


cgwalters:
I don't understand the "delete/added" commits.  This seems like it should be one commit adding the config file option.

rhvgoyal:
agreed. This has to be one simple commit which provides user the option to specify chunk size.

Right now, chunk size is determined by lvm2. So if lvm2 changes their defaults, it will change. I guess it is a good idea to leave it like that, until and unless we know better.

So by default we can leave it to lvm2 to decide chunk size. And user can override that by specifying
a chunk size in CHUNK_SIZE= in /etc/sysconfig/docker-storage-setup.

ekuric:
I think @rhvgoyal patched this already. Thank you.

PR #25, opened 2015-05-18
Disable growpart logic by default

### LINKED ISSUE #24: cleanup related to growpart
docker-storage-setup script does multiple things. Not only it sets up an lvm thin pool, but it can create partitions on disk and also automatically grow root pvs upon reboot. This is highly custom for project atomic. 

It is desirable to be able to use this service outside project atomic too in a more generic settings where auto growing is not required and growpart might not even be available.

Write a patch series which does few things.

- Do not make a dependency on growpart. Use it if available.
- Disable growpart logic by default and provide a knob to enable it through config file.

rhvgoyal:
Fixes #24 

This is generic cleanup related to growpart utility and logic around it. I think for project atomic one shall have to take care of two things.

- Make sure that cloud-utils-growpart gets installed. Installing docker-storage-setup will not pull it in automatically.

- Make sure to modify GROWPART= variable in /etc/sysconfig/docker-storage-setup to enable automatic growing of root pvs each reboot.

### DIFF
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -35,6 +35,12 @@ VG:   The volume group to use for docker storage.  Defaults to the
       lvm2 version should be same or higher than lvm2-2.02.112 for lvm
       thin pool functionality to work properly.
 
+GROWPART:
+      One can use this option to enable/disable growing of partition
+      table backing root volume group. This is intended for virtualization
+      and cloud installations. By default it is disabled. Use GROWPART=true
+      to enable automatic partition table resizing.
+
 The options below should be specified as values acceptable to 'lvextend -L':
 
 ROOT_SIZE: The size to which the root filesystem should be grown.
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -10,7 +10,11 @@ docker\-storage\-setup - Grows the root filesystem and sets up a LVM thin pool f
 None. 
 .SH EXAMPLES
 Run \f[B]docker-storage-setup\f[] after setting up your configuration in 
-/etc/sysconfig/docker-storage-setup. 
+/etc/sysconfig/docker-storage-setup. One can look at
+/usr/lib/docker-storage-setup/docker-storage-setup for various options and
+their default settings. Anything user wants to change, should be changed
+in /etc/sysconfig/docker-storage-setup. This is the file which will
+override any settings specified in /usr/lib/docker-storage-setup/docker-storage-setup.
 
 lvm2 version should be same or higher than lvm2-2.02.112 for lvm thin pool
 functionality to work properly.
--- a/docker-storage-setup.conf
+++ b/docker-storage-setup.conf
@@ -23,3 +23,8 @@
 # Value should be acceptable to -L option of lvextend.
 #
 # DATA_SIZE=8G
+
+# Enable resizing partition table backing root volume group. By default it
+# is disabled until and unless GROWPART=true is specified.
+#
+GROWPART=false
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -158,6 +158,8 @@ is_old_data_meta_mode() {
 }
 
 grow_root_pvs() {
+  [ -x "/usr/bin/growpart" ] || return
+
   # Note that growpart is only variable here because we may someday support
   # using separate partitions on the same disk.  Today we fail early in that
   # case.  Also note that the way we are doing this, it should support LVM
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -171,6 +171,46 @@ grow_root_pvs() {
   done
 }
 
+create_disk_partitions() {
+  for dev in $DEVS; do
+    if expr match $dev ".*[0-9]"; then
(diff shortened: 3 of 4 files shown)


cgwalters:
There are some minor typos in the commit messages "Current1" and "little clear" -> "a little clearer".   Otherwise LGTM :+1:

rhvgoyal:
sure I will take care of typos as well as make message about GROWPART more clear. In fact I am thinking that by default I can leave GROWPART logic on. So existing users will not be impacted over upgrade.

I will put GROWPART=false in docker-storage-setup.conf so that new users will get it off by default.

rhvgoyal:
@cgwalters I cleaned up patches based on your feedback and pushed these again. Now growpart logic is on by default and one need to use GROWPART=false to disable it. I have set GROWPART=false in docker-storage-setup.conf so that new users get it off by default and existing users will have it on over upgrade.

cgwalters:
I think what we discussed is that for the Atomic cloud images, we'd leave growpart off in the code, and I'd turn it on.  Now, if I modify d-s-s.conf in the kickstart file, it will mean that file will count as modified for OSTree upgrades, and users will never see a newer version.

I know this sounds like a bit of a hack, but given what we want is growpart enabled only in cloud images, how about creating `/etc/sysconfig/docker-storage-setup.system.conf` which is sourced before `/etc/sysconfig/docker-storage-setup.conf`?  Then I `echo GROWPART=true > /etc/sysconfig/docker-storage-setup.system.conf` in the kickstart file.

(Really it'd be nice to have a level of config infrastructure equivalent to systemd with drop-in files and the like...)

rhvgoyal:
@cgwalters 

How about creating a file in /lib/docker-storage-setup/docker-storage-setup. This will be sourced before the one in /etc/sysconfig/docker-storage-setup. And user is not supposed to touch the one present in /lib. For some reason I have the impression that packages are supposed to put config
file in /lib/ which can be overridden with the one in /etc/.

rhvgoyal:
That way you can continue to modify /etc/sysconfig/docker-storage-setup. And any new options over upgrade will show up in /lib/docker-storage-setup/docker-storage-setup but your /etc/sysconfig/docker-storage will not be overwritten. 

That will allow you to override default behavior and maintain that behavior over upgrade while getting new defaults for new features at the same time.

cgwalters:
In a regular RPM model that would work, but unfortunately for Atomic we are using *one* tree for both bare metal (where we don't want growpart) and cloud (where we do).  So `/usr` content has to be the same, but we can tweak configuration in `/etc`.

rhvgoyal:
@cgwalters 

Sorry I did not get above comment. /usr content will be same both in baremetal and cloud. And some script will tweak /etc/sysconfig/docker-storage-setup only for cloud image. I am not sure why it will not work (based on what you just described).

cgwalters:
Ok yep.  I think the minor downside of this is that having the defaults be in e.g. `/usr/lib/docker-storage-setup/docker-storage-setup.conf` will be a bit undiscoverable.  Why not just document the defaults in the man page, and have them be hardcoded in the script itself?

rhvgoyal:
I personally find it much more readable and easily usable when defaults and various options are 
in a file. One can simply copy that template file and work with it much more easily. Another advantage it feels much more like an API. Otherwise inside a script, it is another global variable and can easily mix up with general global varibales which are not exported.

So personally I like the idea of shipping a file in /usr/lib/ and dracut and systemd already seems to be using it.

To address the concern of not being discoverable, we can document it in man page so that one can
find it. In fact, it does not have to be discoverable if user knows what to write in /etc/sysconfig/docker-storage-setup. We don't want user to modify /usr/lib/. This is just more of a help, and if user does not find it, its ok. No harm done. It is no worse then not documenting these defaults at all and hardcoding them in the script.

rhvgoyal:
@cgwalters I pushed one more time. Have a look. 

Creating a well documented config file has helped me in the past and I like the idea, that's the reason I have been pushing for it. But if you don't like the idea, I can change it. No issues.

rhvgoyal:
I just checked my atomic virt machine and there is no "man" command. So keeping supported
options in a config file is going to help a bit as there is no tool to read man page.

cgwalters:
Regarding man pages, `atomic run rhel7/rhel-tools` will get you ones for the `atomic` command for the same reason.  We can also add `docker-storage-setup` to the tools container.

cgwalters:
So :+1: from me aside from the `%{config}` for files in `%{_libdir}` issue.

rhvgoyal:
@cgwalters 

So If just get rid of %config, is that good enough? dracut seems to be doing that.

install -m 0644 dracut.conf.d/fedora.conf.example $RPM_BUILD_ROOT%{dracutlibdir}/dracut.conf.d/01-dist.conf
%{dracutlibdir}/dracut.conf.d/01-dist.conf

W.r.t running rhel-tools container, this seems like chicken and egg situation. Before we can run containers we need to setup thin-pool and we need access to docker-storage-setup man page.

rhvgoyal:
pushed again. Got rid of %{config} option from spec file.

cgwalters:
Thanks!  :+1: from me.

Issue #26, opened 2015-05-20
Enforce devicemapper graphdriver in generated /etc/sysconfig/docker-storage file


rhvgoyal:
Right now docker-storage-setup creates storage options which are valid
    only for devicemapper driver. If some other driver is trying to initialize
    and it is not properly written it will return error saying unrecognized
    options dm.fs etc.
    
    I am running into this issue now. Recently upstream has added zfs as
    graphdriver and that takes precedence over devicemapper. And that driver
    is returning error that dm.fs is unrecognized option.
    
    Given that docker-storage-setup is expecting docker to use devicemapper
    graphdriver and generating options which only work with devicemapper,
    it makes sense to enforce the driver to use here to avoid errors.

rhvgoyal:
Patch merged. Closing.

PR #27, opened 2015-05-20
docker-storage-setup: Enforce storage driver to be devicemapper

rhvgoyal:
Fixes issue #26 

Right now docker-storage-setup creates storage options which are valid
only for devicemapper driver. If some other driver is trying to initialize
and it is not properly written it will return error saying unrecognized
options dm.fs etc.

I am running into this issue now. Recently upstream has added zfs as
graphdriver and that takes precedence over devicemapper. And that driver
is returning error that dm.fs is unrecognized option.

Given that docker-storage-setup is expecting docker to use devicemapper
graphdriver and generating options which only work with devicemapper,
it makes sense to enforce the driver to use here to avoid errors.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -76,7 +76,7 @@ write_storage_config_file () {
     fi
     done )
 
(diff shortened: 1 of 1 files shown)


cgwalters:
Initiating press of big green button in 3...2...1...

PR #33, opened 2015-05-21
docker-storage-setup: Reserve 60% of free space for data volume

### LINKED ISSUE #32: Reserve 60% of free space by default for data volume
Right now by default we reserve 98% of free space as data volume. Now we
    have enabled automatic pool extension logic. That will make sure that pool
    data and metadata volumes will grow automatically as they start filling up.
    
    This means we don't have to consume all the free space available at startup.
    It can be consumed dynamically. One advantage of not consuming all the space
    up front is that it can be used for some other purpose. For example, one
    can try to grow root volume. 3G of root might not be sufficient.
    
    Leaving free space also allows for extension of metadata volume later easily.
    Some openshift folks complained that not enough metadata size is being
    reserved at startup time. Leaving space free and growing volumes dynamically
    gives us lot of flexibility here for different kind of workloads.

rhvgoyal:
Fixes #32 

Right now by default we reserve 98% of free space as data volume. Now we
have enabled automatic pool extension logic. That will make sure that pool
data and metadata volumes will grow automatically as they start filling up.

This means we don't have to consume all the free space available at startup.
It can be consumed dynamically. One advantage of not consuming all the space
up front is that it can be used for some other purpose. For example, one
can try to grow root volume. 3G of root might not be sufficient.

Leaving free space also allows for extension of metadata volume later easily.
Some openshift folks complained that not enough metadata size is being
reserved at startup time. Leaving space free and growing volumes dynamically
gives us lot of flexibility here for different kind of workloads.

### DIFF
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -61,7 +61,7 @@ The options below should be specified as values acceptable to 'lvextend -L':
 ROOT_SIZE: The size to which the root filesystem should be grown.
 
 DATA_SIZE: The desired size for the docker data LV.  Defaults to using
-           98% free space in the VG after the root LV and docker
+           60% free space in the VG after the root LV and docker
            metadata LV have been allocated/grown.
 
            DATA_SIZE can take values acceptable to "lvcreate -L" as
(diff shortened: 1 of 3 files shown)


cgwalters:
LGTM :+1: 

Also, this will dovetail nicely with plans I have to use thinp for the OS LV.

PR #34, opened 2015-05-21
docker-storage-setup: Use chunk size 512K by default

### LINKED ISSUE #31: Adjust default chunksize on thinp volume created by docker-storage-setup
Although we expect most persistent I/O to be written to volumes attached to a container, we want the default docker graph storage on RHEL systems to be as performant as possible.

smallfile, random I/O performance improves when increasing thinp chunksize:

image

Increased chunksize also decreases metadata space usage for smallfile, random I/O:
image

The same is not the case for sequential read/write workloads:
image

The trade-off for reduced metadata space is increased space on the data thinp device:

image

The reason for this increased data usage under random I/O workloads is documented here:
http://people.redhat.com/msnitzer/2014-01-21-random-io-and-copy-on-write-schemes.markdown.html

rhvgoyal:
Fixes #31 and #23 

This patch does two things.

- Provides a knob CHUNK_SIZE to allow controlling chunk size of thin pool.
- Switches to default chunk size of 512K instead of 64K default of lvm.

Setting a chunk size is very tricky. larger chunk size can help in reducing
metadata usage and fragmentation. But at the same time consume more disk
space.

This chunk size has been arrived based on jeremy's testing and recommendation
from thin pool developers.

In case this does not work well, we can always change default.
 
### DIFF
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -56,6 +56,10 @@ POOL_AUTOEXTEND_PERCENT:
       that when threhold is hit, pool will be grown by 20% of existing
       pool size.
 
+CHUNK_SIZE:
+      Controls the chunk size/block size of thin pool. Value of CHUNK_SIZE
+      be suitable to be passed to --chunk-size option of lvconvert.
+
 The options below should be specified as values acceptable to 'lvextend -L':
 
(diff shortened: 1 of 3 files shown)


cgwalters:
:+1:

PR #37, opened 2015-06-05
Create temporary files in /tmp, not cwd

cgwalters:
Otherwise we end up polluting `/` on a mainline system, and on Atomic
it will fail.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -222,7 +222,7 @@ enable_auto_pool_extension() {
   local profileName="${volume_group}--${pool_volume}-extend"
   local profileFile="${profileName}.profile"
   local profileDir
-  local tmpFile=`mktemp tmp.XXXXX`
+  local tmpFile=`mktemp -t tmp.XXXXX`
 
   profileDir=$(lvm dumpconfig | grep "profile_dir" | cut -d "=" -f2 | sed 's/"//g')
   [ -n "$profileDir" ] || return 1


rhatdan:
Should we create these in /run , if this is a privileged process.

rhatdan:
My comment ignored?

rhatdan:
mktemp -p /run

cgwalters:
Sorry, I missed it.  I don't understand the argument for `/run` over `/tmp` wrt privileged processes.  Can you elaborate?

rhatdan:
I have been a big proponant over using /run with root privs required prevents unpriv processes attacks even if the names prove guessable and potentially prevents information leak.

I like to state root should always use /run and unpriv users should use /tmp as a rule of thumb

cgwalters:
RE: guessable; mktemp uses `open(...O_EXCL)`, so there's no races, right?  I mean `/tmp` has to be secure too

RE: information leak, similarly mktemp uses mode `0600`...

And finally an argument against `/run` is that there's no cleanup for it.  `/usr/lib/tmpfiles.d/tmp.conf` just has:

    d /tmp 1777 root root 10d
    d /var/tmp 1777 root root 30d

And it'd be fairly disastrous to add it because then we'd start cleaning up old pid files...

I guess you're maybe arguing for `/run/tmp/` ?

rhatdan:
Sure.  I just see all these things to worry about under /tmp, and just want developers to default to /run.

PR #40, opened 2015-06-08
docker-storage-setup: If metadata volume already exists, do not create new one

### LINKED ISSUE #39: if metadata lv already exists do not try to create a new one
I metadata lv already exists (from previous run), then do not try to create a new one as that will fail.

rhvgoyal:
Fixes #39 

It can happen that we created a metadata volume and later there was not
enough space to create data volume and script exited, leaving behind
metadata volume. Later if user adds more space to volume group and re-runs
the script, it fails saying metadata volume already exists.

For now fix this by checking if metadata volume exists or not. If it exists
use it (instead of erroring out).

I think ideally we should have cleanup logic and cleanup any logical
volumes created if error happens later. That probably also means
getting rid of "set -e".

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -78,6 +78,16 @@ EOF
 }
 
 create_metadata_lv() {
+  # If metadata lvm already exists (failures from previous run), then
+  # don't create it.
+  # TODO: Modify script to cleanup meta and data lvs if failure happened
+  # later. Don't exit with error leaving partially created lvs behind.
+
+  if lvs -a $VG/${META_LV_NAME} --noheadings &>/dev/null; then
+        echo "Metadata volume $META_LV_NAME already exists. Not creating a new one."
+	return 0
+  fi
+
   # Reserve 0.1% of the free space in the VG for docker metadata.
   # Calculating the based on actual data size might be better, but is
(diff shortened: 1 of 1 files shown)


rhvgoyal:
I think handling of subshells is little different. if checks for exit code. Here is the sample code.
#!/bin/bash

if i=$(exit 1); then
        echo "subshell exited with zero status"
        exit 0
fi

echo "subshell exited with non-zero status"

rhvgoyal:
My understanding is that one can not create a zero sized logical volume. So the very fact command exited with success status means, lv exists, that also means that size is greater than 0.

rhvgoyal:
@cgwalters does above answer your queries?

cgwalters:
Ok, right.  Your patch will likely work, but I find it confusing that we're retrieving the size and not doing anything with it.  If we're just trying to verify existence, how about:

```bash
if lvs "${VG}/${META_LV_NAME}" &>/dev/null; then
  return 0
fi
```

?

rhvgoyal:
Sounds good. I will modify my patch.

rhvgoyal:
@cgwalters 

Refereshed the patch as per your suggestion. Tested and it works. PTAL.

cgwalters:
:+1:

PR #43, opened 2015-06-15
docker-storage-setup: Enable deferred device removal feature if available

### LINKED ISSUE #42: Enable deferred device removal feature in docker devicemapper graph driver
Docker 1.7 devicemapper graph driver now supports deferred device removal feature. That is if device backing image/container is busy, then docker will not wait for all device refcounts to be dropped. Instead docker will schedule a deferred a deferred removal on device. That is when last reference to
device is dropped, kernel will automatically remove the device.

Enable this feature by default if docker version supports it.

rhvgoyal:
Fixes #42 

Enable deferred device removal feature by default if docker version is
equal or higher than 1.7.

This feature will allow stopping container even if device is busy. Once
the last reference to device is dropped, device will automatically go away
and docker does not have to wait for it.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -59,6 +59,34 @@ META_LV_NAME="${POOL_LV_NAME}meta"
 
 DOCKER_STORAGE="/etc/sysconfig/docker-storage"
 
+get_docker_version() {
+	local version
+
+	if ! version=$(docker version 2>/dev/null | grep "Client version" | cut -d ":" -f2 | sed 's/^ *//');then
+		return 1
+	fi
+	echo $version
+}
+
+get_deferred_removal_string() {
+	local version major minor
+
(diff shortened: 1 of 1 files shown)


rhvgoyal:
@cgwalters ping

cgwalters:
The code looks as good as shell script can ;)

rhatdan:
Lets make sure this DOES NOT get into docker-1.6.2 release.  @lsm5

rhvgoyal:
@rhatdan 

Code detects the version dynamically and enables this feature only if docker version is greater than or equal to 1.7. So even if it gets into 1.6.2 accidently, I think it should cause on harm.

lsm5:
@rhatdan @rhvgoyal no worries, the commit that went into 1.6.2 was #eefbef7

lsm5:
which is the one before this commit

PR #45, opened 2015-07-07
[WIP]: Don't do anything if Docker already has data on loopback

cgwalters:
(This patch is not tested, submitting to gather comments)

If we want this service to start by default, it doesn't make sense to
try to allocate thinpool space if there are already Docker images on
loopback.  People would then just see their images/containers vanish.

Let's not do anything unless the system appears to be a clean slate.

For example, we also now silently do nothing if the user has
explicitly set a backend with `-s`, such as `-s overlayfs`.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -194,6 +194,32 @@ is_old_data_meta_mode() {
   return 0
 }
 
+docker_storage_is_configured() {
+  if [ ! -f "$DOCKER_STORAGE" ];then
+    return 1
+  fi
+  if grep -q -e "^DOCKER_STORAGE_OPTIONS=.*-s " $DOCKER_STORAGE; then
+    return 0
+  fi
+  if grep -q -e "^DOCKER_STORAGE_OPTIONS=.*--storage-opt dm.thinpooldev" $DOCKER_STORAGE; then
+    return 0
+  fi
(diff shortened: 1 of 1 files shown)


rhvgoyal:
I personally feel that it should be docker who should detect that storage has changed from last time and bail out and force user to do cleanup.

docker-storage-setup is all about just setting up an lvm thin pool for docker user. It should not be in business of poking around whether what docker currently is using.

Also by setting up an lvm thin pool we don't lose any data. It is still present in the /var/lib/docker/ dir somewhere.

rhvgoyal:
I do like the part where we make sure that nothing is configured in /etc/sysconfig/docker-storage though. That will protect us from few things.

- If user has edited this file, then we don't generate any thin pool. This will take care of the case of "-s overlay".
- If user has configured its own thin pool and using that, we will not generate our own and rewrite user
  configuration.

I think all the logic of grow root pvs will still get into the way. That is a misfit for this service. I am not sure are we still supposed to grow root pvs if we are not setting up thin volume. Growing root pvs seems to be orthogonal to setting up of storage for docker.

rhvgoyal:
@cgwalters 

I modified your patches a bit and generated two patches. Please have a look.

cgwalters:
This is obsolete.

PR #46, opened 2015-07-07
Detect existing storage config and bail out

rhvgoyal:
Fixes #45 

If docker storage is already configured, don't do anything. (Except growing root pvs and rootfs).

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -301,6 +301,21 @@ disable_auto_pool_extension() {
   rm -f ${profileDir}/${profileFile}
 }
 
+docker_storage_is_configured() {
+  local options
+  if [ ! -f "$DOCKER_STORAGE" ];then
+    return 1
+  fi
+
+  # if there is a value in DOCKER_STORAGE_OPTIONS=, assume storage is already
+  # configured.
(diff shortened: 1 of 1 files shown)


rhvgoyal:
One problem with these patches is that they will log error upon every run (after first run) saying storage is already configured and that can be annoying.

rhvgoyal:
I pushed one more time and lost our conversation in the process. That's bad.

rhvgoyal:
I guess commenting on commits is not a very good idea as doing a push again will lose all that conversation.

rhvgoyal:
@cgwalters 

You might be having all the mails in your mailbox. Point I was making is that growing of root pvs and rootfs does not seem to be connected with DEVS=.

cgwalters:
We may want to step back at some point soon and think about merging some of this functionality into docker itself.   Split off the growpart stuff into a separate service as you said.

rhvgoyal:
@cgwalters 

I created a separate issue and pull request #50 which does this and provides a new storage driver option.

cgwalters:
Good to close this one now?

rhvgoyal:
Yes, we can close it now. Its resolved.

PR #50, opened 2015-07-09
Provide a new option STORAGE_DRIVER

### LINKED ISSUE #49: Provide a new option STORAGE_DRIVER
Right now docker-storage-setup assumes devicemapper driver. Provide a new
    option STORAGE_DRIVER where one can specify other drivers like overlay.
    
    There are few benefits of this.
    
    - We want to avoid people to modify /etc/sysconfig/docker-storage directly
      as it can conflict with what docker-storage-setup script wants to do.
      By providing this option, those who want to use overlay, should be able
      to just edit /etc/sysconfig/docker-storage-setup and start the service.
    
    - One should be able to override STORAGE_DRIVER="". This will indicate that
      don't setup any storage driver. This is for the people who want a way to
      disable docker-storage-setup.

Also add support for configuring overlay through this option.

And in the end fail if user is switching drivers and one is already configured.

rhvgoyal:
Fixes #49 

It does few things.

- Provides a new option STORAGE_DRIVER
- Adds support to configure overlay
- Errors out if one driver is already configured and user is trying to switch drivers.

### DIFF
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -1,7 +1,7 @@
 .TH "DOCKER-STORAGE-SETUP" "1" "NOVEMBER 2014" "Helper Script for Docker Storage Setup" ""
 .SH NAME
 .PP
-docker\-storage\-setup - Grows the root filesystem and sets up a LVM thin pool for docker.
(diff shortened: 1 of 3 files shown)


cgwalters:
This looks pretty good to me.  :+1:

PR #54, opened 2015-07-20
Enables pvcreate to assume "yes" on all questions and not to wait on user confirmation in order to proceed 

ekuric:
If there is file system signature on device then pvcreate will prompt user to confirm to wipe FS signature. When this happens, d-s-s will silently ignore this,and will not create desired storage backend ( in this case thin lvm  ) and will start docker using loop lvm storage configuration. 
d-s-s at time only checks if there is/are partition(s) on device(s) we specify in "DEVS" section. Device can have ( due to usage of device for file system sometimes before ) file system signature - even there is not any partition on it. 

This PR enables pvcreate to assume "yes" on all questions during pvcreate step.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -155,9 +155,9 @@ create_data_lv() {
 
   # TODO: Error handling when DATA_SIZE > available space.
   if [[ $DATA_SIZE == *%* ]]; then
-    lvcreate -l $DATA_SIZE -n $DATA_LV_NAME $VG
+    lvcreate -y -l $DATA_SIZE -n $DATA_LV_NAME $VG
   else
-    lvcreate -L $DATA_SIZE -n $DATA_LV_NAME $VG
+    lvcreate -y -L $DATA_SIZE -n $DATA_LV_NAME $VG
   fi
 }
 
@@ -281,7 +281,7 @@ unit: sectors
 
 ${dev}1 : start=     2048, size=  ${size}, Id=8e
 EOF
-    pvcreate ${dev}1
+    pvcreate -y ${dev}1
     PVS="$PVS ${dev}1"
   done
 }
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -155,9 +155,9 @@ create_data_lv() {
 
   # TODO: Error handling when DATA_SIZE > available space.
   if [[ $DATA_SIZE == *%* ]]; then
-    lvcreate -y -l $DATA_SIZE -n $DATA_LV_NAME $VG
+    lvcreate -l $DATA_SIZE -n $DATA_LV_NAME $VG
   else
-    lvcreate -y -L $DATA_SIZE -n $DATA_LV_NAME $VG
+    lvcreate -L $DATA_SIZE -n $DATA_LV_NAME $VG
   fi
 }
 


rhvgoyal:
You seem to have two patches which contain overlapping contents. You will need to fix that.

Secondly, I am not sure if overwriting the existing fs is always the right thing to do. Somebody might complain that don't overwrite it by default.

I tested and pvcreate does fail if disk already had a valid fs. "sfdisk" did succeed though after giving warning.

Jul 20 08:08:06 vm2-f22 docker-storage-setup[6447]: Failed to get offset of the xfs_external_log signature on /dev/vda1.
Jul 20 08:08:06 vm2-f22 docker-storage-setup[6447]: 1 existing signature left on the device.
Jul 20 08:08:06 vm2-f22 docker-storage-setup[6447]: Aborting pvcreate on /dev/vda1.

rhvgoyal:
Actually above pvcreate error does not get resolved even if I use "-y". Looks like this is not a question at all and it might be a bug with pvcreate.

ekuric:
@rhvgoyal ok, check also lvcreate in d-s-s, ( thought it does not ( or at least not in man ) -y switch )
if you add -y to lvcreate line have you same bahaviour?

cgwalters:
Tools like `pvcreate` abort in this situation for a reason.  Why wouldn't we require system administrators to use tools like `wipefs` to take explicit action?  We don't want a configuration typo to blow away critical data.

ekuric:
@cgwalters we assume here that there are not partitions ( as first requirement for d-s-s ) on device and want to use it for storage, and it is added in DEVS section. So after we setup device ( we decided to do that , we want to use it ) in DEVS="" section afterwards pvcreate fails - admin will already do all what is needed, if device added in DEVS="" section then I guess that is what is wanted.Anyway I think at least we need to print that error message - as of d-s-s will silently fail and start docker with loop lvm.

rhvgoyal:
@ekuric 

I think lvcreate does not have any switch -y. So how are you using it?

w.r.t the error I am seeing, I am checking with lvm developers. There seem to be multiple discrepancies.

First of all sfdisk goes ahead and create the partition table despite the fact that it detects xfs signature. And it does not delete that existing signature and creates partition table signature on top.

/dev/vdb: device contains a valid 'xfs' signature; it is strongly recommended to wipe the device with wipefs(8) if this is unexpected, in order to avoid possible collisions
Checking that no-one is using this disk right now ... OK

I agree with @cgwalters that if a valid signature is there on the disk passed in DEVS=, then we should bail out with proper error message and ask user to wipefs to wipe signature clean and retry. We don't want to destroy user's critical data due to configuration error.

ekuric:
@rhvgoyal no I am not using it, I was just hoping it has it somewhere hidden and not documented. I agree with you and @cgwalters, then we should send that error message informing why device was not possible to use as pv, what do you think?

rhvgoyal:
@ekuric 

If you are not using it and if you have not tested it (lvcreate -y), I am not sure why are you generating the patch? Atleast please test the patch before you request a merge. Otherwise this just increases the work for reviewers.

rhvgoyal:
@ekuric 

Can you paste the error you are facing with "pvcreate"?

ekuric:
@rhvgoyal ok,will reproduce it and will come back, of course I tested it...but seems only -y with pvcreate worked in this particular case - you can see I edited that in second patch...will follow up, thank you

rhvgoyal:
Ok, I came to know about couple of more things.

- lvcreate does support --yes option. It is supposed to part of generic lvm commands.
- I noticed warning during "lvcreate" which says whether you want to wipe xfs signature or not. As we are running from systemd service context, default answer is "no". Still signatures get wiped out anyway. Because lvm seems to have another default --zero=yes which wipes first 4K of block.

I think it should be ok to make sure nobody can add a disk with data in pvcreate. Once we have passed that step, we should not have any need to check preserve signature and we can use "-y" in
lvcreate. I will create another pull request for this and see what does @cgwalters think.

ekuric:
@rhvgoyal I thought that when I faced this issue, but if it is risky to use -y,then at least we need to inform user  that pvcreate failed.

rhvgoyal:
I think at the time of "pvcreate" we should not use -y. But once disk has been added to volume group, it might be reasonably safe to wipe signatures on logical volumes.

cgwalters:
This is fixed by https://github.com/projectatomic/docker-storage-setup/pull/57 right?

rhvgoyal:
Part of it  is fixed by #57 (lvcreate -y). Other part we don't want to fix (pvcreate -y). So I think it can be closed it now.

Issue #55, opened 2015-07-20
Add skeleton /etc/sysconfig/docker-storage-setup


jeremyeder:
We want to encourage the use of LVM to manage docker's storage.  Our documentation instructs users to copy over /usr/lib/docker-storage-setup/docker-storage-setup to /etc/sysconfig/docker-storage-setup in order to customize defaults.

RHEL users are accustomed to looking in /etc/sysconfig to customize services, docker is no exception:

```
# rpm -qc docker
/etc/sysconfig/docker
/etc/sysconfig/docker-network
/etc/sysconfig/docker-storage
```

The lack of any /etc/sysconfig/docker-storage-setup in our package leads to some confusion about why that file does not exist.

This PR is to create a placeholder file at /etc/sysconfig/docker-storage-setup that instructs users where to find documentation on how to use docker-storage-setup, and brings docker-storage-setup in-line with the rest of the docker package's config files:

rhvgoyal:
@jeremyeder 

Will above patch do?

rhvgoyal:
I think I will need a directive in spec file to not overwrite /etc/sysconfig/docker-storage-setup over upgrade. Instead save it in .rpmsave. I forget to remember what was that directive.

rhvgoyal:
Patch merged. Closing issue.

PR #58, opened 2015-07-20
docker-storage-setup: Ship an empty /etc/sysconfig/docker-storage-set…

rhvgoyal:
…up file

Fixes issue #55 

Jeremy complained that we ask users to modify
/etc/sysconfig/docker-storage-setup  and we don't ship one. He prefers that
we ship an empty one with additional information so that it is easier for
users to find it and edit it.

### DIFF
--- a/docker-storage-setup-override.conf
+++ b/docker-storage-setup-override.conf
@@ -1,0 +1,4 @@
+# Edit this file to override any configuration options specified in
+# /usr/lib/docker-storage-setup/docker-storage-setup.
+#
+# For more details refer to "man docker-storage-setup"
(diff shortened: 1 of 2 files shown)


jeremyeder:
LGTM :+1: thanks for this, it should help direct users to the right place.

rhvgoyal:
Added %config(noreplace) to this new file to make sure user's edited /etc/sysconfig/docker-storage-setup does not get overwritten.

cgwalters:
OK.  However I'd recommend committing this to Fedora dist-git immediately as otherwise I suspect Lokesh could miss the changes to sources/Dockerfile here.

rhvgoyal:
@lsm5 Ping.

lsm5:
I should've seen this before I updated, anyway, re-updating now. Also, could you send future spec updates to the docker spec file itself. I'm hoping we can remove the spec file here: https://github.com/projectatomic/docker-storage-setup/pull/59

rhvgoyal:
I think keeping a spec file in this project is important because that way a user knows how different files are to be installed and treated.

upstream docker does not have a spec file. Any changes to fedora docker spec file are specific to fedora only.

PR #59, opened 2015-07-22
rpm spec file removed, no longer used

lsm5:
Send all changes to the docker rpm spec

Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>


rhvgoyal:
I think it is still important to keep this spec file. We have decided to package it as part of docker but somebody else should be able to use this spec file and package docker-storage-setup.

jfarrell:
+1 to leaving the spec included

Issue #67, opened 2015-07-31
By default reserve lesser space for data volume


rhvgoyal:
Right now DATA_SIZE is 60%FREE. That is use 60% of free space for data volume. Dan Walsh was saying that this sounds like a high percentage. It might be a good idea to drop it down a bit. Say 40%FREE.

rhvgoyal:
@rhatdan CC

rhatdan:
If we have auto growth, I don't see a reason to pre-allocate such a huge percentage.  Of course the problem with this is there is no great answer.  Having a minimal size + growth  should solve most problems.

PR #68, opened 2015-08-03
docker-storage-setup: Default to using 40% of free space for data volume

rhvgoyal:
Fixes issue #67 

Some people complained that using 60% of free space might be a lot. Instead
start small and let the pool grow automatically. 

So drop the default data volume size to 40% of free space.


rhvgoyal:
@cgwalters 

Can you please review this one too.

rhvgoyal:
@cgwalters 

Agreed. Both are arbitrary numbers. We can adjust these as we have more data to justify a different number.

Issue #70, opened 2015-08-21
Allow data size suffix of type GB and GiB


rhvgoyal:
With MIN_DATA_SIZE patch, we restricted suffixes to be of single character. But users reported that they also use syntax of type GB. Allow GB and GiB.

Akasurde:
Can we close this issue as PR is merged?

PR #71, opened 2015-08-21
docker-storage-setup: Relax restriction on data size suffixes

rhvgoyal:
Fixes issue #70 

Previously we were not doing any checks on DATA_SIZE suffixes and passing
everything to lvm directly. It turns out lvm just checks if things are
starting with valid suffix and anythign after that is acceptable. For
example -L 5GHello is acceptable and lvm will treat it as 5GiB.

Then with MIN_DATA_SIZE patch, we put limitations on suffixes. One could
use only one valid suffix character and nothing after that. For example,
only 5G is acceptable and 5GB and 5GiB are failing.

This is being seen as regression.

This patch allows now [bB] and i[bB] after the valid suffix. That is,
values of type 5G, 5GB, 5GiB are acceptable and all will be treated
as 5GiB.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -148,7 +148,7 @@ create_metadata_lv() {
 }
 
 convert_size_in_bytes() {
-  local size=$1 prefix
+  local size=$1 prefix suffix
 
   # if it is all numeric, it is valid as by default it will be MiB.
   if [[ $size =~ ^[[:digit:]]+$ ]]; then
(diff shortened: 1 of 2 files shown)


cgwalters:
One trivial question, also wonder if we can add tests for these now too.  But :+1: from me in current form too.

rhvgoyal:
I will create a separate issue and PR for creating some tests. This patch has been tested by Cai Qian. So I prefer that this be merged as it without blocking on tests.

rhvgoyal:
I opened issue #72 to keep track of test cases.

PR #74, opened 2015-08-28
Docker root on volume

### LINKED ISSUE #73: Add a facility to be able to prepare a separate volume for docker root
In case of overlay driver, all images and containers will be stored on rootfs. Some might not want that. Hence create options so that docker-storage-setup can prepare a separate volume and mount it somewhere and force docker to use that as root directory. That way all image/containers/volumes will be stored on a separate volume instead of rootfs.

rhvgoyal:
Fixes #73 

Add facility so that one can ask d-s-s to prepare a seaparate volume for docker root directory.

### DIFF
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -57,6 +57,19 @@ DOCKER_ROOT_DIR:
       backend is in use, containers and images will also be stored
       in this directory.
 
+DOCKER_ROOT_VOLUME:
+      If set to "yes", then a separate logical volume will be setup
+      for docker root and it will be mounted at target DOCKER_ROOT_DIR
+      and docker will use DOCKER_ROOT_DIR as root directory.
+
+      One possible use case is with overlay driver where one might not
+      want docker to store all images/containers/volumes on rootfs.
+
+DOCKER_ROOT_VOLUME_SIZE:
+      This specfies the size of docker root volume. It takes effect only
+      if DOCKER_ROOT_VOLUME=yes. It defaults to 40% of free space in
+      the volume group. It can accept same values as DATA_SIZE.
+
 AUTO_EXTEND_POOL:
       Enable automatic extension of pool by lvm. lvm can monitor
       the pool and automatically extend it when pool is getting full.
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -50,6 +50,13 @@ GROWPART:
       disabled. Use GROWPART=true to enable automatic partition
       table resizing.
 
+DOCKER_ROOT_DIR:
+      Specifies the root directory for docker metadata and volumes.
+      If the devicemapper backend is in use, containers and images
+      will be stored in a dedicated LVM thin pool. If the overlay
+      backend is in use, containers and images will also be stored
+      in this directory.
+
 AUTO_EXTEND_POOL:
       Enable automatic extension of pool by lvm. lvm can monitor
       the pool and automatically extend it when pool is getting full.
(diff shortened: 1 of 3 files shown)


cgwalters:
Overall looks OK.  I actually didn't know about `-g`, I'd been just using symlinks myself.

Hmm, won't the admin doing this also have to relabel the target as `var_t` via e.g. `semanage fcontext`, or alternatively just `mkdir -p && chcon -t var_t`?  That'd be worth noting.

rhvgoyal:
selinux labeling is a good point. I had been doing all my testing with selinux disabled (as for some reason selinux will not work with upstream docker). I will spend some time on figuring out this selinux bit.

rhatdan:
Make sure you have docker-selinux installed and upstream should work fine.

rhvgoyal:
@cgwalters  I have taken care of review comments and pushed patches again. PTAL.

rhvgoyal:
@cgwalters ping.

cgwalters:
I think this one should be run by the Fedora Docker team at least, not just me.  

I'm a bit uncertain about automatically running `semanage` - that's a highly privileged operation as it persistently changes the policy.  Raw `chcon --reference` doesn't (but also isn't "persistent", though I think people running `restorecon -R /` etc. have more serious problems).

Some of this feels like it'd actually be better living in an installation setup program (i.e. Ansible).  Particularly for the complex cases where e.g. one wants to do "converged" storage in a Kubernetes cluster where a portion of the host's disk drives go to persistent volumes, a portion goes to caching images and containers, etc.

That said we do need to make the "single node" case easy, and particularly "switch single node to overlayfs on boot" would be nice.

With this we're starting to grow a lot of functionality and complexity without many test cases to match.  There is some work on that at my employer internally and I think we could try to extend coverage to make use of some of these options.

rhvgoyal:
@rhatdan 

Can you please look at this PR from selinux point of view.

rhvgoyal:
@cgwalters, Your comment is combination of multiple issues, so I will break down my responses and respond to each issue separately.

Though we are changing policy in persistent manner using "semanage" but scope of policy change is limited. That is it happens only for a directory and happens only if that directory is created by d-s-s.

Similarly restorecon impact is limited to only container root and this newly created directory.

But I will let Dan Walsh decide if it is alright in the context of d-s-s or not.

rhvgoyal:
Second *core* issue is that should this functionality be in d-s-s or not. I agree that d-s-s is getting complicated now but it is just offering more complicated storage setup now. I feel that we can have storage setup functionality in d-s-s as well as higher layers. And if installer chooses to setup storage, then they can just disable d-s-s by putting "STORAGE_DRIVER=" in /etc/sysconfig/docker-storage.

I am looking at it from pure fedora host perspective also. Where a user might just add another disk and might want to configure storage there. Or leave extra disk space and might want to configure storage. 

My experience tells me that getting extra configuration screens in anaconda is hard. In fact even one gets it in, maintaining it as a separate package is much easier then when it runs in the context of installer.

So while I am not opposed to somebody introducing custom storage setup features in installers, but that should not preclude enhancing d-s-s. This does solve the case of single host nicely without worrying about what installer is being used.

rhvgoyal:
Third issue is about automated test cases. I agree that being able to put automated test cases is very nice. You mentioned that you are already doing some work on automated test cases and these could be extended to test this. That makes sense. But I guess that's not a blocker for this patch series.

rhatdan:
This one is now supported.

Issue #75, opened 2015-09-09
Error out early if root device is not a logical volume


rhvgoyal:
Right now d-s-s assumes that system is using lvm and root devices is a logical volume backed by a logical volume group. But that's not the case always. Detect this early and exit gracefully.

PR #76, opened 2015-09-09
docker-storage-setup: Exit early if a valid volume group is not specified/found

rhvgoyal:
Fixes issue #75 

Right now docker-storage-setup assumes that root device is a lvm logical
volume and there is a valid volume group backing it. But if user is not
using lvm and root devices is just a block device or a partition, then
this assumption is not valid any more.

Detect such cases and error out early gracefully.


cgwalters:
Looks reasonable, thanks!

Issue #78, opened 2015-09-10
error out if /usr/bin/growpart is not available and GROWPART=true


rhvgoyal:
If user wants to grow partitions and specified GROWPART=true and /usr/bin/growpart executable is not installed, then error out instead of continuing.

PR #79, opened 2015-09-10
docker-storage-setup: Error out if /usr/bin/growpart is not available

rhvgoyal:
Fixes issue #78 

Right now if user wants to grow partitions and specifies GROWPART=true then
d-s-s will not fail even if /usr/bin/growpart is not available.

Hence, give an error message and exit is /usr/bin/growpart is not available
and user specified GROWPART=true.


cgwalters:
:+1:

PR #80, opened 2015-09-28
Correct version check for Docker > 1.7

delitescere:
This line would have failed as soon as Docker `2.0` was released. Also de-quoted for numeric comparison (keeps consistent with rest of file)


rhvgoyal:
So docker will move to 2.0 instead of continuing with 1.11 ?

If that's the case, then above LGTM.

rhvgoyal:
@cgwalters ping

cgwalters:
:+1: from me.

PR #81, opened 2015-10-20
Enable deferred device deletion feature for devicemapper

rhvgoyal:
Docker 1.9 has now deferred device deletion feature support for devicemapper. That means a container can be removed even if underlying device is busy as it has leaked into another mount namespaces. We are having issues where a container is launched and then SPC is launched and root device of containers launched before SPC leaks into SPC container under /hosts. That means none of the container launched before SPC can be deleted till SPC exits.

This should help fix that issue.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -88,6 +88,25 @@ get_deferred_removal_string() {
 	fi
 }
 
+get_deferred_deletion_string() {
+	local version major minor
+
+	if ! version=$(get_docker_version);then
+		return 0
+	fi
+	[ -z "$version" ] && return 0
+
+	major=$(echo $version | cut -d "." -f1)
+	minor=$(echo $version | cut -d "." -f2)
+	[ -z "$major" ] && return 0
+	[ -z "$minor" ] && return 0
+
+	# docker 1.9 onwards supports deferred device removal. Enable it.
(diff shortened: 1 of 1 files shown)


rhvgoyal:
@cgwalters PTAL.

rhatdan:
I think we should cleanup the code and figure it will work on all modern versions of docker.  At least all of the ones we currently support.  Or just do it the old way.

rhatdan:
Looks like it was added in July
```
commit e221291a25fb905a476d9328785bdbfbd9cb7373
Merge: 60b34cb 41588a5
Author: Jessie Frazelle <jfrazelle@users.noreply.github.com>
Date:   Mon Jul 13 17:04:06 2015 -0700

    Merge pull request #14194 from phemmer/cli-version-format
    
    add --format flag to `docker version`
```

rhvgoyal:
@rhatdan Thanks for finding that commit. So yes, if it has been added in july (3 months back), then it is fairly recent. So that means docker 1.7 does not support it and docker 1.8 likely supports it.

So what do you consider modern version of docker. Should this work with docker 1.7 or not?

rhatdan:
I think we should move forward to docker-1.8 and just require that version or greater for docker-storage-setup

rhvgoyal:
@rhatdan @cgwalters I got rid of old method of querying version. PTAL.

PR #83, opened 2015-11-12
Allow overlay to be used even if image is not using lvm

rhvgoyal:
Right now we are checking early that either rootfs is backed by a volume group or user has specified a volume group using VG=. Otherwise script exits early thinking no progress can be made.

This was true when we only supported devicemapper storage driver. Now we also support overlay and overlay can still be configured even if a valid VG is not there.

Fix it.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -291,6 +291,13 @@ create_lvm_thin_pool () {
 }
 
 setup_lvm_thin_pool () {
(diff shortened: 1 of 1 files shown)


rhvgoyal:
@cgwalters @rhatdan PTAL.

rhatdan:
I don't see any problems, but I don't exactly no the code.

cgwalters:
LGTM, thanks!

PR #84, opened 2015-11-19
Set VG_EXISTS=1 after volume group creation

rhvgoyal:
Set VG_EXISTS=1 after volume group creation. Later code checks it to
make sure volume group exists otherwise code errors out.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -435,6 +435,7 @@ EOF
 create_extend_volume_group() {
   if [ -z "$VG_EXISTS" ]; then
     vgcreate $VG $PVS
+    VG_EXISTS=1
   else
     # TODO:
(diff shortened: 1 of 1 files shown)


rhvgoyal:
@rhatdan @cgwalters PTAL.

This is breaking simple case of DEVS=foo VG=bar. When volume group is created we need to set VG_EXISTS=1 so that later logic can make sure VG exists before trying to create lvm thin pool.

rhatdan:
LGTM

cgwalters:
:+1: 

We are going to need to get serious about some automated testing here.

rhvgoyal:
@cgwalters I agree. I am tired of fixing these things. There are too many small breakges. We really really need automated tests now.

PR #85, opened 2015-11-23
Bug 1261565 - Allow docker-storage-setup.sh to be run n-times

jwhonce:
* query the current volume groups and partitions to allow
  docker-storage-setup.sh to be idempotent

### DIFF
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -11,7 +11,7 @@ docker\-storage\-setup - Grows the root filesystem and sets up storage for docke
   Print usage statement
 
 .SH EXAMPLES
-Run \f[B]docker-storage-setup\f[] after setting up your configuration in 
+Run \f[B]docker-storage-setup\f[] after setting up your configuration in
 /etc/sysconfig/docker-storage-setup. One can look at
 /usr/lib/docker-storage-setup/docker-storage-setup for various options and
 their default settings. Anything user wants to change, should be changed
@@ -25,13 +25,15 @@ functionality to work properly.
 
 STORAGE_DRIVER:
       Specify a storage driver to be used with docker. Default
-      driver is devicemapper. Other valid values are "overlay"
+      driver is "devicemapper". Other valid values are "overlay"
       and "". Empty string tells docker-storage-setup to not do
       any storage setup.
 
-DEVS: A quoted, space-separated list of devices to be used.  This
-      currently expects the devices to be unpartitioned drives.  If
-      "VG" is not specified, then use of the root disk's extra space
+DEVS: A quoted, space-separated list of devices to be used.
+      If a drive is partitioned and contains a ${dev}1 partition,
+      that partition will be configured for use. Unpartitioned
+      drives will be partitioned and configured for use. If "VG"
+      is not specified, then use of the root disk's extra space
       is implied.
 
 VG:   The volume group to use for docker storage.  Defaults to the
@@ -100,18 +102,18 @@ MIN_DATA_SIZE:
 
 \f[B]Sample\f[]
 
-A simple, sample /etc/sysconfig/docker-storage-setup: 
+A simple, sample /etc/sysconfig/docker-storage-setup:
 
 DEVS=/dev/vdb
 
 DATA_SIZE=8GB
 
-.fi 
+.fi
 
-.SH HISTORY 
+.SH HISTORY
 
-.PP 
-November 2014, originally compiled by Joe Brockmeier <jzb@redhat.com> 
+.PP
+November 2014, originally compiled by Joe Brockmeier <jzb@redhat.com>
 based on comments in Andy Grimm's <agrimm@redhat.com> script.
 .SH AUTHORS
 Joe Brockmeier
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 #--
-# Copyright 2014 Red Hat, Inc.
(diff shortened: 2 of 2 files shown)


jwhonce:
@rhvgoyal Please review.  Thanks.

jwhonce:
@rhvgoyal Updated from your comments.

rhvgoyal:
@jwhonce Can you please provide some description of the problem and how it is being fixed in the patch commit message.

rhvgoyal:
I think we also need to prefix some kind of tag to messages to figure out what message is info and what message is warning and what message is error. For example, I ran docker-storage-setup and added an already partitioned disk and expected error. I see following on console.

`INFO: Volume group backing root filesystem could not be determined
Using existing /dev/sda1 partition
Device /dev/sdb is partitioned but missing /dev/sdb1 partition
`
In this case, first message is INFO and second is an ERROR. If we prefix it accordingly, one can easily figure out if script completed successfully or not. Just more intuitive logging.

rhvgoyal:
@jwhonce I had commented on getting INFO and Error messages right. I don't see those changes in latest patches.

rhvgoyal:
I had suggested the actual messages which I felt were much more clear then current once. Can't find these in this thread. Must have been in commit.

Also I had requrested for prefixing atleast "Error" or "ERROR" before actual error. Otherwise I can't differentiate between info and error. That change is also not there.

jwhonce:
@rhvgoyal Try a refresh maybe?  Those changes are in commit d0580ca8b6e22f774aa09a2df7fc538f6d0c8d10

rhvgoyal:
@jhwonce Ok, I see that new messages are in last patch of the series. Can we please modify first patch to put in the desired messages. It is never a good practice to first introduce a set of changes in first patch and revert these changes in 4th patch.

rhatdan:
If you merge the patches into one, this is not a problem.  :8ball:

rhvgoyal:
@rhatdan, sure. this is the extra overhead of breaking down changes into a patch series. If changes are requested, then one needs to figure out which patch is going to change and change that patch. Keeping everything in one patch simplifies that. But makes review harder.  So we gain some and loose some. I think it is just a matter of getting used to the workflow. I have come from a background where we always breakdown patches and john has always done everything in one patch and it is taking a bit to getting used to.

rhvgoyal:
For future we can debate what's the patch convention we want for docker-storage-setup project. One patch vs broken down patch series. I personally think that we need to strike right balance. Too much of fine grained patches are not good and putting everything into a single patch is not good either (depending on amount of change).

rhatdan:
@rhvgoyal Just joking....

jwhonce:
@rhatdan @rhvgoyal Still striving for that happy medium, too much muscle memory in my fingers for amend'ing commits all into one. :-)

I believe this set of commits is what you are looking for @rhvgoyal

rhvgoyal:
@jwhonce cool that you classified all the messages either as INFO or ERROR. This helps a lot.

This patch set now looks good to me except one minor nit. I don't think we should add idea IDE generated files into the docker-storage-setup project. Generally a project carries files in .gitignore which are generated by that project. Anything generated by other tools outside of the project, it is better to create a gloabl gitignore file. 

docker seems to follow the same thing. 

https://github.com/docker/docker/blob/master/.gitignore#L2

jwhonce:
@rhvgoyal kubernetes uses a project .gitignore for these files to ensure they never get accidentally added.  I suspect this is personal/project style issue.

rhvgoyal:
LGTM

rhvgoyal:
@cgwalters Would you like to review this one before I merge it.

jwhonce:
@cgwalters I fixed the error message.  I did not fix the test, as anything using bash arrays is a messier syntax.

rhvgoyal:
@jwhonce do we really need bash arrays. Can't we simple run for loop on current_devices.

for d in $current_devices; do
done

rhvgoyal:
@jwhonce @cgwalters I wrote a small patch on top of john's to replace using regexp for device mapthing. Does it look better.

https://github.com/rhvgoyal/docker-storage-setup/commit/ff5d3fb91c14be3392b2c492a0a842c537a5a2ea

jwhonce:
@rhvgoyal Please echo the message to >&2. Thanks

cgwalters:
rhvgoyal@ff5d3fb looks OK to me.  Squash into this commit and let's merge?

rhvgoyal:
@jhwonce do you mind apply my patch on top of yours and push your patch again. Also please take care of outputting error message to stderr.

jwhonce:
@rhvgoyal Done

rhvgoyal:
@jwhonce Thanks. Merging it now.

rhvgoyal:
@jwhonce looks like following code is not working. I passed in loop device /dev/sda1 and it is not detected as parition. 

  #validate DEVS elements
  for dev in $DEVS; do
    if [[ $dev =~ ".*[0-9]$" ]]; then
      echo "ERROR: Partition specification unsupported at this time." >&2
      exit 1
    fi

rhvgoyal:
@jhwhonce

I think there is one more bug. I specified two devices in DEVS and only one device gets partitioned and gets added to volume group. one device remains untouched.

DEVS="/dev/sda /dev/sdb"
VG=foo-vg

rhvgoyal:
I think issue is that variable name is not proper and one `s` is missing.

needs_partitioned="$dev $need_partitioned"

It should be

needs_partitioned="$dev $needs_partitioned"

PR #87, opened 2015-12-04
Add quotes around generated DOCKER_STORAGE_OPTIONS

samuelkarp:
Quotes around `DOCKER_STORAGE_OPTIONS` are required for an init script which does `source /etc/sysconfig/docker-storage`.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -119,12 +119,12 @@ get_devicemapper_config_options() {
     fi
   done )
 
-  storage_options="DOCKER_STORAGE_OPTIONS=--storage-driver devicemapper --storage-opt dm.fs=xfs --storage-opt dm.thinpooldev=$POOL_DEVICE_PATH $(get_deferred_removal_string) $(get_deferred_deletion_string)"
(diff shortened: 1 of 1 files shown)


rhvgoyal:
@samuelkarp what's broken if I don't add these double quotes?

rhvgoyal:
I did some basic testing and it seems to be working fine. I am assuming you are not using systemd as it works fine with system as systemd seems to be separating variables based on new line.

I am assuming you might be doing 'source /etc/sysconfig/docker-storage-setup' and then running into issues?

rhvgoyal:
@samuelkarp Can you proivde a description for the patch.

samuelkarp:
@rhvgoyal My apologies for the lack of description.  Your assumption is correct; I am using an init script invoked by Upstart rather than systemd.  As part of the init script, it executes `[ -e /etc/sysconfig/docker-storage ] && . /etc/sysconfig/docker-storage`, which fails with a bash error when quotes are missing.

rhvgoyal:
LGTM.

rhvgoyal:
@cgwalters ping. Would you like to have a quick look at it?

cgwalters:
LGTM too, we'll just have to remember to keep doing this as the code changes.

PR #88, opened 2015-12-08
The  right hand string of the bash regex operator should not be quoted.

jwhonce:
@rhvgoyal to correct issue from comment https://github.com/projectatomic/docker-storage-setup/pull/85#issuecomment-163029033


rhvgoyal:
@jwhonce. Thanks. I tested it and it is working for me again.

LGTM.

PR #89, opened 2015-12-10
Fix needs_partitioned variable name

rhvgoyal:
Variable name should be needs_partitioned instead of need_partitioned.


rhvgoyal:
CC @jwhonce @cgwalters

jwhonce:
@rhvgoyal LGTM

PR #90, opened 2015-12-10
Make sure passed in block devices are valid block devices

rhvgoyal:
Right now one can pass in any xyz string in DEVS= and there are no error
checks and sfdisk tries to partition it and fails with odd error messages
like as follows.

./docker-storage-setup.sh: line 463: * 2 - 2048 : syntax error: operand expected (error token is "* 2 - 2048 ")
ERROR: No valid volume group found. Exiting.

Instead check for valid block devices early and fail gracefully.

### DIFF
--- a/libdss.sh
+++ b/libdss.sh
@@ -1,6 +1,22 @@
 #!/bin/bash
 # Library for common functions
 
+# echo info messages on stdout
+Info() {
+  echo "INFO: ${1}"
+}
+
+# echo error messages on stderr
+Error() {
+  echo "ERROR: ${1}" >&2
+}
+
+# echo error on stderr and exit with error code 1
+Fatal() {
+  Error "${1}"
+  exit 1
+}
+
 # checks the size specifications acceptable to -L
 check_numeric_size_syntax() {
   data_size=$1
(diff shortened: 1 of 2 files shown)


rhvgoyal:
CC @jwhonce @cgwalters @rhatdan @cevich Any review/testing is appreciated.

rhvgoyal:
@rhatdan Ok, I took care of your comments. Though I had to create more patches in the process but that's all good cleanup.

libdss.sh
```diff
@@ -1,6 +1,22 @@
 #!/bin/bash
 # Library for common functions
 
+# echo info messages on stderr
+Info() {
+  echo "INFO: ${1}" >&2
+}
```
rhatdan: Should info go to stdout?

rhatdan:
Other then comment on Infor :+1:

libdss.sh
rhvgoyal: I think so. Will change it.

rhvgoyal:
@rhatdan Fixed Info(). Looking for your LGTM and I will merge it.

rhatdan:
LGTM.

rhatdan:
One think you might want to do with your functions is to output $* or $@ (Never remember which is which.

Then you could just do

```
Error -n This will not print a newline
```
WOuld work

rhvgoyal:
Ok, got it. I think that improvement for some other time. Merged PR for now.

rhatdan:
Yes I agree.

PR #91, opened 2015-12-14
Allow block device name as valid input to DEVS

rhvgoyal:
Looks like we allow just block device name as valid input. Say DEVS=sda.
Recent changes broke it as it expected input to be a full device path
name (/dev/sda). 

For now fix it by continuing to accept just device name as input and
prefix /dev/ internally. 


rhvgoyal:
@rhatdan @cgwalters PTAL.

rhatdan:
:+1:

PR #93, opened 2015-12-18
Add a simple bash script to run some tests

rhvgoyal:
Add a simple test case and bash script to run tests.

### DIFF
--- a/tests/001-test-use-devs-to-create-thin-pool.sh
+++ b/tests/001-test-use-devs-to-create-thin-pool.sh
@@ -1,0 +1,49 @@
+# Test DEVS= directive. Returns 0 on success and 1 on failure.
+test_devs() {
+  local devs=$DEVS
+  local test_status
+  local testname=`basename "$0"`
+  local vg_name
+  local vg_name="dss-test-foo"
+
+ # Error out if any pre-existing volume group vg named dss-test-foo
+  for vg in $(vgs --noheadings -o vg_name); do
+    if [ "$vg" == "$vg_name" ]; then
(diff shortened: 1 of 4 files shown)


rhvgoyal:
cc @cgwalters @rhatdan 

I am tired of not having a way to tests new changes to this script. So here is the start. A simple test case which creates a thin pool from devices specified in DEVS. And a script to run tests. One should be able to easily add more tests.

Idea is to keep on adding tests and run all these tests on new PR. These are simple bash scripts. So one can easily run these tests on laptop as long as they have a spare disk to run tests on.

cgwalters:
This looks like it's potentially destructive for the current OS if we're looking at `/dev/sdb` by default?

Is this intended for a disposable VM, or for direct use on a Linux workstation?  If it's only the former, I'd add some verbose warnings against running it directly.

rhvgoyal:
@cgwalters Actually I forget to coment out /dev/sdb. That's was supposed to be there for an example and I was testing with /dev/sdb. I will fix it.

In practice user will have to figure out which disk can be used for testing and put it in dss-test-config.

rhvgoyal:
@cgwalters Fixed. Now /dev/sdb is not the default. Tests will error out if there is no disk to test. User will have to specify a disk to test with.

cgwalters:
Looks like a fine start to me.  There's obviously a lot more we can do but let's just try to get this wired up to Jenkins and build from there.

rhvgoyal:
@cgwalters Yes we can do lot more in terms of making sure disk passed in is not used. For now I just want to get started and keep on modifying it.

I will merge this PR and then figure out how to get this wired up with jenkins.

PR #94, opened 2015-12-18
Give run-tests.sh execute permission

rhvgoyal:
Right now run-tests.sh is not executable. Make it executable.


rhvgoyal:
cc @cgwalters @rhatdan

PR #95, opened 2015-12-18
Fix exit code of run-tests.sh

rhvgoyal:
Right now run-tests.sh exits with success even if a test failed. Fix it.


rhvgoyal:
cc @cgwalters @jlebon @rhatdan

jlebon:
Makes sense to me.

PR #102, opened 2016-02-12
Cloud-Init triggered install/start of docker will always hang

### LINKED ISSUE #77: Cloud-Init triggered install/start of docker will always hang
If Cloud-init is used to trigger installs of docker, docker startup, and provisioning in general will just hang. This is irrespective of what the actual method of install/config is - ie. Puppet/Ansible etc.

Expanding on that:

* In our method, we use cloud-init to trigger Puppet or Ansible for the initial configuration run.
* This succeeds until the docker daemon is requested to start.
* The docker daemon service requires ```docker-storage-setup``` to have already run.
* The ```docker-storage-setup``` systemd config requires ```cloud-final.service``` ( ie. cloud-init) to have finished - https://github.com/projectatomic/docker-storage-setup/blob/master/docker-storage-setup.service#L3
* The daemon start process just hangs at this point, because of this circular dependency.

The result is that provisioning cannot complete until cloud-init has failed/died - or the box has been rebooted.

Commenting out ```After=cloud-final.service``` makes this problem go away.

I understand the thinking behind the requirement - basically, ensure the box has had it's storage configured by Cloud-init, to be used by ```docker-storage-setup``` - but the existing restriction makes more sophisticated hands-off provisioning awkward.

I'm not sure what to suggest as a solution here, sorry!

jfarrell:
Removes cloud-init dependency from the service to avoid hangs.

Fixes #77

### DIFF
--- a/docker-storage-setup.service
+++ b/docker-storage-setup.service
@@ -1,6 +1,5 @@
 [Unit]
 Description=Docker Storage Setup
-After=cloud-final.service
 Before=docker.service
 
 [Service]


jlebon:
But what if cloud-init is used to do some storage preparations? If you need to start docker within cloud-init, try using the `--no-block` option.

jlebon:
Ahh, apologies. I see `--no-block` was already mentioned in the original bug report. And right, that only helps if you don't need to do other docker things afterwards.

jfarrell:
We hit this issue when launching an aws cloudformation stack which at boot ran cloud-init with docker-storage-setup. using a drop-in to add this ordering if needed and removing it from the service seems like it would be a more flexible pattern rather than requiring it or the `--no-block` route

rhvgoyal:
As a workaround, you should be able to override docker-storage-setup.service from cloud-init before starting docker, and get rid of dependency on cloud-init. right?

jfarrell:
This feels backwards to me to make the default behavior depend on cloud-init when not all systems will use it. Is this package based on some atomic bootstrapping defaults and not really intended to be used outside that universe? 

The rational behind this pr went along with a patch I submitted to the docker puppet module to allow it to support docker-storage-setup once again due to a race condition in that module between docker-storage-setup sysconfig and the docker service starting on centos 7 systems. If using a full override is the preferred route then this pr can be closed.

rhvgoyal:
I don't know why there is this dependency on cloud-final.service. 

One reason I can think of that using cloud-init one can bring a storage online and expect docker-storage-setup to use it. I guess that users of cloud-init can override docker-storage-setup.service to make that case work.

The other reason colin alluded to was growing rootfs. I have not looked into cloud-init code but my understanding is that they also grow rootfs. But may be they are not able to deal with lvm so that codee had to be duplicated in docker-storage-setup. Now if that's the case, on a non-lvm cloud image, who is going to make sure that both cloud-init and docker-storage-setup don't try to grow rootfs at the same time and step on each other's toes.

May be we need to get rid of growing rootfs code from docker-storage-setup and teach cloud-init to be able to deal with this and then remove this dependency.

@cgwalters What do you think?

cgwalters:
Yeah, deduping d-s-s growpart and cloud-init growpart would help a lot.  Maybe even stick it in systemd.

However, things get tricky here because remember we want to be able to *configure* d-s-s from cloud-init.

The root issue here is triggered by trying to start docker inside cloud-init.  You should be able to do instead:

```
systemctl enable docker
systemctl start --no-block docker
```

rhvgoyal:
--no-block helps only in certain cases. If one wants to make use of docker service from cloud-init, then --no-block will not help. And that seems to be their concern.

@jfarrell, I think till we  de-dedupe grow rootfs logic between cloud-init and docker-storage-setup, removing this depedency can create issues. So please use the workaround of overriding docker-storage-setup.service file in cloud-init.

PR #103, opened 2016-02-26
Allow dm.fs to be overridden

samuelkarp:
We're seeing some [bad behavior](https://github.com/docker/docker/issues/20707) when xfs is used and the underlying LV/VG/PV are completely full.  Experimentally, ext4 has better failure modes, so this commit allows setting the filesystem.

### DIFF
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -29,6 +29,10 @@ STORAGE_DRIVER:
       and "". Empty string tells docker-storage-setup to not do
       any storage setup.
 
+EXTRA_DOCKER_STORAGE_OPTIONS:
+      A set of extra options that should be passed to the Docker
+      daemon.
+
 DEVS: A quoted, space-separated list of devices to be used.
       If a drive is partitioned and contains a ${dev}1 partition,
       that partition will be configured for use. Unpartitioned
--- a/docker-storage-setup.conf
+++ b/docker-storage-setup.conf
@@ -4,6 +4,11 @@
 #
 STORAGE_DRIVER=devicemapper
 
+# A set of extra options that should be appended to the generated
+# DOCKER_STORAGE_OPTIONS string. These options will be passed to the Docker
+# daemon as-is and should be valid Docker storage options.
+# EXTRA_DOCKER_STORAGE_OPTIONS="--storage-opt dm.fs=ext4"
+
 # A quoted, space-separated list of devices to be used.  This currently
 # expects the devices to be unpartitioned drives.  If "VG" is not specified,
 # then use of the root disk's extra space is implied.
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -108,8 +108,19 @@ get_deferred_deletion_string() {
   fi
 }
 
+extra_options_has_dm_fs() {
+  local option
+  for option in ${EXTRA_DOCKER_STORAGE_OPTIONS}; do
+    if grep -q "dm.fs=" <<< $option; then
+      return 0
+    fi
+  done
+  return 1
+}
+
 get_devicemapper_config_options() {
   local storage_options
+  local dm_fs="--storage-opt dm.fs=xfs"
 
   # docker expects device mapper device and not lvm device. Do the conversion.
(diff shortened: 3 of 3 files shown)


jfarrell:
+1 looks good

rhvgoyal:
This is an option which we directly pass to docker and don't take any action ourselves. And there are many docker options one might need to pass. For example, dm.mountopts, dm.basesize etc.

I am wondering why not create a generic option which will simply pass the option string to docker daemon. For example.

DOCKER_STORAGE_OPTS="--storage-opt dm.fs=xfs --storage-opt dm.mountopt=foo"

It is user's responsibility to provide option string which makes sense to docker and also which does not conflict with other options which will be generated internally by docker-storage-setup. That way we will not end up defining one option for every docker storage option.

rhvgoyal:
Thinking more about it. Problem with DOCKER_STORAGE_OPTS is that we are not in a position to override some of the defaults and user specified options might conflict. For example, by default we are specifiying dm.fs=xfs and if user provides dm.fs=ext4 then it is not a pretty situation.

May be we can scan DOCKER_STORAGE_OPTS string and see if user has passed in dm.fs option or not. If not, we pass in dm.fs=xfs

samuelkarp:
> Thinking more about it. Problem with DOCKER_STORAGE_OPTS is that we are not in a position to override some of the defaults and user specified options might conflict.

Yep, that's why I went for the simplest option here.

rhvgoyal:
Can we scan and use dm.fs accordingly. Otherwise we will just end up creating a new knob for every docker storage option.

samuelkarp:
@rhvgoyal I've updated the pull request to scan and use the existing `dm.fs` accordingly.

rhvgoyal:
CC @shishir-a412ed

rhvgoyal:
bot, test pull request

samuelkarp:
@rhvgoyal Is there anywhere specific I should call out this new behavior in the man page?

rhvgoyal:
@samuelkarp I thought you will create a new option say "DOCKER_STORAGE_OPTS=" which can be specified in docker-storage-setup. But you seem to be using `get_docker_storage_options` functions which reads /etc/sysconfig/docker-storage. That's not what we are looking for. That's the output file we are generating?

samuelkarp:
Ah, I misunderstood what you were asking.  I will change it as you suggest.  (As a note, my existing change does work, you just set `--storage-driver devicemapper --storage-opt dm.fs=ext4` in `/etc/sysconfig/docker-storage` prior to running `docker-storage-setup`.  However, I think it's clearer with a separate variable in `/etc/sysconfig/docker-storage-setup`, so I will make that change.)

rhvgoyal:
/etc/sysconfig/docker-storage is the output file of docker-storage-setup. And /usr/lib/docker-storage-setup/docker-storage-setup is the input file. So we should allow user to specify the options in input which get generated in output file appropriately.

samuelkarp:
@rhvgoyal I've updated it per your comments.

rhvgoyal:
We need to modify man page too (docker-storage-setup.1)

samuelkarp:
I'm limiting it to `dm.fs` as that's the only one I really have time to test right now.  For other options, a more complete merging strategy is likely appropriate, but that can always be added in later.

`docker-storage-setup.1` is modified with the latest revision.

rhvgoyal:
You don't have to test everything. I think just test 2 together and if it works, it is reasonably safe to assume things are working. May be --storage-opt dm.fs=ext4 and --storage-dm.basesize=foo can be tested together. I don't want to limit it to dm.fs only. That defeats the purpose of introducing a generic option.

samuelkarp:
@rhvgoyal I've made it more generic.  Can you take a look?

samuelkarp:
@rhvgoyal Thanks for the review!  I've addressed your comments; can you take a look again?

rhvgoyal:
bot, add author to whitelist

samuelkarp:
@rhvgoyal Updated

rhvgoyal:
LGTM

rhvgoyal:
@shishir-a412ed @rhatdan @cgwalters I am looking for one more review and LGTM and then I will merge it.

rhatdan:
LGTM

PR #104, opened 2016-02-26
Allow the LV to be set to error when full

samuelkarp:
Related to #103, setting the `lv_when_full` behavior to `error` and pre-allocating the LV to use the whole VG seems to have better behavior when the VG is not used for other LVs.

### DIFF
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -71,6 +71,11 @@ CHUNK_SIZE:
       Controls the chunk size/block size of thin pool. Value of CHUNK_SIZE
       be suitable to be passed to --chunk-size option of lvconvert.
 
+LV_ERROR_WHEN_FULL
+      Controls the LV behavior when full. Default behavior is controlled by
+      lvm.conf. Setting LV_ERROR_WHEN_FULL=yes forces the LV to error when
+      writes are attempted rather than queuing.
+
 The options below should be specified as values acceptable to 'lvextend -L':
 
 ROOT_SIZE: The size to which the root filesystem should be grown.
--- a/docker-storage-setup.conf
+++ b/docker-storage-setup.conf
@@ -66,3 +66,7 @@ POOL_AUTOEXTEND_THRESHOLD=60
 
 # Extend the pool by specified percentage when threshold is hit.
 POOL_AUTOEXTEND_PERCENT=20
+
+# Error when full instead of queue. Defaults to "no", change to "yes" to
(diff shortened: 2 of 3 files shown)


rhvgoyal:
I am not sure how does this benefit. When pool gets full, default behavior of thin pool is to queue IO for 60 seconds as it is hoping that soon thin pool will be expanded. If not, then IO errors out after 60 seconds.

If you disable this queuing behavior and error starts returning immediately, how do we benefit?

samuelkarp:
This is beneficial when you have a fixed-size LV (i.e., setting `AUTO_EXTEND_POOL=no`) and don't want to wait for 60 seconds for the IO errors to timeout.

rhvgoyal:
XFS will still hang if pool is full (with AUTO_EXTEND_POOL=no). so how does it help?

samuelkarp:
Yes, but ext4 will not.  #103 allows us to set ext4.

rhvgoyal:
In case of ext4, pool will grow just fine and when pool is full ext4 will go to read only mode. So in case of ext4, there does not seem to be a reason to set AUTO_EXTEND_POOL=n

See, we can provide hundredes of knobs if we want in docker-storage-setup. maintaining and parsing all that will become very messy pretty soon. So I really want to focus on knobs which really matter and make a difference and not try to support every possible configuration/setting.

samuelkarp:
I'll run some tests with ext4 and these options unset.  I'm just not sure why I wouldn't want to size the LV up to the full available space and set `AUTO_EXTEND_POOL=no` if the whole PV and VG are dedicated to the single LV.

rhvgoyal:
I am not completely opposed to adding an option like that. Just want to make sure that it is really useful for people and there are more than one users/consumers.

samuelkarp:
I would imagine that anyone who is setting a fixed size for the thin pool would probably prefer immediate error rather than queuing for 60s.  With that said, would it make more sense to change the logic to just key from the value of `AUTO_EXTEND_POOL` rather than a separate environment variable?

rhvgoyal:
One might have AUTO_EXTEND_POOL=no, but I think thin pool can still be extended driven by global lvm policy. Look at /etc/lvm/lvm.conf  `thin_pool_autoextend_threshold`. So disabling queueing of IO just based on AUTO_EXTEND_POOL might not be best.

rhvgoyal:
Closing this for now. Open it again if need be.

PR #105, opened 2016-03-01
use lsblk to figure out if a device is partition or not

### LINKED ISSUE #99: Devices /dev/cciss/c0d1 detected as partition
Hi,

Specifying DEVS="/dev/cciss/c0d1" which is a correct device label fails with the following error:
"Partition specification unsupported at this time."

The following on line 426
```
    if [[ $dev =~ .*[0-9]$ ]]; then
       Fatal "Partition specification unsupported at this time."
    fi
```
Identifies any number as a partition which is false in the above case.

/dev/cciss/c0d1 is the device name of a logical drive on an HP DL 380 server using a RAID controller.

My suggestion would be to remove the test for the number as it is not sufficient, and creates false answers.

rhvgoyal:
Fixes #99 

Right now we have very crude logic to figure out if a device is partition
or not. It fails on some devices whose names end with a number. Use lsblk
instead to figure out if a device is partition or not.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -411,6 +411,21 @@ is_dev_part_of_vg() {
  return 1
 }
 
+is_block_dev_partition() {
+  local bdev=$1
+
+  [ ! -b "$bdev" ] && bdev="/dev/${bdev}"
+  if ! disktype=$(lsblk -n --nodeps --output type ${bdev}); then
+    Fatal "Failed to run lsblk on device $bdev"
+  fi
+
+  if [ "$disktype" == "part" ];then
+    return 0
+  fi
+
+  return 1
+}
+
 # Make sure passed in devices are valid block devies. Also make sure they
 # are not partitions.
 check_block_devs() {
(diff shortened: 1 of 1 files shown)


rhvgoyal:
CC @rhatdan @cgwalters @shishir-a412ed

cgwalters:
LGTM.

rhvgoyal:
Cool. Merging.

frankamp:
I tried to build a raid0 and pass it as a device, and it failed this check. Do you know if that would have worked otherwise? I used a command like:

`mdadm --create --verbose --auto=yes /dev/md0 --level=0 --raid-devices=2 /dev/xvdcz1 /dev/xvdd1`

After this completed I passed /dev/md0 into the container-storage-setup (actually docker-storage-setup).

It failed on this check

`INFO: EXTRA_DOCKER_STORAGE_OPTIONS is deprecated, please use EXTRA_STORAGE_OPTIONS
ERROR: Partition specification unsupported at this time.`

Thoughts?

PR #110, opened 2016-03-23
Introduce a helper to wait for a device

rhvgoyal:
Introduce an helper to wait for a device. Also make wait time configurable
using knob DEVICE_WAIT_TIMEOUT. Default value is 60 seconds. One can disable
wait by specifying DEVICE_WAIT_TIMEOUT=0.

### DIFF
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -75,6 +75,14 @@ CHUNK_SIZE:
       Controls the chunk size/block size of thin pool. Value of CHUNK_SIZE
       be suitable to be passed to --chunk-size option of lvconvert.
 
+DEVICE_WAIT_TIMEOUT:
+           Specifies a device wait timeout value in seconds. In
+           certain cases required devices might not be immediately
+           available and docker-storage-setup might decide to wait for
+           it. This timeout specifies how long one should wait for the
+           device. Default is 60 seconds. One can disable the wait by
+           specifying a value of 0.
+
 The options below should be specified as values acceptable to 'lvextend -L':
 
 ROOT_SIZE: The size to which the root filesystem should be grown.
--- a/docker-storage-setup.conf
+++ b/docker-storage-setup.conf
@@ -71,3 +71,8 @@ POOL_AUTOEXTEND_THRESHOLD=60
 
 # Extend the pool by specified percentage when threshold is hit.
 POOL_AUTOEXTEND_PERCENT=20
+
+# Device wait timeout in seconds. This is generic timeout which can be used by
+# docker storage setup service to wait on various kind of block devices.
+# Setting a value of 0 can disable this wait.
+DEVICE_WAIT_TIMEOUT=60
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -118,6 +118,33 @@ extra_options_has_dm_fs() {
   return 1
 }
 
+# Wait for a device for certain time interval. If device is found 0 is
+# returned otherwise 1.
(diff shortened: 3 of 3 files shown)


rhvgoyal:
cc @thoraxe

rhvgoyal:
Erik Jacobs is facing some docker-storage-setup issues once in a while with virtual machine in amazon. Especially if a block device is passed in for thin pool say /dev/xvdb, then sometimes docker-storage-setup exits without creating thin pool.

Some debugging has revealed that after creating partition /dev/xvdb1, we try to do "pvcreate /dev/xvdb1" and that fails. It fails because node /dev/xvdb1 is not available yet.

So basically there is a race condition where partition is not immediately available. I suspect it might be happening due to extremely slow disk sometimes. May be udev rule creating the device node is stuck on reading something from disk and that's why node creation gets delayed.

This patch introduces a facility to wait for the device upto certain timeout before failure is hit.

@thoraxe is testing it.

thoraxe:
Can you explain what the expected behavior is if the disk is not ready? Does it simply try again later or does it do something else?

How would I know it is "working"?

cgwalters:
Instead of `sleep`, can we try using `udevadm monitor`?  I'm not finding a shell script example of this offhand, though I did find http://stackoverflow.com/questions/20355616/how-to-wait-in-script-until-device-is-connected

But going that route is turning more towards an event-driven/daemon model which is to say the least a large architectural change for d-s-s, so I don't want to block this patch on that necessarily.  Maybe just reduce the timeout to `1`?

rhvgoyal:
@thoraxe Check the logs after successful run. If you see following message in the logs, then you know that patch worked for you.

`Waiting for device <device-name> to be available. Wait time remaining is <timeout> seconds`

rhvgoyal:
@cgwalters 

Trying to use event driven model makes it more complicated. So sleep and wait is atleast a simple start. So when you say 1 second, do you mean total wait or you mean sleep for 1 second and check if device is available or not. Total 1 second might be very less. I have seen bigger timeouts in systemd and dracut when they wait for devices. 

systemd also has the notion of waiting on .device unit files and one could create dependencies on these unit files for startup.

https://lists.fedoraproject.org/pipermail/devel/2012-January/160917.html

One problem with this approach is that in our case dependencies are dynamic. That is device name is configurable and hence every time we will have to first determine the foo.device dependency and override docker-storage-setup dependencies. That possibly means splitting docker-storage-setup in two parts. First part just determines the dependencies and sets up second part to be dependent on those dependencies.  Sounds fairly twisted. So for now a simple wait might be good enough.

Having said that, I am seeing other bug reports where people have setup thin pool on a separate disk and somehow that disk is slow to come up and we should wait for /dev/foo and thin pool to come up.

I am thinking that for the devices in DEVS= directive, I will ask user to manually edit /etc/systemd/system/docker-storage-setup file and create following dependencies.

Wants=dev-vdb.device
After=dev-vdb.device

And for waiting on thin pool (if it was configured in /etc/sysconfig/docker-storage), then do a sleep wait for certain time interval.

cgwalters:
Yep, agreed with all of that, I don't think we should block on a rearchitecting.  I'm :+1: on the patch as is.

rhvgoyal:
Refreshed patch. Fixed a typo in man page.

rhvgoyal:
Ok, I am merging this PR. There is another one blocked behind it where I need to wait for thin pool to come up.

cgwalters:
@cgwalters-bot r+ 6239762

PR #111, opened 2016-03-29
[merged] Wait for thin pool

rhvgoyal:
There are situations where it takes a while for thin pool to come up. In that case docker-storage-setup can exit pretty soon and docker fails saying thin pool is not there.

So if a thin pool is configured in /etc/sysconfig/docker-storage, wait for that thin pool to come up. That will block docker too till thin pool comes up.

We just committed the patches to wait for devices and this patch series re-uses it. So default wait time continues to be 60 seconds.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -61,6 +61,9 @@ META_LV_NAME="${POOL_LV_NAME}meta"
 DOCKER_STORAGE="/etc/sysconfig/docker-storage"
 STORAGE_DRIVERS="devicemapper overlay"
 
+# Will have currently configured storage options in $DOCKER_STORAGE
+CURRENT_STORAGE_OPTIONS=""
+
 get_docker_version() {
   local version
 
@@ -611,6 +614,11 @@ disable_auto_pool_extension() {
 # Gets the current DOCKER_STORAGE_OPTIONS= string.
 get_docker_storage_options() {
   local options
+
+  if [ ! -f "$DOCKER_STORAGE" ];then
+    return 0
+  fi
+
   if options=$(grep -e "^DOCKER_STORAGE_OPTIONS=" $DOCKER_STORAGE | sed 's/DOCKER_STORAGE_OPTIONS=//' | sed 's/^ *//' | sed 's/^"//' | sed 's/"$//');then
     echo $options
     return 0
@@ -633,15 +641,8 @@ is_valid_storage_driver() {
 get_existing_storage_driver() {
   local options driver
 
-  if [ ! -f "$DOCKER_STORAGE" ];then
-    return 0
-  fi
+  options=$CURRENT_STORAGE_OPTIONS
 
-  if ! options=$(get_docker_storage_options); then
-    return 1
-  fi
-
-  # DOCKER_STORAGE_OPTIONS= is empty there is no storage driver configured yet.
   [ -z "$options" ] && return 0
 
   # Check if -storage-driver <driver> is there.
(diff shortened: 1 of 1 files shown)


rhvgoyal:
cc @cgwalters

docker-storage-setup.sh
```diff
@@ -346,11 +346,58 @@ create_lvm_thin_pool () {
   lvconvert -y --zero n $CHUNK_SIZE_ARG --thinpool $VG/$DATA_LV_NAME --poolmetadata $VG/$META_LV_NAME
 }
 
+get_configured_thin_pool() {
+  local options tpool opt
+
+  if [ ! -f "$DOCKER_STORAGE" ];then
```
cgwalters: Hoist this into `get_docker_storage_options`?

docker-storage-setup.sh
```diff
@@ -346,11 +346,58 @@ create_lvm_thin_pool () {
...
+
+  if ! options=$(get_docker_storage_options); then
+    return 1
+  fi
+
+  # DOCKER_STORAGE_OPTIONS= is empty there is no storage driver configured yet.
+  [ -z "$options" ] && return 0
+
+  # No thin pool configured
+  if ! grep -q -e "--storage-opt dm.thinpooldev=" <<< $options;then
```
cgwalters: `get_existing_storage_driver` does something similar...can we reuse that?

cgwalters: Looks like this code is currently only called if `[ "$STORAGE_DRIVER" == "devicemapper" ]` so maybe just drop this hunk?

docker-storage-setup.sh
```diff
@@ -346,11 +346,58 @@ create_lvm_thin_pool () {
...
+	return 0
+  fi
+
+  # This assumes that thin pool is specified as dm.thinpooldev=foo. There
+  # are no spaces in between.
+  for opt in $options; do
+    if [[ $opt =~ dm.thinpooldev* ]];then
+      tpool=`cut -d "=" -f2 <<< $opt`
```
cgwalters: tpool=${tpool#*=} ?

cgwalters:
Some minor style comments, otherwise looks reasonable.

docker-storage-setup.sh
rhvgoyal: Ok, will drop grep dm.thinpooldev check.

I can't reuse get_existing_storage_driver() as it is but I took some shared code out into common area. That is querying for storage options. Now we query storage options once and save in a variable and both the functions use this variable.

docker-storage-setup.sh
rhvgoyal: thanks. this looks better.

docker-storage-setup.sh
rhvgoyal: done. In fact I moved get_docker_storage_options() call into setup_storage() so that we call it once and then multiple functions can look at existing storage options.

rhvgoyal:
@cgwalters Updated the patches. PTAL.

docker-storage-setup.sh
```diff
@@ -633,15 +641,8 @@ is_valid_storage_driver() {
...
   local options driver
 
-  if [ ! -f "$DOCKER_STORAGE" ];then
-    return 0
-  fi
-
-  if ! options=$(get_docker_storage_options); then
-    return 1
-  fi
+  options=$CURRENT_STORAGE_OPTIONS
```
cgwalters: This feels weird as if we have global state we could now just always reference it.  But not necessarily worth redoing the patch again.

rhvgoyal: Hmm..., not sure what you did not like. Did you not like assigning CURRENT_STORAGE_OPTIONS to local variable or you did not like querying storage options and saving it in a global variable.

rhvgoyal:
I think querying the existing storage options once and saving in a global variable makes sense. Now I don't have to query it twice and both get_existing_storage_driver() and get_configured_thin_pool() can parse this string.

PR #115, opened 2016-04-12
WIP - Status output

mvollmer:
This is intended for driving a UI and provides information about
requested configuration vs actual configuration.

### DIFF
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -10,6 +10,9 @@ docker\-storage\-setup - Grows the root filesystem and sets up storage for docke
 \f[B]-h, --help\f[]
   Print usage statement
 
+\f[B]--machine-readable-status\f[]
+  Show current status in machine-readable form, and then exit
+
 .SH EXAMPLES
 Run \f[B]docker-storage-setup\f[] after setting up your configuration in
 /etc/sysconfig/docker-storage-setup. One can look at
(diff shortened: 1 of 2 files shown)


mvollmer:
Docs, tests, and human readable output are still missing.

mvollmer:
Docs added.  I'll wait for some feedback before taking this any further.

mvollmer:
Does this make sense with the "overlay" storage driver?

rhvgoyal:
I think it will be more useful to output status in the format that a human can read first. Then tools can either parse that or we could extend same to a machine readable format using some sort of switch.

rhvgoyal:
At higher level when it comes to status/info, I would think of following items.
- Storage Driver

And then details specific to that storage driver. For devicemapper storage driver, following additional 
details should be helpful.

- Thin Pool Name and some properties like "state", size etc.

Being able to provide more details about devices seems low level. In first step, I think it should be
enough to just list the devices specified in DEVS?

mvollmer:
> I think it will be more useful to output status in the format that a human can read first. Then tools can either parse that or we could extend same to a machine readable format using some sort of switch.

Without machine readable output, this change would not be useful to me, so we need to include that from the get go.

Thanks for the suggestions, I'll work on adding humane friendly output.

(Do you have a official way to mark this as 'work in process'?)

rhvgoyal:
If storage driver is overlay, we should supress rest of the output. Something like "docker info" Where every storage driver gets to output its own status. So there are some generic fields and there are some driver specific fields.

mvollmer:
@rhvgoyal, do you still think we need human readable output here?  Or should that be done via `atomic storage`?

mvollmer:
I think goes to far, and I am not really working on this anymore.

PR #117, opened 2016-04-12
[merged] Do not create a new thin pool if docker is already having devicemapper metadata

### LINKED ISSUE #116: Running docker-storage-setup after docker starts breaks the configuration
On Fedora, RHEL and CentOS systems - when launched by systemd - docker is configured to first run  docker-storage-setup.service. If docker-storage-setup completes without success (such as because insufficient unused space is available on the system), docker will go ahead and initialize itself with loop devices.

However, if the space becomes available later (such as another disk being added to the volume group), the next time that docker-storage-setup.service is run will create the thin pool and attempt to configure docker to use it. This breaks docker and causes it to report errors on startup such as:
```
Error starting daemon: error initializing graphdriver: devmapper: Base Device UUID and Fileystem verification failed.devicemapper: Error running deviceCreate (ActivateDevice) dm_task_run failed
```
This buggy situation is tricky to get out of.

I discussed this with @rhvgoyal on IRC today and suggested the following:
When docker-storage-setup starts, it should attempt to read in the metadata from /var/lib/docker. If it exists and is pointing at the loop devices, then docker-storage-setup should print a warning to the logs advising the user of how to properly reset the system to use the preferred storage mechanism and then exit with success (not changing the storage configuration at all).

This will prevent docker-storage-setup from breaking docker and will provide the user with the information necessary to convert their system over to the better storage layout.

This approach will probably have relevance to https://github.com/projectatomic/docker-storage-setup/issues/62

rhvgoyal:
Fixes #116 

If docker already has devicemapper metadata dir, do not create a new thin
pool. This most likely means that docker was already started and it used
some other thin pool (most likely using loop devices) and now we are setting
up a new thin pool and asking docker to use it. This will not work. Old
metadata of docker will fail to work with new thin pool.

This is a common situation and lot of people have complained about running
into this. Right now docker message is  that base device UUID verification
failed but that does not tell user much about what to do.

So detect this in docker-storage-setup and error out and display a message
which explains how one can do complete cleanup. This will destroy all 
existing images and containers and all docker metadata.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -61,6 +61,8 @@ META_LV_NAME="${POOL_LV_NAME}meta"
 DOCKER_STORAGE="/etc/sysconfig/docker-storage"
 STORAGE_DRIVERS="devicemapper overlay"
 
+DOCKER_METADATA_DIR="/var/lib/docker"
+
 # Will have currently configured storage options in $DOCKER_STORAGE
 CURRENT_STORAGE_OPTIONS=""
 
@@ -366,6 +368,17 @@ get_configured_thin_pool() {
   done
 }
 
+check_docker_storage_metadata() {
+  local docker_devmapper_meta_dir="$DOCKER_METADATA_DIR/devicemapper/metadata/"
+
+  [ ! -d "$docker_devmapper_meta_dir" ] && return 0
+
+  # Docker seems to be already using devicemapper storage driver. Error out.
+  Error "Docker has been previously configured for use with devicemapper graph driver. Not creating a new thin pool as existing docker metadata will fail to work with it. Manual cleanup is required before this will succeed."
+  Info "Docker state can be reset by stopping docker and by removing ${DOCKER_METADATA_DIR} directory. This will destroy existing docker images and containers and all the docker metadata."
+  exit 1
+}
+
 setup_lvm_thin_pool () {
   local tpool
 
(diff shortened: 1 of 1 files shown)


rhvgoyal:
@sgallagher Does this look good enough? I am not looking for loop devices as such. If you like that better, I can make this loop device specific check.

docker-storage-setup.sh
```diff
@@ -366,6 +368,19 @@ get_configured_thin_pool() {
...
 }
 
+check_docker_storage_metadata() {
+  local docker_devmapper_meta_dir="$DOCKER_METADATA_DIR/devicemapper/metadata/"
+
+  [ ! -d "$docker_devmapper_meta_dir" ] && return 0
+
+  # Docker seems to be already using devicemapper storage driver. Error out.
+  Error "Docker devicemapper metadata directory $docker_devmapper_meta_dir is present."
+  Error "It seems docker is already having devicemapper metadata for a specific thin pool. Not creating a new thin pool as existing docker metadata will fail to work with new thin pool. Cleanup existing docker metadata and retry."
```
sgallagher: Since this is a user-visible error, I'd suggest cleaning up the language:
"Docker has been previously configured for use with devicemapper metadata. Not creating a new thin pool as existing docker metadata will fail to work with it. Manual cleanup is required before this will succeed."

rhvgoyal: @sgallagher sounds good. Will change.

rhvgoyal:
@sgallagher Updated patch. PTAL.

rhvgoyal:
CC @cgwalters

sgallagher:
LGTM

rhatdan:
LGTM

rhvgoyal:
@cgwalters-bot r+

PR #118, opened 2016-04-12
Test: Cleanup even after test failure

rhvgoyal:
If test fails, right now we are not doing cleanup. This leaves some of 
the state behind and next run of test fails. Cleanup even if test fails.


jlebon:
LGTM!

PR #119, opened 2016-04-13
[merged] Recreate missing /etc/sysconfig/docker-storage file

rhvgoyal:
Some users complained that if they delete /etc/sysconfig/docker-storage file
without deleting lvm thin pool, then they are forced to do full cleanup. That
is delete lvm thin pool, delete /var/lib/docker and restart docker.

This might not be necessarily required. If thin pool is still around, then
we should be able to recreate /etc/sysconfig/docker-storage. 

So this patch introduces this change. If thin pool is around but /etc/sysconfig/docker-storage is missing, recreate it.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -405,6 +405,13 @@ setup_lvm_thin_pool () {
     check_docker_storage_metadata
     create_lvm_thin_pool
     write_storage_config_file
+  else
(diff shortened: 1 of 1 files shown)


rhvgoyal:
cc @rhatdan @cgwalters

qiancai:
LGTM

rhatdan:
LGTM

rhvgoyal:
@cgwalters-bot r+

PR #120, opened 2016-04-15
Wipe newly created partitions

mvollmer:
This avoids a bogus prompt by pvcreate which would make non-interactive runs fail, such as during boot.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -578,6 +578,11 @@ EOF
     if ! wait_for_dev ${dev}1; then
       Fatal "Partition device ${dev}1 is not available"
     fi
+    # Sometimes there is still a old filesystem signature where we
+    # have just created our new partition.  Pvcreate will find it and
+    # ask for confirmation to overwrite it.  So we wipe the new
+    # partition first.
+    wipefs -a ${dev}1
     pvcreate ${dev}1
     PVS="$PVS ${dev}1"
   done
--- a/tests/002-no-pvcreate-prompts-for-stale-fs-signatures
+++ b/tests/002-no-pvcreate-prompts-for-stale-fs-signatures
@@ -1,0 +1,66 @@
+# Checks that stale filesystem signatures on one of the devices do not
+# cause pvcreate to prompt.
+
+test_stale_fs_signature() {
+  local devs=$DEVS
+  local test_status
+  local testname=`basename "$0"`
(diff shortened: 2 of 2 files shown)


rhatdan:
bot, add author to whitelist

rhvgoyal:
Prompt by pvcreate is not bogus. In non-interactive mode it defaults to not use the disk if it finds other signatures on disk. I think this is the safe default. We don't want to be wiping a user's data if user passed the disk to dss accidently.

docker-storage-setup.sh
```diff
@@ -578,6 +578,11 @@ EOF
     if ! wait_for_dev ${dev}1; then
       Fatal "Partition device ${dev}1 is not available"
     fi
+    # Sometimes there is still a old filesystem signature where we
+    # have just created our new partition.  Pvcreate will find it and
+    # ask for confirmation to overwrite it.  So we wipe the new
+    # partition first.
+    wipefs -a ${dev}1
```
rhvgoyal: I don't think we should be wiping signatures by default. If there is data, one can still delete the partition just created and recover the data. 

So I like the idea of bailing out if disk has fs signatures present and user can run wipefs before passing that disk to dss.

Alternatively, we can implement a config option for dss, say WIPE_SIGNATURES_DISK and if this is set to true, we will always first wipe all signatures on disk before further processing. It will be off by default.

mvollmer:
(The new test didn't run. I'll check.)

docker-storage-setup.sh
mvollmer: > So I like the idea of bailing out if disk has fs signatures present and user can run wipefs before passing that disk to dss.

For me, the confusing bit is that the signature is on a partition that d-s-s has just created itself.  The disk didn't have any partition table on it when I gave it to d-s-s.

It appears to me to be quite tricky to know in what way d-s-s wants its disks to be wiped if it reserves the right to reject a disk as soon as it finds a tiny hint of old content.

mvollmer:
> Prompt by pvcreate is not bogus.

Fair enough.

docker-storage-setup.sh
rhvgoyal: Actually, I think we should check for signatures on disk before we create partition. Otherwise, pvcreate will reject it anyaway later. That way dss behavior will be more consistent.

PR #122, opened 2016-04-19
[merged] Wipe Signature Related Cleanup

### LINKED ISSUE #121: docker-storage-setup destroys existing physical volumes
Docker-storage-setup will happily call sfdisk on a block device that is currently in use as a LVM2 physical volume, thereby potentially destroying any logical volume that is stored on that device.

Transcript:
```
[root@f24 atomic-storaged]# pvcreate /dev/sdb
  Physical volume "/dev/sdb" successfully created
[root@f24 atomic-storaged]# vgcreate TEST /dev/sdb
  Volume group "TEST" successfully created
[root@f24 atomic-storaged]# lvcreate -l 100%FREE TEST -n foo
  Logical volume "foo" created.
[root@f24 atomic-storaged]# cat /etc/sysconfig/docker-storage-setup 
DEVS="/dev/sdb"
[root@f24 atomic-storaged]# docker-storage-setup 
Checking that no-one is using this disk right now ... OK

Device /dev/sdb already contains a LVM2_member signature.
It is strongly recommended to wipe the device with wipefs(8), in order to avoid possible collisions.

Disk /dev/sdb: 2 GiB, 2147483648 bytes, 4194304 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes

>>> Script header accepted.
>>> Created a new DOS disklabel with disk identifier 0x9165a0a8.
Created a new partition 1 of type 'Linux LVM' and of size 2 GiB.
/dev/sdb2: 
New situation:

Device     Boot Start     End Sectors Size Id Type
/dev/sdb1        2048 4194303 4192256   2G 8e Linux LVM

The partition table has been altered.
Calling ioctl() to re-read partition table.
Syncing disks.
  Can't open /dev/sdb1 exclusively.  Mounted filesystem?
[root@f24 atomic-storaged]# echo $?
5
[root@f24 atomic-storaged]# vgs TEST
  Volume group "TEST" not found
  Cannot process volume group TEST
[root@f24 atomic-storaged]# dmsetup table
fedora-swap: 0 1679360 linear 252:2 2048
fedora-root: 0 14065664 linear 252:2 1681408
TEST-foo: 0 4186112 linear 8:16 2048
```

rhvgoyal:
Fixes #121 

This patch set does few things.

- Patch 1 and Patch2 are simple code reorganization and cleanup. No functionality change.

-  Reject a disk if it already contains signatures like LVM. Right now only exception is "dos" signatures. We allow that. And reason being that in the past we have allowed that. A disk might have dos signature without having any partition and we have accepted it. Don't want to create a regression scenario. We check for partitions anyway, so we know that disk is not partitioned.

-  Use pvcreate with option -f all the time. This means that if when we scan the disk for signatures, and if no objectionable signatures are there, then we own the disk and we will not worry about signatures which can become in visible later in the middle of disk. For example, after creating a partition if we
find a left over signature, we will not abort processing. By now we own the disk.

- Implement an option OVERRIDE_SIGNATURES which will allow one to wipe signature of disk even if one is present. By default it is set to false and that means we will reject disk if it contains a valid signature (except dos one).

### DIFF
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -83,6 +83,14 @@ DEVICE_WAIT_TIMEOUT:
            device. Default is 60 seconds. One can disable the wait by
            specifying a value of 0.
 
+WIPE_SIGNATURES:
+      Wipe any signatures found on disk. Valid values are
+      true/false and default value is false. By default if any
+      signatures are found on disk operation is aborted. If this value
+      is set to true, then signatures will either be wiped or
+      overwritten as suitable. This also means that if there is any
+      data on disk, it will be lost.
+
 The options below should be specified as values acceptable to 'lvextend -L':
 
 ROOT_SIZE: The size to which the root filesystem should be grown.
--- a/docker-storage-setup.conf
+++ b/docker-storage-setup.conf
@@ -76,3 +76,11 @@ POOL_AUTOEXTEND_PERCENT=20
 # docker storage setup service to wait on various kind of block devices.
 # Setting a value of 0 can disable this wait.
 DEVICE_WAIT_TIMEOUT=60
+
+# Wipe any signatures (partition, filesystem, lvm etc) found on disk.
+# This could mean wiping the signature explicitly or using force options
+# of various commands to wipe/overwrite signatures. By default signatures
+# are not wiped and user needs to wipe these. One can change default behavior
+# by setting WIPE_SIGNATURES=true. Be careful before using this option
+# as this means if there was any leftover data on disk, it will be lost.
+WIPE_SIGNATURES=false
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -537,6 +537,24 @@ is_block_dev_partition() {
   return 1
 }
 
+check_block_dev_sig() {
+  local bdev=$1
+  local sig
+
+  [ ! -b "$bdev" ] && bdev="/dev/${bdev}"
+
+  if ! sig=$(wipefs -p $bdev); then
+    Fatal "Failed to check signatures on device $bdev"
+  fi
+
+  [ "$sig" == "" ] && return 0
+
+  while IFS=, read offset uuid label type; do
+    [ "$offset" == "# offset" ] && continue
+    Fatal "Found $type signature on device ${bdev} at offset ${offset}. Wipe signatures using wipefs and retry."
+  done <<< "$sig"
+}
+
 # Make sure passed in devices are valid block devies. Also make sure they
 # are not partitions.
 check_block_devs() {
@@ -852,6 +870,9 @@ if [[ -n "$DEVS" && -n "$VG" ]]; then
   # nothing more to do
   P=$(scan_disks)
   if [[ -n "$P" ]]; then
+    for dev in $P; do
+      check_block_dev_sig $dev
+    done
     create_disk_partitions "$P"
     create_extend_volume_group
   fi
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -537,7 +537,7 @@ is_block_dev_partition() {
   return 1
 }
 
-check_block_dev_sig() {
+check_wipe_block_dev_sig() {
   local bdev=$1
   local sig
 
@@ -549,9 +549,17 @@ check_block_dev_sig() {
 
   [ "$sig" == "" ] && return 0
 
+  if [ "$WIPE_SIGNATURES" == "true" ];then
+    Info "Wipe Signatures is set to true. Any signatures on $bdev will be wiped."
+    if ! wipefs -a $bdev; then
+      Fatal "Failed to wipe signatures on device $bdev"
+    fi
+    return 0
+  fi
+
   while IFS=, read offset uuid label type; do
     [ "$offset" == "# offset" ] && continue
-    Fatal "Found $type signature on device ${bdev} at offset ${offset}. Wipe signatures using wipefs and retry."
+    Fatal "Found $type signature on device ${bdev} at offset ${offset}. Wipe signatures using wipefs or use WIPE_SIGNATURES=true and retry."
   done <<< "$sig"
 }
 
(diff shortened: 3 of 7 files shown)


rhvgoyal:
cc @rhatdan @mvollmer @cgwalters

rhvgoyal:
cc @jlebon

rhvgoyal:
I am adding few tests which ensure that disk does not have any signatures on it and require clean disk. We will need to clean all signatures on disk before giving it to tests.

wipefs -a /dev/sdb

rhvgoyal:
bot, test

rhvgoyal:
bot, retest

jlebon:
@rhvgoyal try "bot, retest this please"

rhvgoyal:
bot, retest this please

jlebon:
The bot has a minimal courtesy level trigger

rhvgoyal:
@jlebon :-)

docker-storage-setup.1
```diff
@@ -83,6 +83,13 @@ DEVICE_WAIT_TIMEOUT:
            device. Default is 60 seconds. One can disable the wait by
            specifying a value of 0.
 
+OVERRIDE_SIGNATURES:
```
mvollmer: What about "OVERWRITE_SIGNATURES" or "WIPE_SIGNATURES"?  "Override" implies to me that the signature would ordinarily be used for something, unless this option is set.

docker-storage-setup.conf
```diff
@@ -76,3 +76,11 @@ POOL_AUTOEXTEND_PERCENT=20
 # docker storage setup service to wait on various kind of block devices.
 # Setting a value of 0 can disable this wait.
 DEVICE_WAIT_TIMEOUT=60
+
+# Override any signatures (partition, filesystem, lvm etc) found on disk.
+# This could mean wiping the signature explicitly # or using force options
```
mvollmer: Stray "#" in the middle of the line.

mvollmer:
The commit message of "Check block device signature before creating partition" seems to be truncated.

mvollmer:
Nice, thanks!

docker-storage-setup.1
rhvgoyal: Agreed. I guess I will use WIPE_SIGNATURES. That probably is most intuitive.

docker-storage-setup.conf
rhvgoyal: Will fix it.

rhvgoyal:
Will fix the truncated commit message of one of the patch. Thanks for reviewing.

rhvgoyal:
@mvollmer Took care of your comments. PTAL.

rhvgoyal:
@rhatdan  @cgwalters Can I get some review or LGTM on this.

docker-storage-setup.sh
```diff
@@ -537,6 +537,32 @@ is_block_dev_partition() {
...
+  [ "$sig" == "" ] && return 0
+
+  while IFS=, read offset uuid label type; do
+    [ "$offset" == "# offset" ] && continue
+
+    # Ignore dos signatures for now. We have already checked that disk
+    # does not have partitions. But there might be leftover partition
+    # signature and so far we have not forced user to wipe it. So in an
+    # attempt to not change user visible behavior, don't force user to
+    # wipe dos signature and just make sure disk is not partitioned.
```
cgwalters: Did the dos issue actually come up in practice?

rhvgoyal: Actual user did not experience it. I faced it in my testing.

If I delete a partition from disk using fdisk or parted, dos signatures are still present and dss still fails. I am fairly certain that some users will complain about it.

rhvgoyal: Actual users will not see this issue till I push this code and disallow even dos signatures and users will be forced to use "wipefs -a disk" before that disk can be used with dss. Not sure how many usres are used to doing that. I certainly am not used to doing it. Once partitions have been removed using fdisk, I considered disk clean.

tests/003-test-override-signature-wipes-existing-signatures.sh
```diff
@@ -0,0 +1,65 @@
+cleanup() {
+  local vg_name=$1
+  local devs=$2
+
+  vgremove -y $vg_name >> $LOGS 2>&1
+  remove_pvs "$devs"
+  remove_partitions "$devs"
+  clean_config_files
+  wipe_signatures "$devs"
+}
```
cgwalters: At this point should consider a `libtest.sh` like we have in ostree: https://github.com/ostreedev/ostree/blob/master/tests/libtest.sh

The individual tests source it for funcitons.

docker-storage-setup.sh
cgwalters: It sounds like the special case isn't `dos`, it's "partition tables" which also include `gpt`, right?

That said...I think we *should* require people to understand how to use `wipefs` on their own rather than doing it specially for them just for disks with empty (but initialized) partition tables.

rhvgoyal: Right, this should apply to gpt partition table as well. 

So looks like you like the other approach better where if any kind of signatures are present on disk, reject it? 

I am fine with it. I am only concerned that some people will call it regression. They will say that previous version of dss accepted disks with dos/gpt signatures but this new version does not. 

I guess this should not create issues with atomic upgrade, as thin pool is already in place and new disks are not being added. So only thing we have to deal with is telling people to use wipefs with new version of docker-storage-setup and handle the surprise factor.

I can make that change if you like it better.

tests/003-test-override-signature-wipes-existing-signatures.sh
rhvgoyal: ok, will look into adding one.

docker-storage-setup.sh
cgwalters: I am at the moment not able to think of a realistic scenario where a user would be commonly passing empty-partitioned disks to d-s-s.  So yeah, I'd vote for not special casing `dos/gpt`.

rhvgoyal: Ok, lets go with that. If there is lot of hue and cry, we can always add a patch to special case this. I was not very happy with special casing either.

rhvgoyal:
@cgwalters Updated patches. Have taken care of your comments. PTAL.

cgwalters:
Still seems like there's a fair amount of copy-paste between tests, but it's OK, we can address that later.

@cgwalters-bot r+ c576a20

PR #123, opened 2016-04-25
[merged] Create a test library

rhvgoyal:
This PR does following.

- Give execute permission to existing test.
- Move common functions used by tests in libtest.sh and tests will source that file.
- Run tests in separate shell instead of sourcing these into parent shell.


cgwalters:
@cgwalters-bot r+ 8c74378

cgwalters:
@cgwalters-bot r+ 8c74e78

PR #124, opened 2016-04-26
[merged] Do not use pvcreate -f option

rhvgoyal:
Do not use pvcreate -f option instead wipe signatures explicitly on the
partition device. Reason being that -f is not very well defined and 
tomorrow it can be overloaded to do more things. So it is safer to wipe
signatures explicitly and use pvcreate without -f.

At this point we own the disk. If there are any signatures on disk to
being with, we will not be here until and unless user asked to wipe
signatures using WIPE_SIGNATURES=true. We will not care for any signatures
found on disk after creating partition.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -651,10 +651,13 @@ create_disk_partitions() {
     create_partition $dev
 
     # By now we have ownership of disk and we have checked there are no
-    # signatures on disk or signatures have been overridden. Don't care
+    # signatures on disk or signatures should be wiped. Don't care
     # about any signatures found in the middle of disk after creating
-    # partition and use force option.
-    pvcreate -f ${dev}1
+    # partition and wipe signatures if any are found.
+    if ! wipefs -a ${dev}1; then
+      Fatal "Failed to wipe signatures on device ${dev}1"
+    fi
+    pvcreate ${dev}1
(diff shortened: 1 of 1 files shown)


cgwalters:
The test failure looks something like wipefs is destroying the partition table (as expected), but we still try to clean up VGs/partitions that existed before?

rhvgoyal:
Looks like wipefs could not find device /dev/vdb1 after partition creation.

wipefs: error: /dev/vdb1: probing initialization failed: No such file or directory

That does not make much sense. We wait for partition device in a loop before we continue with rest of the commands. I will push another debug PR. This test seems to be passing on my local machine.

rhvgoyal:
DO NOT MERGE yet. This is debug version of patch.

rhvgoyal:
Ok, problem seems to be that DEVICE_WAIT_TIMEOUT is not defined in testing environment. This is defined in /lib/docker-storage-setup/docker-storage-setup. This is a new option. That means image we are testing on, does not have new enough docker-storage-setup. And that means due to udev races, we can run into this.

rhvgoyal:
@cgwalters PTAL.

I have pushed the patch again and this time it passes because "udevadm settle" makes sure partition device is there before wipefs is called on it.

cgwalters:
@cgwalters-bot r+ 9418463

PR #125, opened 2016-04-26
[merged] [WIP] Output more Info messages in wait_device()

rhvgoyal:
We are facing quite a few issues w.r.t block device not being available before pvcreate is called or wipefs is called. wait_device() should have done its work. Output more info messages so that things can be proven one way or other.


rhatdan:
LGTM

rhatdan:
@cgwalters-bot r+ 9b19ad8

PR #128, opened 2016-05-03
[merged] Add reset function to cleanup docker-storage-setup configuration

rhatdan:
This script should cleanup devicemapper storage as well as
remove /etc/sysconfig/docker-storage

### DIFF
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -7,8 +7,15 @@ docker\-storage\-setup - Grows the root filesystem and sets up storage for docke
 \f[B]docker-storage-setup\f[] [OPTIONS]
 .SH OPTIONS
 .PP
-\f[B]-h, --help\f[]
+\f[B]--help\f[]
   Print usage statement
+
+\f[B]--reset\f[]
+  Reset your docker storage to init state. Reset does not try to remove volume groups or try to remove any of the disks added previously.
+  
+Note: The
+\f[B]--reset\f[]
+command is not sufficient to cleanup your docker environment.  This option is used by other tools. (\f[B]atomic storage reset\f[])
 
 .SH EXAMPLES
 Run \f[B]docker-storage-setup\f[] after setting up your configuration in
@@ -129,6 +136,9 @@ DEVS=/dev/vdb
 DATA_SIZE=8GB
 
 .fi
+
+.SH "SEE ALSO"
+.BR atomic "(1)"
 
 .SH HISTORY
 
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -390,6 +390,13 @@ check_docker_storage_metadata() {
   exit 1
 }
 
+reset_lvm_thin_pool () {
+  if lvm_pool_exists; then
+      lvchange -an $VG/${POOL_LV_NAME}
+      lvremove $VG/${POOL_LV_NAME}
+  fi
+}
+
 setup_lvm_thin_pool () {
   local tpool
 
@@ -834,23 +841,24 @@ setup_storage() {
   fi
 }
 
+reset_storage() {
+    if [ "$STORAGE_DRIVER" == "devicemapper" ]; then
+	reset_lvm_thin_pool
+    fi
+    rm -f $DOCKER_STORAGE
+}
+
 usage() {
-  cat >&2 <<-FOE
+  cat <<-FOE
     Usage: $1 [OPTIONS]
 
     Grows the root filesystem and sets up storage for docker
 
     Options:
-      -h, --help    Print help message
+      --help    Print help message
+      --reset   Reset your docker storage to init state. 
 FOE
 }
-
-# Main Script
-
-if [ $# -gt 0 ]; then
-  usage $(basename $0)
-  exit 0
-fi
 
 # Source library. If there is a library present in same dir as d-s-s, source
 # that otherwise fall back to standard library. This is useful when modifyin
@@ -897,6 +905,21 @@ else
   if vg_exists "$VG";then
     VG_EXISTS=1
   fi
+fi
+
+# Main Script
+
+if [ $# -gt 0 ]; then
+    if [ "$1" == "--help" ]; then
+	usage $(basename $0)
+	exit 0
+    elif [ "$1" == "--reset" ]; then
+	reset_storage
+	exit 0
+    else
+        usage $(basename $0) >&2
+	exit 1
+    fi
 fi
 
 # If there is no volume group specified or no root volume group, there is
--- a/tests/005-test-devmapper-cleanup.sh
+++ b/tests/005-test-devmapper-cleanup.sh
@@ -1,0 +1,55 @@
+source $SRCDIR/libtest.sh
+
+# Test "docker-storage-setup reset". Returns 0 on success and 1 on failure.
+test_reset_devmapper() {
+  local devs=${TEST_DEVS}
+  local test_status
+  local testname=`basename "$0"`
+  local vg_name="dss-test-foo"
+
+ # Error out if any pre-existing volume group vg named dss-test-foo
+  for vg in $(vgs --noheadings -o vg_name); do
+    if [ "$vg" == "$vg_name" ]; then
+      echo "ERROR: $testname: Volume group $vg_name already exists."
+      return 1
+    fi
+  done
+
+  # Create config file
+  clean_config_files
+  cat << EOF > /etc/sysconfig/docker-storage-setup
+DEVS="$devs"
+VG=$vg_name
+EOF
+
+ # Run docker-storage-setup
+ $DSSBIN >> $LOGS 2>&1
+
+ # Test failed.
+ if [ $? -ne 0 ]; then
+    echo "ERROR: $testname: $DSSBIN Failed." >> $LOGS
+    cleanup $vg_name "$devs"
+    return 1
+ fi
+
+  test_status=1
+  $DSSBIN --reset >> $LOGS 2>&1
+  # Test failed.
+  if [ $? -eq 0 ]; then
+     if [ ! -e /etc/sysconfig/docker-storage ]; then
+          test_status=0
+     fi
+  fi
+  if [ ${test_status} -eq 1 ]; then
+     echo "ERROR: $testname: $DSSBIN --reset Failed." >> $LOGS
+  fi
+
+  cleanup $vg_name "$devs"
+
+  return $test_status
+}
+
+# Create a devicemapper docker backend and then make sure the
+# `docker-storage-setup --reset`
+# cleans it up properly.
+test_reset_devmapper
--- a/tests/006-test-overlay-cleanup.sh
+++ b/tests/006-test-overlay-cleanup.sh
@@ -1,0 +1,44 @@
+source $SRCDIR/libtest.sh
+
+# Test "docker-storage-setup reset". Returns 0 on success and 1 on failure.
+test_reset_overlay() {
+  local test_status
+  local testname=`basename "$0"`
+
+  # Create config file
+  clean_config_files
+  cat << EOF > /etc/sysconfig/docker-storage-setup
+STORAGE_DRIVER=overlay
+EOF
+
+ # Run docker-storage-setup
+ $DSSBIN >> $LOGS 2>&1
+
+ # Test failed.
+ if [ $? -ne 0 ]; then
+    echo "ERROR: $testname: $DSSBIN Failed." >> $LOGS
+    clean_config_files
+    return 1
+ fi
+
+ test_status=0
+ $DSSBIN --reset >> $LOGS 2>&1
+ if [ $? -ne 0 ]; then
+    # Test failed.
+    test_status=1
+ elif [ -e /etc/sysconfig/docker-storage ]; then
+    # Test failed.
+    test_status=1
+ fi
+  if [ ${test_status} -eq 1 ]; then
+    echo "ERROR: $testname: $DSSBIN --reset Failed." >> $LOGS
+  fi
+
+ clean_config_files
+ return $test_status
+}
+
+# Create a overlay docker backend and then make sure the
+# docker-storage-setup --reset
+# cleans it up properly.
+test_reset_overlay


rhatdan:
@cgwalters @rhvgoyal PTAL

docker-storage-setup.sh
```diff
@@ -823,6 +830,13 @@ setup_storage() {
   fi
 }
 
+reset_storage() {
+    if [ "$STORAGE_DRIVER" == "devicemapper" ]; then
+	reset_lvm_thin_pool
+    fi
+    rm -f $DOCKER_STORAGE
```
cgwalters: Seems like it'd be a bit friendlier to reset this to the default.  On an Atomic Host system, we could actually just do:

```
cp /{usr/,}etc/sysconfig/docker-storage
```

On a yum-managed system this isn't available though, so we could just test for it, like:

```
if test -f /usr/$DOCKER_STORAGE; then
  cp /usr/$DOCKER_STORAGE $DOCKER_STORAGE
else
  rm -f $DOCKER_STORAGE
fi
```
?

docker-storage-setup.sh
```diff
@@ -888,6 +895,17 @@ else
   fi
 fi
 
+# Main Script
+
+if [ $# -gt 0 ]; then
+    if [ "$1" == "reset" ]; then
```
cgwalters: Need to add this to usage and the man page.

rhvgoyal:
Please add a test case also.

docker-storage-setup.sh
rhvgoyal: @cgwalters We don't seem to install /etc/sysconfig/docker-storage by default. So there is nothing to restore over reset.

cgwalters: Yes we do.  `rpm -ql docker |grep docker-storage`.  Also on an Atomic Host system, I clearly see `/usr/etc/sysconfig/docker-storage`.

rhvgoyal: Hmm.., strange. I can't seem to find in spec file where we install it. And what's the purpose of default file. Default file is just supposed to be empty and user is not supposed to touch it anyaway. I see following which installs default /etc/sysconfig/docker-storage-setup.

Something is not adding up.

install -d %{buildroot}%{_sysconfdir}/sysconfig
install -p -m 644 %{repo}-storage-setup-override.conf %{buildroot}%{_sysconfdir}/sysconfig/%{repo}-storage-setup

rhvgoyal: May be it is coming from docker and not docker-storage-setup

rhvgoyal: Found it. It is coming from docker.

rhatdan:
Added test case.

docker-storage-setup.sh
rhatdan: We might want to keep this secret since it does not do it all, or we would have to at least document that you still need to rm -rf /var/lib/docker to get this to work.

We are just doing this in docker-storage-setup for a helper to atomic storage reset.

cgwalters: Call it `internal-only-for-atomic-reset` then?

rhatdan: I'm fine with this.

rhvgoyal: This name looks pretty odd to me. I prefer that we continue to call it "reset". If need be, just output a message saying that one will need to cleanup docker metadata also.

rhvgoyal: In fact, initially don't do anything. We will just document it. Also we can document in man page that preferred way to do this is by calling "atomic storage reset".

Also if docker metadata is not cleaned and next time one starts docker, docker-storage-setup will fail saying there is already devicemapper metadata in dokcer and it will refuse to setup new thin pool.

So I really don't see a need of an option like `internal-only-for-atomic-reset`

docker-storage-setup.sh
rhvgoyal: Having said that, I feel we probably should get rid of that file from docker package. It seems to be just a place holder for documentation. All the documentation should go in web pages and in /usr/share/doc/ dir. We are not expecting users to modify /etc/sysconfig/docker-storage and its generated by tool.

So I would think, that don't restore it and in a separate PR to docker, simply remove it and put relevant documentation as appropriate.

tests/004-test-cleanup
```diff
@@ -0,0 +1,82 @@
+source $SRCDIR/libtest.sh
+
+cleanup() {
+  local vg_name=$1
+  local devs=$2
+  $DSSBIN internal-only-for-atomic-reset >> $LOGS 2>&1
+  test -e /etc/sysconfig/docker-storage && echo "ERROR: $0: Failed." && exit 1
+  clean_config_files
```
rhvgoyal: Tests are not supposed to exit on their own. They are supposed to do all the cleanup and return success/failure status. Parent does all the logging w.r.t whether test passed or failed.

So once we have called "dss reset", we just need to make sure lvm thin pool went away and /etc/sysconfig/docker-storage went away and determine test status based on that.

And then call cleanup() to cleanup everything (even if test failed). This will allow us to continue running other tests even if some tests failed and these tests will require a clean environment to run.

tests/004-test-cleanup
```diff
@@ -0,0 +1,82 @@
...
+  for vg in $(vgs --noheadings -o vg_name); do
+    if [ "$vg" == "$vg_name" ]; then
+      test_status=0
+      break
+    fi
+  done
+
+  cleanup $vg_name $devs
+  return $test_status
+}
```
rhvgoyal: This test_status should be set based on whether lvm thin pool, volume group and /etc/sysconfig/docker-storage got removed or not and returned to caller.

tests/004-test-cleanup
```diff
@@ -0,0 +1,82 @@
...
+    cleanup
+    return 1
+ fi
+
+  cleanup
+  return 0
+}
+
+test_devs
+test_overlay
```
rhvgoyal: How about keeping testing overlay and lvm thin pool configuration in separate tests.

tests/libtest.sh
```diff
@@ -29,3 +29,14 @@ wipe_signatures() {
...
+cleanup() {
+  local vg_name=$1
+  local devs=$2
+
+  vgremove -y $vg_name >> $LOGS 2>&1
+  remove_pvs "$devs"
+  remove_partitions "$devs"
+  clean_config_files
+  wipe_signatures "$devs"
+}
```
rhvgoyal: Thanks for moving this common code in libtest.sh. Down the line I want to do more cleanup and move more common code in libtest.sh. That's for separate PR though.

docker-storage-setup.sh
cgwalters: This is one of those things where I feel the push from some groups to have an empty `/etc` by default sounds good in theory, but in practice would be painful.  There's a massive amount of precedence for having stubbed out configuration files in `/etc`.  It greatly aids "discoverability".

I think deleting it now is not such a good idea.  But I'm not strictly opposed to it either.

rhvgoyal: I understand what you are saying. We will continue to have /etc/sysconfig/docker-storage-setup which already has couple of lines and we can add more.

Key question is, by retaining /etc/sysconfig/docker-storage, we are promoting users to pass in dm.metadatadev and dm.datadev and that's a mode we dss does not support and we don't want to support.

So even if we want to retain the file, that's fine but atleast we should cleanup the text and get rid of mentions of doing dm.datatdev and dm.metadatadev.

jlebon:
Ahh, I see the issue with the tester now. Seems like somebody wrote one bash bracket too many. :innocent:

jlebon:
bot, retest this please

rhatdan:
Woohoo finally got the tests to work.

@rhvgoyal PTAL

docker-storage-setup.1
```diff
@@ -9,6 +9,9 @@ docker\-storage\-setup - Grows the root filesystem and sets up storage for docke
 .PP
 \f[B]-h, --help\f[]
   Print usage statement
+\f[B]--reset\f[]
+  Reset your docker storage to init state.  This command is not sufficient to cleanup your
+  docker environment.  It is used by other tools like \f[B]atomic storage setup\f[]
```
rhvgoyal: Can we mention that reset does not try to remove volume groups or try to remove any of the disks added previously.

tests/005-test-devmapper-cleanup
```diff
@@ -0,0 +1,61 @@
...
+    if [ "$vg" == "$vg_name" ]; then
+      test_status=0
+      break
+    fi
+  done
+
+  return $(cleanup_local $vg_name $devs)
+}
+
+test_devs
```
rhvgoyal: Can we put couple of lines in statement mentioning what this test is testing.

tests/005-test-devmapper-cleanup
```diff
@@ -0,0 +1,61 @@
...
+  test_status=0
+  if [ -e /etc/sysconfig/docker-storage ]; then
+     echo "  $DSSBIN reset Failed: $(basename testname)"
+     test_status=1
+  fi
+  cleanup $vg_name $devs
+  echo $test_status
+}
+
+# Test DEVS= directive. Returns 0 on success and 1 on failure.
```
rhvgoyal: This comment is not appropriate for this test.

tests/005-test-devmapper-cleanup
```diff
@@ -0,0 +1,61 @@
...
+test_devs() {
+  local devs=$DEVS
```
rhvgoyal: maybe call it test_reset_devmaper()

tests/005-test-devmapper-cleanup
```diff
@@ -0,0 +1,61 @@
...
+  local test_status
+  local testname=`basename "$0"`
+  local vg_name
```
rhvgoyal: This vg_name is redundant. I need to cleanup from other tests too. Will take care of that later.

tests/005-test-devmapper-cleanup
```diff
@@ -0,0 +1,61 @@
...
+  local vg_name="dss-test-dev"
+
```
rhvgoyal: In other tests we are using vg name dss-test-foo. Can we use the same please. At some point of time I want to move them to some common code.

tests/005-test-devmapper-cleanup
```diff
@@ -0,0 +1,61 @@
...
+VG=$vg_name
+EOF
+
+ # Run docker-storage-setup
+ $DSSBIN >> $LOGS 2>&1
+
+ # Test failed.
+ if [ $? -ne 0 ]; then
+    cleanup_local $vg_name $devs
+    return 1
```
rhvgoyal: Here we can simply call cleanup() instead of cleanup_local() right?

tests/005-test-devmapper-cleanup
```diff
@@ -0,0 +1,61 @@
...
+ fi
+
+ test_status=1
+ # Make sure volume group $VG got created.
+  for vg in $(vgs --noheadings -o vg_name); do
+    if [ "$vg" == "$vg_name" ]; then
+      test_status=0
+      break
```
rhvgoyal: Here test_status should not be set to 0. It should be set to zero after cleanup is done successfully.

tests/005-test-devmapper-cleanup
```diff
@@ -0,0 +1,61 @@
+source $SRCDIR/libtest.sh
+
+cleanup_local() {
+  local test_status
+  local vg_name=$1
+  local devs=$2
+  $DSSBIN --reset >> $LOGS 2>&1
+  test_status=0
```
rhvgoyal: Can we capture return code of DSSBIN and make sure it is 0 other test failed anyway.

tests/005-test-devmapper-cleanup
```diff
@@ -0,0 +1,61 @@
...
+  if [ -e /etc/sysconfig/docker-storage ]; then
+     echo "  $DSSBIN reset Failed: $(basename testname)"
+     test_status=1
+  fi
+  cleanup $vg_name $devs
```
rhvgoyal: Can we move this check outside of cleanup_local(). In fact we can get rid of cleanup_local() completely and keep everything in parent function, That's much more readable. This function is doing too much. Calling --reset, settin test status as well as doing actual cleanup. 

I think we can easily do that from parent function and that's more readable.

tests/005-test-devmapper-cleanup
rhatdan: Done

tests/005-test-devmapper-cleanup
rhatdan: Done

tests/005-test-devmapper-cleanup
rhatdan: I got a failure on this test, which is why I changed it, I will change it back and see if tests succeed.

tests/005-test-devmapper-cleanup
rhatdan: Done

rhatdan:
@rhvgoyal I think I got them all.

docker-storage-setup.1
```diff
@@ -7,9 +7,16 @@ docker\-storage\-setup - Grows the root filesystem and sets up storage for docke
...
+\f[B]--help\f[]
   Print usage statement
 
+\f[B]--reset\f[]
+  Reset your docker storage to init state. Reset does not try to remove volume groups or try to remove any of the disks added previously.
+  
+Note: The
+\f[B]--reset\f[]
+command is not sufficient to cleanup your docker environment.  This option is used by other tools. (\f[B]atomic storage setup\f[])
+
```
rhvgoyal: you mean "atomic storage reset" ?

tests/003-test-override-signature-wipes-existing-signatures.sh
```diff
@@ -50,7 +50,7 @@ EOF
   return $test_status
 }
 
-# Create a disk with some signature, say lvm signature and make sure
-# override signature can override that, wipe signature and create thin
-# pool.
+# Create a overlay docker backend and then make sure the
+# `docker-storage-setup --reset`
+# cleans it up properly.
```
rhvgoyal: Why are you modifying an exisitng test case description.

tests/005-test-devmapper-cleanup
```diff
@@ -0,0 +1,61 @@
+source $SRCDIR/libtest.sh
+
+# Test DEVS= directive. Returns 0 on success and 1 on failure.
```
rhvgoyal: This comment needs to be changed to reflect that reset operation is being tested.

tests/005-test-devmapper-cleanup
```diff
@@ -0,0 +1,61 @@
...
+DEVS="$devs"
+VG=$vg_name
+EOF
+
+ # Run docker-storage-setup
+ $DSSBIN >> $LOGS 2>&1
+
+ # Test failed.
+ if [ $? -ne 0 ]; then
+    cleanup $vg_name $devs
```
rhvgoyal: It should be "$devs". Otherwise if there are multiple devices, only first device is cleaned and rest are ignored.

tests/005-test-devmapper-cleanup
```diff
@@ -0,0 +1,61 @@
...
+  test_status=0
+  $DSSBIN --reset >> $LOGS 2>&1
+  # Test failed.
+  if [ $? -ne 0 ]; then
+     test_status=1
+  elif [ -e /etc/sysconfig/docker-storage ]; then
+     echo "  $DSSBIN reset Failed: $(basename testname)"
+     test_status=1
+  fi
+  cleanup $vg_name $devs
```
rhvgoyal: same here. "$devs"

tests/006-test-overlay-cleanup
```diff
@@ -0,0 +1,38 @@
...
+STORAGE_DRIVER=overlay
+EOF
+
+ # Run docker-storage-setup
+ $DSSBIN >> $LOGS 2>&1
+
+ # Test failed.
+ if [ $? -ne 0 ]; then
+    echo "ERROR: $testname: Failed."
+    cleanup
```
rhvgoyal: we can probably just call clean_config_files().

rhatdan:
@rhvgoyal PTAL, need to get this committed so we can continue working on atomic storage.

tests/005-test-devmapper-cleanup
```diff
@@ -0,0 +1,61 @@
+source $SRCDIR/libtest.sh
+
+# Test "docker-storage-setup reset". Returns 0 on success and 1 on failure.
+test_reset_devmapper() {
+  local devs=$DEVS
+  local test_status
+  local testname=`basename "$0"`
+  local vg_name
```
rhvgoyal: This vg_name is redundant and can be removed.

tests/005-test-devmapper-cleanup
rhvgoyal: test_status should not be 0 here.

tests/005-test-devmapper-cleanup
```diff
@@ -0,0 +1,61 @@
...
+      break
+    fi
+  done
+
+  test_status=0
+  $DSSBIN --reset >> $LOGS 2>&1
+  # Test failed.
+  if [ $? -ne 0 ]; then
+     test_status=1
+  elif [ -e /etc/sysconfig/docker-storage ]; then
```
rhvgoyal: I think we nee to make test_status thing little cleaner. It would be good if test_status is initialized to 1 and we set it to zero once only after test is successful. In all the other cases simple do cleanup and return.

This test case now seems to be a mix. First we initialized test_status=1, then checked presence of volume group and set to 0. And then set to zero again and then setting it to 1 on failure etc.

So please untangle this.

rhvgoyal:
Just ran these tests on local machine with two disks. First of all after the tests have run, I see that both my disks still have partitoins. So somwehere cleanup has not succeeded.  Also I see following on stdout.

PASS: 004-test-non-absolute-disk-name-support.sh
  ../docker-storage-setup.sh reset Failed: testname
PASS: 005-test-devmapper-cleanup
PASS: 006-test-overlay-cleanup

tests/006-test-overlay-cleanup
```diff
@@ -0,0 +1,39 @@
...
+
+ test_status=0
+ $DSSBIN --reset >> $LOGS 2>&1
+ if [ $? -ne 0 ]; then
+    # Test failed.
+    test_status=1
+ elif [ -e /etc/sysconfig/docker-storage ]; then
+    # Test failed.
+    echo "  $DSSBIN reset Failed: $(basename testname)"
+    test_status=1
```
rhvgoyal: We will need to cleanup config file even in case of failure.

Also we already did "basename" on testname. So no need to do it again.

rhatdan: All code paths cleanup config.

rhvgoyal:
@rhatdan I have generated another PR and that should fix the case of overlay test adding disks to system volume group.

Can you please renaming your tests to end with ".sh" suffix. That way syntax highlighting works better and it is consistent with other tests.

rhatdan:
@rhvgoyal Changed name and switch to not use DEVS.

rhvgoyal:
LGTM

rhvgoyal:
@cgwalters-bot r+

PR #131, opened 2016-05-05
[merged] Convert device names to absolute names starting with /dev

rhvgoyal:
Convert device names to absolute names in the beginning and use those names throughout the script.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -63,6 +63,10 @@ STORAGE_DRIVERS="devicemapper overlay"
 
 DOCKER_METADATA_DIR="/var/lib/docker"
 
+# DEVS can have device names without absolute path. Convert these to absolute
+# paths and save in ABS_DEVS and use in rest of the code.
+DEVS_ABS=""
+
 # Will have currently configured storage options in $DOCKER_STORAGE
 CURRENT_STORAGE_OPTIONS=""
 
@@ -337,7 +341,7 @@ create_data_lv() {
 }
 
 create_lvm_thin_pool () {
-  if [ -z "$DEVS" ] && [ -z "$VG_EXISTS" ]; then
+  if [ -z "$DEVS_ABS" ] && [ -z "$VG_EXISTS" ]; then
     Fatal "Specified volume group $VG does not exist, and no devices were specified"
   fi
 
@@ -525,7 +529,6 @@ vg_exists() {
 is_block_dev_partition() {
   local bdev=$1
 
-  [ ! -b "$bdev" ] && bdev="/dev/${bdev}"
   if ! disktype=$(lsblk -n --nodeps --output type ${bdev}); then
     Fatal "Failed to run lsblk on device $bdev"
   fi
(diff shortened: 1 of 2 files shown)


rhvgoyal:
cc @cgwalters @rhatdan

docker-storage-setup.sh
```diff
@@ -63,6 +63,10 @@ STORAGE_DRIVERS="devicemapper overlay"
 
 DOCKER_METADATA_DIR="/var/lib/docker"
 
+# DEVS can have device names without absolute path. Convert these to absolute
+# paths and save in ABS_DEVS and use in rest of the code.
+DEVS_ABS=""
+
```
rhatdan: Why not leave them as DEVS and then fix them in your process_devs option.

docker-storage-setup.sh
```diff
@@ -891,7 +891,7 @@ fi
 # If there is no volume group specified or no root volume group, there is
 # nothing to do in terms of dealing with disks.
 if [[ -n "$DEVS" && -n "$VG" ]]; then
-  check_block_devs ${DEVS}
+  process_block_devs ${DEVS}
```
rhatdan: Something like 

DEVS=$(process_block_devs ${DEVS})

rhvgoyal: Wanted to retain $DEVS, just in case we need original names of devices somewhere.

Can do following instead.

DEVS_ABS=$(process_block_devs ${DEVS})

docker-storage-setup.sh
rhvgoyal: Being paranoid. Wanted to retain original names just in case we end up using them for output messages. Not very particular about it, but don't want to lose track of what user passed in originally.

rhvgoyal:
@rhatdan Took care of one of your comments. PTAL.

rhvgoyal:
Don't merge this yet. Doing some more testing locally.

rhatdan:
LGTM.

rhvgoyal:
@rhatdan I noticed that this patch introduced issues if multiple devices are passed in DEVS=. Had to pass arguments in double quotes. Fixed it and pushed again. PTAL.

docker-storage-setup.sh
```diff
@@ -564,29 +565,35 @@ check_wipe_block_dev_sig() {
 }
 
 # Make sure passed in devices are valid block devies. Also make sure they
-# are not partitions.
-check_block_devs() {
-  local devs=$1
+# are not partitions. Names which are of the form "sdb", convert them to
+# their absolute path for processing in rest of the script.
+process_block_devs() {
```
cgwalters: I'd call this `canonicalize_` rather than `process_` personally.

rhvgoyal: Sounds better. Will do.

rhvgoyal:
@cgwalters Renamed function. PTAL.

cgwalters:
One thing you can do with Homu BTW is use `git commit --fixup`.  That preserves a linear PR history.

cgwalters:
@cgwalters-bot r+ 79fd355

PR #133, opened 2016-05-06
Support for adding/removing physical volumes

mvollmer:
This sits on top of #128.  The idea is that this is wrapped by "atomic storage", just as "--reset".

This change is intended to support the Cockpit use cases describes here:

    https://github.com/cockpit-project/cockpit/wiki/Atomic:-Docker-Storage

This change allows adding disks to the storage pool, and removing them again.  Providing information about the storage pool will be supported by a later PR.

The `--add` option has been added to offer easy adding of disks, without having to edit /etc/sysconfig/docker-storage-setup.  The `--add` operation is a one-time, immediate change to the storage pool that either works or doesn't and does not have any effect beyond that.  For example, the block devices given to `--add` are not processed again the next time docker is started.

The ``--reset-and-reduce` option has been added to offer opportunistic removal of unused block devices from the pool, in a way that puts the devices back into a state that is acceptable to `--add`.

The user interface and API to these features is via the `atomic storage` command, see https://github.com/projectatomic/atomic/pull/373.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -379,6 +379,13 @@ check_docker_storage_metadata() {
   exit 1
 }
 
+reset_lvm_thin_pool () {
+  if lvm_pool_exists; then
+      lvchange -an $VG/${POOL_LV_NAME}
+      lvremove $VG/${POOL_LV_NAME}
+  fi
+}
+
 setup_lvm_thin_pool () {
   local tpool
 
@@ -823,6 +830,13 @@ setup_storage() {
   fi
 }
 
+reset_storage() {
+    if [ "$STORAGE_DRIVER" == "devicemapper" ]; then
+	reset_lvm_thin_pool
+    fi
+    rm -f $DOCKER_STORAGE
+}
+
 usage() {
   cat >&2 <<-FOE
     Usage: $1 [OPTIONS]
@@ -833,13 +847,6 @@ usage() {
       -h, --help    Print help message
 FOE
 }
-
-# Main Script
-
-if [ $# -gt 0 ]; then
-  usage $(basename $0)
-  exit 0
-fi
 
 # Source library. If there is a library present in same dir as d-s-s, source
 # that otherwise fall back to standard library. This is useful when modifyin
@@ -886,6 +893,17 @@ else
   if vg_exists "$VG";then
     VG_EXISTS=1
   fi
+fi
+
+# Main Script
+
+if [ $# -gt 0 ]; then
+    if [ "$1" == "--reset" ]; then
+	reset_storage
+	exit 0
+    fi
+    usage $(basename $0)
+    exit 0
 fi
 
 # If there is no volume group specified or no root volume group, there is
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -837,6 +837,28 @@ reset_storage() {
     rm -f $DOCKER_STORAGE
 }
 
+maybe_wipe_disk() {
+    # Wipe $1 if it has a partition table with exactly one partition
+    base=$(basename $1)
+    partcount=$(awk "\$4 ~ /^${base}[0-9]+/ {count += 1} END { print count; }" /proc/partitions)
+    if [ "$partcount" == "1" ]; then
+        wipefs -a $1
+    fi
+}
+
+reduce_storage() {
+    if [ -n "$VG" ]; then
+        pvs=$(pvs --noheadings -o pv_name,vg_name | awk "\$2 ~ /^$VG\$/ { print \$1 }" )
+        for pv in $pvs; do
+            if [ $(pvs --noheadings --unit B -o pv_used $pv) == "0B" ]; then
+                vgreduce $VG $pv
+                wipefs -a $pv
+                maybe_wipe_disk $(echo $pv | sed -r 's/([^0-9]*)([0-9]+)/\1/')
+            fi
+        done
+    fi
+}
+
 usage() {
   cat <<-FOE
     Usage: $1 [OPTIONS]
@@ -845,6 +867,12 @@ usage() {
 
     Options:
      --help    Print help message
+
+     --reset   Remove all images and reset storage
+
+     --reset-and-reduce
+               Remove all images, reset storage, and reduce
+               the storage pool to its minimum size
 FOE
 }
 
@@ -903,6 +931,10 @@ if [ $# -gt 0 ]; then
         exit 0
     elif [ "$1" == "--reset" ]; then
 	reset_storage
+	exit 0
+    elif [ "$1" == "--reset-and-reduce" ]; then
+	reset_storage
+        reduce_storage
 	exit 0
     else
         usage $(basename $0) >&2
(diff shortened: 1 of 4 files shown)


docker-storage-setup.sh
```diff
@@ -837,6 +837,28 @@ reset_storage() {
...
+}
+
+reduce_storage() {
+    if [ -n "$VG" ]; then
+        pvs=$(pvs --noheadings -o pv_name,vg_name | awk "\$2 ~ /^$VG\$/ { print \$1 }" )
+        for pv in $pvs; do
+            if [ $(pvs --noheadings --unit B -o pv_used $pv) == "0B" ]; then
+                vgreduce $VG $pv
+                wipefs -a $pv
+                maybe_wipe_disk $(echo /dev/sda1 | sed -r 's/([^0-9]*)([0-9]+)/\1/')
```
mvollmer: Ouch.

rhvgoyal:
@mvollmer It will help a lot if you write a good description of what you are doing and why you are doing. One line descriptions really don't help.

rhvgoyal:
--add should really not be a command line option. One can easily add a disk in /etc/sysconfig/docker-storage-setup (DEVS="/dev/sda /dev/sdb") and restart docker-storage-setup service.

rhvgoyal:
I think removing of disks and volume group can easily be done outside of docker-storage-setup. Why not do it that way.

docker thin pool and root lv share same volume group by default. And these disk might easily be used by root lv and attempt to reduce volume group and disk will fail anyway.

So instead of making docker-storage-setup complicated, I think it might be better that removing the disk from volume group is done outside of docker-storage-setup using lvm commands.

rhatdan:
@mvollmer I think we should open issues to discuss.

A user wants to add additional storage to the docker storage, How would he do it? How could I do this from program control?

A user wants to remove storage from docker storage, How would he do it?  How could I do this from program control?  

Then we need to document these use cases and figure out how we could do this from atomic or cockpit, if we choose to implement them.

rhatdan:
Most of these use cases should be just modify /etc/sysconfig/docker-storage-setup and restart docker-storage-setup command.

mvollmer:
> --add should really not be a command line option. One can easily add a disk in /etc/sysconfig/docker-storage-setup (DEVS="/dev/sda /dev/sdb") and restart docker-storage-setup service.

/e/s/d-s-s is a shell script, and having to edit it is not a nice API.

mvollmer:
> I think removing of disks and volume group can easily be done outside of docker-storage-setup. Why not do it that way.

Because d-s-s is the entity that adds disks, and it thus makes sense to me that it should also be the entity that removes them.  Wiping the disks just right is not entirely straightforward, and d-s-s know what it does with partitions and it can thus take the responsibility to undo that upon removal.

> docker thin pool and root lv share same volume group by default. And these disk might easily be used by root lv and attempt to reduce volume group and disk will fail anyway.

Yeah, sh*t happens. :-)  But not always, and if we deem it important, we can use a dedicated volume group by default, or advise people to do that.  The alternative would be to require people to reinstall the host in order to get the disks out of the docker storage pool, which might be totally acceptable.

> So instead of making docker-storage-setup complicated, I think it might be better that removing the disk from volume group is done outside of docker-storage-setup using lvm commands.

d-s-s might have less code, but I think the bigger picture would be more complex.

mvollmer:
> Most of these use cases should be just modify /etc/sysconfig/docker-storage-setup 

Editing /e/s/d-s-s programmatically is painful and unreliable.  We can do better.

Removing disks is not possible by changing /e/s/d-s-s.

The use cases should be docker-agnostic and be done via the `atomic storage` command, not d-s-s.

mvollmer:
> I think removing of disks and volume group can easily be done outside of docker-storage-setup.

Would it be appropriate to do it inside `atomic storage`?

mvollmer:
> It will help a lot if you write a good description of what you are doing and why you are doing. One line descriptions really don't help.

I have included a longer description at the top.  Does this help?

mvollmer:
> The user interface and API to these features is via the atomic storage command, see projectatomic/atomic#373.

In other words, I don't really care how the work is divided between `atomic` and `docker-storage-setup` and whether `atomic` does its job by editing /e/s/d-s-s or not.

mvollmer:
Are the tests expected to pass?  This looks like it probably isn't my fault:

```
tests/004-test-devmapper-cleanup: line 62: test_overlay: command not found
FAIL: 004-test-devmapper-cleanup
```

rhatdan:
I think adding this to atomic storage is a better idea,  I believe we should keep d-s-s to reading the config and creating the storage, Reset is fine in d-s-s since it is doing some stuff inside that external tools would not know about.

```
atomic storage --add-device=/dev/foobar
atomic storage --list
atomic storage --remove-device=/dev/foobar
```

atomic also has the nice benefit of being written in python versus Bash, so doing more complex things like managing config files is a little easier to do.

mvollmer:
> I think adding this to atomic storage is a better idea,

How would you propose to implement `atomic storage add`, in contrast to https://github.com/projectatomic/atomic/pull/373?

rhvgoyal:
I think adding additional commands in atomic sounds like a decent idea.

What I find little odd about it though that part of the configuration happens directly by working on config file (/etc/sysconfig/docker-storage-setup) and then rest of the configuration happens using atomic.

I think lets take it case by case basis. For the "status" of current state, I feel it should be implemented in docker-storage-setup and atomic status can just call it. Adding and removing disks can completely be an atomic implementation without any support from docker-storage-setup.

rhatdan:
Yes anthing to do with modifying /etc/sysconfig/docker-storage-setup should be done with external tools like vi, atomic and potentially cockpit.  Then just restart docker-storage-setup to read the modified config and fix the configuration.

rhvgoyal:
Sure, now only thing is that instead of cockpit, you are willing to provide all that functionality in atomic. I think that's fine. Life will become easier for normal command line users as well. Dealing with volume groups and lvs and pvs is not always easy.

mvollmer:
> Yes anthing to do with modifying /etc/sysconfig/docker-storage-setup should be done with external tools like vi, atomic and potentially cockpit. Then just restart docker-storage-setup to read the modified config and fix the configuration.

So `atomic storage add /dev/foo` would change /etc/sysconfig/docker-storage-config and make sure that DEVS includes /dev/foo, right?  Would it also run d-s-s?  Would it remove /dev/foo from DEVS afterwards?  Always, or only when d-s-s fails?

rhatdan:
It would leave it there permanently, and yes, I would figure it would rerun the service. If it fails it would restore the /etc/sysconfig/docker-storage-setup.

Something like:

cp /etc/sysconfig/docker-storage-setup  /etc/sysconfig/docker-storage-setup.bkp
modify   /etc/sysconfig/docker-storage-setup
run d-s-s
if Success
      rm  /etc/sysconfig/docker-storage-setup.bkp
else
    mv  /etc/sysconfig/docker-storage-setup.bkp   /etc/sysconfig/docker-storage-setup
    d-s-s
fi

mvollmer:
> Something like: [...]

Alright, I'll do that. Thanks!

mvollmer:
This is obsolete, see https://github.com/projectatomic/atomic/pull/373.

PR #134, opened 2016-05-09
[merged] test: Export TEST_DEVS instead of DEVS

rhvgoyal:
DEVS is parsed by docker-storage-setup. So if some test does not want to
use it, still it is used by the test and that can lead to unexpected
results. For example, when somebody tries to test overlay, DEVS is
still set and all the disks are unexpectedly added to volume group
backing root lv. That's not what overlay test wanted to test.


rhvgoyal:
cc @rhatdan @cgwalters

rhatdan:
LGTM

rhatdan:
@cgwalters-bot r+ 49c6304

PR #135, opened 2016-05-10
Add --show-vgroup command

mvollmer:
This is helpful for programs that want to work on the volume group themselves.

This sits on top of #128 only to avoid a merge conflict.

### DIFF
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -7,8 +7,15 @@ docker\-storage\-setup - Grows the root filesystem and sets up storage for docke
 \f[B]docker-storage-setup\f[] [OPTIONS]
 .SH OPTIONS
 .PP
-\f[B]-h, --help\f[]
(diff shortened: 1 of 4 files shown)


rhvgoyal:
I think show volume group should not be a separate option. Instead we should implement a `status` command which displays status and volume group details can be part of it.

mvollmer:
> I think show volume group should not be a separate option. Instead we should implement a status command which displays status and volume group details can be part of it.

What else do you want to include?

rhvgoyal:
Storage Driver and then storage driver specific details. For example list of DEVS, Volume group, LVM thin pool name.

I think this might be a good starting point. And more things can be added if need be.

rhatdan:
Yes I agree, lets just dump an name value pair that would be easy for humans and atomic CLI to consume

mvollmer:
Continued in #136.

PR #136, opened 2016-05-11
Add --dump-config command

mvollmer:


mvollmer:
What about this?

rhatdan:
Why not just use --status which @rhvgoyal requested?

rhvgoyal:
I am wondering what will happen when docker-storage-setup service and command are running in parallel. I feel we need some kind of file based locking now to take care of race conditions.

docker-storage-setup.1
```diff
@@ -12,11 +12,17 @@ docker\-storage\-setup - Grows the root filesystem and sets up storage for docke
...
   Reset your docker storage to init state. Reset does not try to remove volume groups or try to remove any of the disks added previously.
-  
+
 Note: The
 \f[B]--reset\f[]
 command is not sufficient to cleanup your docker environment.  This option is used by other tools. (\f[B]atomic storage reset\f[])
 
+\f[B]--dump-config\f[]
+  Write part of the configuration to stdout in a machine readable
+  format.  Note that this does not describe the current Docker
```
rhvgoyal: Let these just be key=value pairs and these should be readable by humans as well as other programs. No need to call it machine readable.

docker-storage-setup.1
```diff
@@ -12,11 +12,17 @@ docker\-storage\-setup - Grows the root filesystem and sets up storage for docke
...
+  storage setup, but only what docker-storage-setup will try to setup
+  on its next run.
+
```
rhvgoyal: I feel current state of storage is much more important. How does it matter what dss will try to setup next?

docker-storage-setup.sh
```diff
@@ -848,6 +848,17 @@ reset_storage() {
...
+
+dump_config() {
+    echo STORAGE_DRIVER=$STORAGE_DRIVER
+    echo VG=$VG
+    if [ "$STORAGE_DRIVER" == "devicemapper" ]; then
+        echo LV=$POOL_LV_NAME
+    fi
+    echo DEVS=$DEVS
+}
+
```
rhvgoyal: I feel current configuration makes lot of sense and for future configuration, caller knows what modification he/she has done to docker-storage-setup file.

As a user I think what makes sense to me, is what's current storage driver, what's current lvm thin pool and which volume group it was carved out from and what disks were added.

What's future can easily be parsed from /etc/sysconfig/docker-storage-setup anyway.

docker-storage-setup.1
mvollmer: I agree.  I just did it this way because it is simple and enough for the needs of `atomic storage reset`.

docker-storage-setup.sh
mvollmer: > What's future can easily be parsed from /etc/sysconfig/docker-storage-setup anyway.

"easily" is relative. /e/s/d-s-s is a shell script, and the only way to parse it correctly is to execute it.  You can think of `--dump-config` as a utility that parses /e/s/d-s-s for you.

Also, default values are not "easily" computed, such as digging out the volume group behind "/" as the default value for $VG.  `--dump-config` will also compute these default values correctly for you.

Thus, even if a real `--status` is more important than a mere `--dump-config`, `--dump-config` is still useful since it's the best way to parse /e/s/d-s-s correctly if that's what you want.  (We could make it output everything and carefully document the format, though.  Maybe even make it optionally output JSON because inventing our own key/value syntax is a sin. :-)

docker-storage-setup.1
mvollmer: But we need to guarantee it anyway.

docker-storage-setup.1
mvollmer: In fact, it might be necessary to read the config.  Otherwise this sequence would fail:
```
atomic storage reset
atomic storage reduce
```
There is no current status after `atomic storage reset` and `atomic storage reduce` needs to read the configuration to figure out from which volume group to remove the disks,

mvollmer:
> Why not just use --status which @rhvgoyal requested? 

Using `--status` as the command name for the code here would be misleading.

The real `--status` is vastly different beast and not what is needed for `atomic storage reduce`.

(I realize I should have motivated this PR much better since the history is a bit convoluted now.  I will make a PR for `atomic storage reduce` that uses this PR, to show how it all could hang together.)

mvollmer:
See https://github.com/projectatomic/atomic/pull/382 for `atomic storage reduce`

docker-storage-setup.1
rhvgoyal: atomic can parse /etc/sysconfig/docker-sotrage-setup to figure out which disks to remove. What's the difference. --dump-config will have key value pairs and /etc/sysconfig/docker-storage-setup has key value pairs.

docker-storage-setup.sh
rhvgoyal: /etc/sysconfig/docker-storage-setup is a config file and not script.

rhvgoyal:
"atomic storage reduce" seems to need VG and DEVS to be able to remove disks from volume group. Both can be easily obtained from /etc/sysconfig/docker-storage-setup. Please parse that. I don't think we need an command line option for that.

rhvgoyal:
Rather, "atomic storage reduce" will need to know the current status of storage (not future configuration). So this goes back to the need of showing already configured state of docker-storage-setup and not proposed state.

rhvgoyal:
--dump-config seems to imply the input configuration to dss. If user did not pass in "VG", then it does not imply displaying VG. It does not implying displaying thin pool volume name. Or it does not imply displaying any other useful default value (chunk size of thin pool etc).

So --status seems to make sense to me which displays currently configured state of storage. This will be a mix of input configuration parameters and defaults chosen by dss.

rhvgoyal:
May be we can look at third option say "--info". This seems to be even more generic.

mvollmer:
> Both can be **easily** obtained from /etc/sysconfig/docker-storage-setup.

"You keep using that word. I do not think it means what you think it means."﻿ :-)

> Please parse that. I don't think we need an command line option for that.

Alright.

mvollmer:
> May be we can look at third option say "--info". This seems to be even more generic.

Makes sense to me.

docker-storage-setup.sh
mvollmer: I would totally put things like
```
VG="dockah-$(hostname)"
```
in it, if that gives the result I want.

PR #138, opened 2016-05-16
Handle partition names like "loop0p1"

mvollmer:


rhvgoyal:
I am very skeptical of allowing loop devices for docker-storage-setup. Reason being that devicemapper thin pool target developers don't like the idea of using loop devices for thin pool and have resued to support this configuration. 

And this is one reason for docker storage setup's existence. Setup a real thin pool which does not use loop devices to get the configuration right for user out of the box.

So please get a real block device (disk) to test it. In virtual environment it is really easy to export a virtual disk.

docker-storage-setup.sh
```diff
@@ -657,26 +657,36 @@ EOF
...
   fi
 }
 
+partition_name() {
+    if echo "$1" | grep -q '[0-9]$'; then
+        echo "${1}p1"
+    else
+        echo "${1}1"
+    fi
+}
```
rhvgoyal: How about if we scan the name of partition from disk using "lsblk" command. I think "lsblk -l" lists devices and all its partitions and we can scan that to figure out name of the partition.

docker-storage-setup.sh
```diff
@@ -657,26 +657,36 @@ EOF
...
+
 create_disk_partitions() {
-  local devs="$1"
+  local devs="$1" part
 
   for dev in $devs; do
-    create_partition $dev
+    part=$(partition_name $dev)
+
+    create_partition $dev $part
```
rhvgoyal: Either create_partition() can return name of partition or we could scan it after creating partition using lsblk.

docker-storage-setup.sh
```diff
@@ -657,26 +657,36 @@ EOF
     Fatal "udevadm settle after partition creation failed. Exiting."
   fi
 
-  if ! wait_for_dev ${dev}1; then
-    Fatal "Partition device ${dev}1 is not available"
+  if ! wait_for_dev ${part}; then
+    Fatal "Partition device ${part} is not available"
```
rhvgoyal: This is tricky. I suspect that after addition of udevadm settle, this code is unnecessary. Now partition should exist. So we could think of getting rid of this code entirely.

docker-storage-setup.sh
```diff
@@ -637,18 +638,17 @@ scan_disks() {
...
   # TODO:
   #   * Consider gpt, or unpartitioned volumes
   #   * Error handling when partition(s) already exist
-  #   * Deal with loop/nbd device names. See growpart code
   size=$(( $( awk "\$4 ~ /"$( basename $dev )"/ { print \$3 }" /proc/partitions ) * 2 - 2048 ))
     cat <<EOF | sfdisk $dev
 unit: sectors
 
-${dev}1 : start=     2048, size=  ${size}, Id=8e
+${part} : start=     2048, size=  ${size}, Id=8e
```
rhvgoyal: Does this option work for dos table? IOW, who decides what's will be the name of the partition?

mvollmer:
> So please get a real block device (disk) to test it. In virtual environment it is really easy to export a virtual disk.

That's what I'll do for the Cockpit integration tests. I am afraid I need help to do the same for the `atomic` integration tests.

docker-storage-setup.sh
mvollmer: That's a good idea.

docker-storage-setup.sh
mvollmer: create_partition already outputs to stdout, so we would have to make it stop doing that first...

docker-storage-setup.sh
mvollmer: Creating partitions _is_ tricky, unfortunately.  See https://bugs.freedesktop.org/show_bug.cgi?id=85477, for example.

The explicit wait is probably not necessary, but even when using `lsblk` we can still wait until lsblk outputs the expected child.

docker-storage-setup.sh
mvollmer: I have no idea.  Block device names are decided somewhere between kernel and udev.

docker-storage-setup.sh
mvollmer: Done.

rhvgoyal:
For docker-storage-setup CI tests, @jlebon setup an environment where there is spare disk for testing. I am hoping same can be done for atomic integration tests.

jlebon:
There is no secondary disk on the atomic tester at the moment, but that could be fixed.

mvollmer:
> is this change needed?

The atomic test_storage.sh does indeed rely on this, but only in an optional part of the test.  That part normally doesn't run for pull requests here since the test machine doesn't have the rootfs in a LVM2 volume group.

giuseppe:
ok I see, nevermind my comment.  The patch works for me.  IMHO, even if supporting /dev/loop* devices is not the goal of docker-storage-setup, it can be handy for testing

mvollmer:
> IMHO, even if supporting /dev/loop* devices is not the goal of docker-storage-setup, it can be handy for testing

It's a complete accident that docker-storage-setup does not support loopback devices, and the failure mode when you give it one is really bad.  (It hangs for a minute and then fails with a unhelpful error message.)

There are probably other block devices that use the `/dev/foo5p4` naming scheme, and they might be worth supporting.

(Incidentally, I don't fully understand why docker-storage-setup goes to the trouble of making a partition on the block device.  This used to be recommended, but I think the tide is turning on that.)

pdericson:
@mvollmer fyi I needed to make the following change for this to work

```diff
--- docker-storage-setup.sh	2016-09-05 16:41:45.000000000 +1000
+++ docker-storage-setup.sh.local	2016-09-05 17:06:39.000000000 +1000
@@ -645,7 +645,7 @@
     cat <<EOF | sfdisk $dev
 unit: sectors
 
-start=     2048, size=  ${size}, Id=8e
+${dev}p1 : start=     2048, size=  ${size}, Id=8e
 EOF
 
   # Sometimes on slow storage it takes a while for partition device to
```

mvollmer:
> -start=     2048, size=  ${size}, Id=8e
> +${dev}p1 : start=     2048, size=  ${size}, Id=8e

Can you show the error that you get without this change?

The point of this change is that we can only know the name of a partition after we create it.  So putting the correct name into the sfdisk instructions is impossible.

The sfdisk documentation says:

```
        [device :] name[=value], ...

              The device field is optional.  sfdisk extracts the partition number from the device name.  It allows to
              specify the partitions in random order.  This functionality is mostly used by --dump.  Don't use it  if
              you are not sure.
```

This makes me believe that it is actually recommended to not use a `device:` prefix.

So what breaks for you without it?

jlebon:
@rhvgoyal Can we reconsider this? To not officially support it is reasonable, but shouldn't it be at the user's discretion whether they should go against recommendations? (The same way as docker on loopback just issues a warning, but doesn't forbid it). Not to mention, this patch seems to resolve a more generic issue than that, no?

rhatdan:
Why not add --allow-loop flag to allow this and print a warning?

astatham:
Another vote for this to be integrated - for example cannot use nvme devices on AWS i3 instances at the moment

    [root@ip-XXXXXXXX rules.d]# lsblk
    NAME                                                                                     MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
    xvda                                                                                     202:0    0    8G  0 disk
    └─xvda1                                                                                  202:1    0    8G  0 part /
    nvme0n1                                                                                  259:0    0  1.7T  0 disk
    └─nvme0n1p1                                                                              259:2    0  1.7T  0 part
    nvme1n1                                                                                  259:1    0  1.7T  0 disk

rhvgoyal:
We definitely need to improve our logic of detecting partitions so that it works on nvme devices as well.

shishir-a412ed:
I discussed this with @rhvgoyal and he is not very keen on supporting loopback devices.
Although we still want to detect partitions in a more dynamic way (rather than adding a prefix in `sfdisk`).

@mvollmer Your patch addresses that. There were a few issues in your PR, so I just opened a new PR with fixes. Since most of the work was done by you, I have added your sign-off in the commit.

PR #234 should support nvme devices with partition names ending in p1 (You can use loopback devices to carve out a thin pool if you like, but we are not recommending that).

Shishir

mvollmer:
This is continued in #234

PR #142, opened 2016-06-15
Test 007: EXTRA_DOCKER_STORAGE_OPTIONS

stephen679:
Adding a test case for extra docker storage options, and seeing if user config inputs are reflected after docker-storage-setup runs.


rhvgoyal:
You introduced helper function vg_exists() but did not make use of it in test?

stephen679:
I did for test 007, I want to merge the helper before applying all the changes for tests 1-6

rhvgoyal:
Actually introduce helper first and then use it in test7, that would be better. 

In fact, it would be better that open a PR where you introduce helper vg_exists and move all the existing tests to make use of this helper. Once that is merged, open separate PR for test 7.

stephen679:
@rhvgoyal see PR #143

PR #143, opened 2016-06-15
[merged] modified old tests to use vg_exists helper function

stephen679:
Changed old test files to use vg_exists() helper function introduced in tests/libtest.sh


rhvgoyal:
LGTM

Thanks. This is good cleanup.

rhvgoyal:
@rh-atomic-bot r+

rhvgoyal:
@stephen679 I just noticed that in some of the tests you removed the call to clean_config_file(). Why?

PR #144, opened 2016-06-16
check if extra options get appended to docker config file

stephen679:
If a user specifies something in EXTRA_DOCKER_STORAGE_OPTIONS in hte /etc/sysconfig/docker-storage-setup file, these options should be reflected in the /etc/sysconfig/docker-storage config file after running docker-storage-setup in the option DOCKER_STORAGE_OPTIONS.


rhvgoyal:
Hey, first can we restore the calls to clean_config_files() which your previous PR removed. (Until and unless there is a valid reason to remove it).

PR #145, opened 2016-06-16
[merged] remove redundant calls to clean_config_files

stephen679:
Added back in clean_config_files that I accidentally deleted in previous merge.


rhvgoyal:
@stephen679 I am rethinking. Tests are passing even without clean_config_files(), so may be it is redundant. 

In fact we call cleanup() after test finishes and that calls clean_config_files(). So new test should get a clean slate. And run-tests.sh makes sure config files are not present before running first test.

So may be this clean_config_file() in the middle of test is redundant. 

Lets leave it for now as it is. If we run into issues we will introduce it again.

Test1 and Test6, still seem to be cleaning files, can you remove that call from those tests as well.

stephen679:
@rhvgoyal I cleaned up the files, PTAL.

rhvgoyal:
Can you merge both the patches into one and push again.

tests/001-test-use-devs-to-create-thin-pool.sh
```diff
@@ -3,7 +3,7 @@ source $SRCDIR/libtest.sh
 # Test DEVS= directive. Returns 0 on success and 1 on failure.
 test_devs() {
   local devs=$TEST_DEVS
-  local test_status
+  local test_status=1
```
rhvgoyal: I think making test_status related cleanup is a good idea but you seem to be making only for 1 test while others test seem to have the same thing.

Whatever change you do, please keep it same for all tests.

I would suggest that remove test_status related change from this PR and generate a separate PR for all test_status related cleanup for all tests.

tests/003-test-override-signature-wipes-existing-signatures.sh
```diff
@@ -12,6 +12,7 @@ test_override_signatures() {
     return 1
   fi 
 
+  # Create config file
```
rhvgoyal: There is no need to keep comments consistent across all tests.

tests/004-test-non-absolute-disk-name-support.sh
```diff
@@ -18,7 +18,8 @@ test_non_absolute_disk_name() {
     echo "ERROR: $testname: Volume group $vg_name already exists." >> $LOGS
     return 1
   fi
- 
+
+  # Create config files
```
rhvgoyal: No need to keep comments consistent across all tests.

tests/002-test-reject-disk-with-lvm-signature.sh
```diff
@@ -9,11 +9,12 @@ test_lvm_sig() {
   local vg_name="dss-test-foo"
 
   # Error out if any pre-existing volume group vg named dss-test-foo
-  if vg_exists "vg_name"; then
+  if vg_exists "$vg_name"; then
```
rhvgoyal: All related changes are generally kept in one patch.

So one good way to organize this patch series would be.

- First patch fixes this vg_name issue
- Second patch fixes cleanup config file issue across remaining tests.
- Third patch fixes test_status issue across all tests.

rhvgoyal:
LGTM. Thanks.

rhvgoyal:
@cgwalters-bot r+

rhvgoyal:
@rh-atomic-bot r+

rhvgoyal:
@cgwalters I see author name as "stephen679 <schung@redhat.com>" and not "Stephen Chung <schung@redhat.com>" in commits. Is it because of bot?

cgwalters:
The bot will use whatever is in the original git commit.

rhvgoyal:
@cgwalters ok, thanks. I think he needs to fix author name in his git settings.

PR #146, opened 2016-06-21
[merged] Test 007: Check EXTRA_DOCKER_STORAGE_OPTIONS

stephen679:
Added a test file that checks if EXTRA_DOCKER_STORAGE_OPTIONS actually shows up in /etc/sysconfig/docker-storage


tests/007-test-setting-extra-opts.sh
```diff
@@ -0,0 +1,57 @@
...
+
+  # dss failed
+  if [ $? -ne 0 ]; then 
+    cleanup $vg_name "$devs"
+    return $test_status
+  fi
+
+  # Check if docker-storage config file was created by dss
+  if [ ! -f /etc/sysconfig/docker-storage ]; then
+    echo "ERROR: $testname: docker-storage config file was not created during dss" >> $LOGS
```
rhvgoyal: cat we give full name of file including path in error message. Also get rid of "during dss".

"ERROR: $testname: /etc/sysconfig/docker-storage was not created."

tests/007-test-setting-extra-opts.sh
```diff
@@ -0,0 +1,57 @@
...
+    return $test_status
+  fi
+
+  source /etc/sysconfig/docker-storage
+  local options=$DOCKER_STORAGE_OPTIONS
+  
+  # Search for $extra_options in $options. 
+  echo $options | grep -- "$extra_options" >> $LOGS 2>&1
+  
+  # Successful appending utno DOCKER_STORAGE_OPTIONS
```
rhvgoyal: what is "utno" ?

tests/007-test-setting-extra-opts.sh
```diff
@@ -0,0 +1,57 @@
...
+
+  # Check if docker-storage config file was created by dss
+  if [ ! -f /etc/sysconfig/docker-storage ]; then
+    echo "ERROR: $testname: /etc/sysconfig/docker-storage file was not created." >> $LOGS
+    cleanup $vg_name "$devs"
+    return $test_status
+  fi
+
+  source /etc/sysconfig/docker-storage
+  local options=$DOCKER_STORAGE_OPTIONS
```
rhvgoyal: Do we need this local variable? just use $DOCKER_STORAGE_OPTIONS directly.

tests/007-test-setting-extra-opts.sh
stephen679: oops, meant to type "onto", but just going to change it to "to"

tests/007-test-setting-extra-opts.sh
```diff
@@ -0,0 +1,56 @@
...
+  if [ ! -f /etc/sysconfig/docker-storage ]; then
+    echo "ERROR: $testname: /etc/sysconfig/docker-storage file was not created." >> $LOGS
+    cleanup $vg_name "$devs"
+    return $test_status
+  fi
+
+  source /etc/sysconfig/docker-storage
+  
+  # Search for $extra_options in $options. 
+  echo $DOCKER_STORAGE_OPTIONS | grep -- "$extra_options" >> $LOGS 2>&1
```
rhvgoyal: we probably dont need "--". It seems to work even without it.

rhvgoyal: I think output of grep does not have to be sent to logs.  just send it to /dev/null.

rhvgoyal: May be -q option will do what you are looking to do.

stephen679: Hmm, regarding the "--", I did that because $extra_options would usually have a format of "--xxxxxxx", telling grep it might be an option.

rhvgoyal: That's a good point. I tried it and if extra_options has --, it creates problems.

rhvgoyal:
Apart from minor nit of sending grep output to LOGS, this looks good to me. Fix that and push patch again and I will merge it.

rhvgoyal:
@rh-atomic-bot r+

PR #148, opened 2016-06-23
[merged] error out if docker.service is running before tests

### LINKED ISSUE #147: dss/tests: error out if docker.service is already running
docker.service should not be running/active before running tests, because storage configuration and old meta-data may still be in /var/lib/docker. As a user trying to test docker-storage-setup, it would be helpful if ./run-tests.sh errors out and outputs a message that informs us that docker is running and what actions to take to test from a clean slate, instead of continuing to run tests and obtaining results that the user may not have expected.

stephen679:
Resolves #147 ; ./run-tests.sh should now error out and output a message to user with actions they can take to address active docker: i.e,
  1) Stop the docker service and 2)remove /var/lib/docker/ directory


rhvgoyal:
Hi Stephen,

I think we will have to have 2 checks. We also need to make sure that /var/lib/docker/ dir does not exist or is empty. Otherwise pool setup will fail too.

stephen679:
Good point Vivek. docker.service may not be running, but /var/lib/docker/ could still be non-empty which can contain metadata that can affect thin pool setup.

Will make those changes.

rhvgoyal:
How about using following check as docker-storage-setup is using.

  local docker_devmapper_meta_dir="$DOCKER_METADATA_DIR/devicemapper/metadata/"

  [ ! -d "$docker_devmapper_meta_dir" ] && return 0

If docker is using overlay, we don't have a problem with running tests. (As of now).

stephen679:
@rhvgoyal ok. And when you refer to $DOCKER_METADATA_DIR, do you mean /var/lib/docker?

rhvgoyal:
yes. That's default. though user can change it. I guess there is no easy way to detect it until and unless we go through options in /etc/sysconfig/docker. I think we can do that in a separate PR if need be.

rhvgoyal:
LGTM. Thanks.

rhvgoyal:
@rh-atomic-bot r+

PR #149, opened 2016-06-28
Test 00X: Check DATA_SIZE option [look back into this later]

stephen679:
Test case that checks if docker-storage-setup configures the data LV to the correct size according to user-specified DATA_SIZE option in /etc/sysconfig/docker-storage-setup.


tests/008-test-change-lv-data-size.sh
```diff
@@ -0,0 +1,61 @@
...
+ 
+  # Physical extent information about the VG
+  pe_total=$(vgdisplay --noheading -C -o vg_extent_count $vg_name)
+  pe_free=$(vgdisplay --noheading -C -o vg_free_count $vg_name)
+  pe_used_percentage=$(( (pe_total - pe_free) * 100 / pe_total))
+  
+  # Give some room for error since some rounding may have occured
+  local lower_bound=$(( data_size - 3 ))
+  local upper_bound=$(( data_size + 2 ))
+
```
rhvgoyal: Instrad of doing 5% margin of error, can we calculate available PE in volume group before thin pool creation, and then calculate 70% of that and then make sure used PEs are same as calculated. If 70% is not a integer number then lvm must be either rounding it up or down. So say 70% comes out to be 213.3. Then either 213 or 214 can both the right answer for number of PEs used after thin pool creation.

stephen679:
@rhvgoyal After doing some manual inspecting after calling d-s-s with default 40%FREE, it seems like lvm does not behave as exact as we want to when creating the VG. 

`[root@localhost ~]# vgdisplay
  --- Volume group ---
  VG Name               docker-vg
  System ID             
  Format                lvm2
  Metadata Areas        1
  Metadata Sequence No  6
  VG Access             read/write
  VG Status             resizable
              .
              .
              .
  VG Size               100.00 GiB
  PE Size               4.00 MiB
  Total PE              25599
  Alloc PE / Size       10281 / 40.16 GiB
  Free  PE / Size       15318 / 59.84 GiB`

40% of 25599 is 10239.6, so we should expect either 10239 or 10240 PE's to be used. However you can see that 10281 PE's have been used for docker-vg, which is around 40.1% of the total PE used. Unless we think that this value (40.1%) results from incorrect behavior, it seems like we still need to have a notion of a margin for error; perhaps we could decrease the size of this margin?

rhvgoyal:
@stephen679 can you delete the the thin pool and give me vgdisplay output. I want to see how many PEs are reported in "Free PE" when vg is fully free.

rhvgoyal:
Also, DATA_SIZE only specifies the data volume size. We also allocate metadata volume and these additional PEs must have been for that. May be we just need to check the size of data logical volume.

stephen679:
@rhvgoyal this is what I get when I delete the thin pool:

[root@localhost ~]# lvremove docker-vg/docker-pool
Do you really want to remove active logical volume docker-pool? [y/n]: y
  Logical volume "docker-pool" successfully removed
```
[root@localhost ~]# vgdisplay
  --- Volume group ---
  VG Name               docker-vg
  System ID             
  Format                lvm2
  Metadata Areas        1
  Metadata Sequence No  7
  VG Access             read/write
  VG Status             resizable
  MAX LV                0
  Cur LV                0
  Open LV               0
  Max PV                0
  Cur PV                1
  Act PV                1
  VG Size               100.00 GiB
  PE Size               4.00 MiB
  Total PE              25599
  Alloc PE / Size       0 / 0   
  Free  PE / Size       25599 / 100.00 GiB
  VG UUID               hn94hD-NHdK-jmtB-iawR-IMg8-RRZg-DawZD3
```

stephen679:
Also, 

```
[root@localhost ~]# pvdisplay -m
  --- Physical volume ---
  PV Name               /dev/vdb1
  VG Name               docker-vg
  PV Size               100.00 GiB / not usable 3.00 MiB
  Allocatable           yes 
  PE Size               4.00 MiB
  Total PE              25599
  Free PE               15318
  Allocated PE          10281
  PV UUID               13EYrs-zohl-IjBX-1eSG-Rzwl-dR0D-KJdgMl
   
  --- Physical Segments ---
  Physical extent 0 to 25:
    Logical volume	/dev/docker-vg/docker-pool_tmeta
    Logical extents	0 to 25
  Physical extent 26 to 10254:
    Logical volume	/dev/docker-vg/docker-pool_tdata
    Logical extents	0 to 10228
  Physical extent 10255 to 10280:
    Logical volume	/dev/docker-vg/lvol0_pmspare
    Logical extents	0 to 25
  Physical extent 10281 to 25598:
    FREE
```

stephen679:
Okay I think our solution seems to be:

```
expected_data_PE= $(total_PE - meta_PE) * data_percentage 
if expected_data_PE == total_data_LE  #(give or take 1 LE): 
   test passes
test fails
```

Seems okay to you?

rhvgoyal:
@stephen679 I don't have a good feeling about this test. I am not convinced that this test has to be this complicated. I am checking with lvm developers and figure out if there is an easy way to ensure this. If not, for now, lets put this test on back burner and focus on others.

stephen679:
@rhvgoyal ok that's fine. I admit it was a bit messy trying to produce the results that we wanted based off of PE's. Was it the fact that we couldn't simply extract $data_PE and $meta_PE that concerns you?

rhvgoyal:
@stephen679 What concerns me is that we don't know what kind of rounding off lvm does when it determines X% of free space. Its not exact science. If we don't know what lvm does, writing tests around it can be easily broken down the line or in different configuration etc. No point in writing test cases which can be easily broken. They will give us more headache.

stephen679:
@rhvgoyal Ok that makes sense. While we wait for further information from lvm developers, are there other things you want to see tested that don't really pertain to exact size checking? Most of my remaining tests are related to the function of docker-storage-setup changing the size of the docker VG or pool.

rhvgoyal:
We should be able to test CHUNK_SIZE, AUTO_EXTEND_POOL and WIPE_SIGNATURES options.

stephen679:
this test should not be included until LVM behavior can be confirmed

rhvgoyal:
For now I am closing this. Open this later if need be.

PR #151, opened 2016-07-01
Test 008: check AUTO_EXTEND_POOL

stephen679:
This test checks to make sure that docker-storage-setup handles AUTO_EXTEND_POOL option. If we set AUTO_EXTEND_POOL=no, the docker pool should not autoextend when it hits the autoextend threshold.

I'm not sure if I like the `sleep $wait_time_for_update` line, but some times it just takes a bit of time for information about the docker pool to update, so I think it is reasonable for now.


rhvgoyal:
Closing it. Open it again whenever need be.

PR #152, opened 2016-07-05
[merged] Create tmp files in /run and fix SELinux labels

rhatdan:
https://bugzilla.redhat.com/show_bug.cgi?id=1352713 reports some issues
with SELinux and docker-storage-setup.

This fixes attempt to create the content in a /run instead of on /tmp which
prevents potentially hostile users from attacking the script.

The `mv -Z` calls will guarantee the content gets created with the correct
label.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -245,7 +245,7 @@ write_storage_config_file () {
 $storage_options
 EOF
 
-  mv $DOCKER_STORAGE.tmp $DOCKER_STORAGE
+  mv -Z $DOCKER_STORAGE.tmp $DOCKER_STORAGE
 }
 
 create_metadata_lv() {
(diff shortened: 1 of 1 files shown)


rhatdan:
Going through my old pull requests.  Could @rhvgoyal or @shishir-a412ed Look at this.

shishir-a412ed:
LGTM.

shishir-a412ed:
@rhatdan Your title says make temp files in `/run/docker` however you are creating it in `/run`. One of them needs to be updated, not sure which one. Otherwise LGTM.

rhatdan:
Fixed the label on github, the layer in the pull request was correct.

rhvgoyal:
@rhatdan, went through the bug quickly and could not understand what's the core issue. Can you please summarize the issue here and how this patch fixes it. In fact it would be better if it is part of commit message of the bug so that in future somebody can go through it and understand why this piece of code was changed.

rhatdan:
I updated the pull request with updated comments

rhvgoyal:
LGTM

rhvgoyal:
@rh-atomic-bot r+

PR #153, opened 2016-08-12
Fix problems on update F22->F24

mwilck:
This merge request fixes problems I encountered when updating my system from F22 to F24. I had a DM thinpooldev with a name different from "docker-pool", and I was using the shorthand dm.thinpooldev syntax (which used to be recommended some time ago IIRC). See commits for details.

If docker-storage-setup fails to import a preexisiting storage configuration, it recommends the user to completely delete /var/lib/docker and restart from scratch which is a pretty scary option if you have some productive containers.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -398,7 +398,7 @@ reset_lvm_thin_pool () {
 }
 
 setup_lvm_thin_pool () {
-  local tpool
+  local tpool lv_name
 
   # Check if a thin pool is already configured in /etc/sysconfig/docker-storage.
   # If yes, wait for that thin pool to come up.
@@ -408,6 +408,12 @@ setup_lvm_thin_pool () {
      Info "Found an already configured thin pool $tpool in ${DOCKER_STORAGE}"
      if ! wait_for_dev "$tpool"; then
        Fatal "Already configured thin pool $tpool is not available. If thin pool exists and is taking longer to activate, set DEVICE_WAIT_TIMEOUT to a higher value and retry. If thin pool does not exist any more, remove ${DOCKER_STORAGE} and retry"
+     fi
+     lv_name=${tpool#/dev/mapper/}
+     lv_name=${lv_name#${VG}-}
+     if [[ -n "$lv_name" && x"$lv_name" != x"$POOL_LV_NAME" ]]; then
+         Info "overriding default Pool name: $lv_name"
+         POOL_LV_NAME=$lv_name
      fi
   fi
 
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -373,6 +373,10 @@ get_configured_thin_pool() {
   for opt in $options; do
     if [[ $opt =~ dm.thinpooldev* ]];then
       tpool=${opt#*=}
+      case "$tpool" in
+      /dev/mapper/*) ;;
+      *) tpool="/dev/mapper/$tpool";;
+      esac
       echo "$tpool"
       return 0
     fi


rhatdan:
@rhvgoyal PTAL

rhvgoyal:
For the first patch, I don't understand that why did you change the thin pool device name. If you did, then its not docker-storage-setup problem. How would docker-storage-setup know that this new thin pool belongs to docker and not in use by something else?

rhvgoyal:
I don't understand the dm.thinpooldev change either. dss uses /dev/mapper/<pool-device> syntax and docker daemon understands and expects it. So what's wrong.

mwilck:
@rhvgoyal 

1. I didn't change the device name, it was different from the default right from the start (this volume 
pre-dates docker-storage-setup and has survived Fedora updates at least since F20; I doubt that at the time there was any convention to call this volume "docker-pool"). Anyway, the dss code first checks for an already configured volume and then bails out if that volume has a different name than expected, and does that with a less-than-helpful error message. IMO that's a bug. You might as well skip the check for exisiting configuration altogether and just check for "docker-pool".

2. Not sure why you say docker daemon "expects" /dev/mapper/"; it works just fine without it. IIRC the syntax without /dev/mapper used to be recommended in older docker versions but I find it difficult to prove because the old docs are difficult to find. Anyway, this is a similar problem as 1): an old setup that used to work fine and was broken by the update to F24 (perhaps F23 already - I did the F23 and F24 updates quickly after each other and didn't test docker in the F23 stage).

The patches are simple and applying them might save some users (like myself) some hair-pulling.

rhvgoyal:
We check for configured thin pool in /etc/sysconfig/docker-storage. This is file supposed to be owned by docker-storage-setup. Anyway, how does your /etc/sysconfig/docker-storage file look like?

Who generated the thin pool and who created the contents in /etc/sysconfig/docker-storage? Did you do it manually yourself?

rhvgoyal:
Can you also paste the error message you are facing dss bails out.

rhvgoyal:
Would be good if you can provide all the messages from dss (info as well as error)

rhvgoyal:
I don't want to take over file created by user. /etc/syconfig/docker-storage pre-dates dss, I think we should simply detect it (whereever we can) and bail out and not do anything.

mwilck:
Here is /etc/sysconfig/docker-storage from my F22 installation:

`DOCKER_STORAGE_OPTIONS="--storage-driver devicemapper --storage-opt dm.thinpooldev=vg-thin_whale"
`
The unpatched dss script fails as follows: 
```
INFO: Found an already configured thin pool vg-thin_whale in /etc/sysconfig/docker-storage
INFO: Waiting for device vg-thin_whale to be available. Wait time remaining is 60 seconds
INFO: Waiting for device vg-thin_whale to be available. Wait time remaining is 55 seconds
INFO: Waiting for device vg-thin_whale to be available. Wait time remaining is 50 seconds
INFO: Waiting for device vg-thin_whale to be available. Wait time remaining is 45 seconds
INFO: Waiting for device vg-thin_whale to be available. Wait time remaining is 40 seconds
INFO: Waiting for device vg-thin_whale to be available. Wait time remaining is 35 seconds
INFO: Waiting for device vg-thin_whale to be available. Wait time remaining is 30 seconds
INFO: Waiting for device vg-thin_whale to be available. Wait time remaining is 25 seconds
INFO: Waiting for device vg-thin_whale to be available. Wait time remaining is 20 seconds
INFO: Waiting for device vg-thin_whale to be available. Wait time remaining is 15 seconds
INFO: Waiting for device vg-thin_whale to be available. Wait time remaining is 10 seconds
INFO: Waiting for device vg-thin_whale to be available. Wait time remaining is 5 seconds
INFO: Timed out waiting for device vg-thin_whale
ERROR: Already configured thin pool vg-thin_whale is not available. If thin pool exists and is taking longer to activate, set DEVICE_WAIT_TIMEOUT to a higher value and retry. If thin pool does not exist any more, remove /etc/sysconfig/docker-storage and retry

```

If I replace the LV name in the sysconfig above by  "/dev/mapper/vg-thin-whale", the following error occurs:

```
INFO: Found an already configured thin pool /dev/mapper/vg-thin_whale in /etc/sysconfig/docker-storage
INFO: Device node /dev/mapper/vg-thin_whale exists.
ERROR: Docker has been previously configured for use with devicemapper graph driver. Not creating a new thin pool as existing docker metadata will fail to work with it. Manual cleanup is required before this will succeed.
INFO: Docker state can be reset by stopping docker and by removing /var/lib/docker directory. This will destroy existing docker images and containers and all the docker metadata.
```

I found especially the last paragraph pretty disconcerting. That's why I posted my patches.

rhvgoyal:
@mwilck I created a PR #156 to not touch the thin pool which are not managed by dss. I think this should leave your setup untouched over upgrade. Does this work for you?

mwilck:
@rhvgoyal, I am not certain if this will work - the question is whether docker will be able to access its storage volumes if dss simply doesn't touch the setup at all? I thought dss was now a prerequisite for running docker on Fedora... But I admit I haven't studies how the dependencies work.

rhvgoyal:
@mwilck I think your current setup will continue to work (if it was working before the upgrade). Docker reads storage setting from /etc/sysconfig/docker-storage. If user has already modified that file, then dss will not touch it.

Ideally we don't want users to touch /etc/sysconfig/docker-storage at all. All the modifications should be done to /etc/sysconfig/docker-storage-setup file and let dss manage /etc/sysconfig/docker-storage.

mwilck:
@rhvgoyal: OK, got it. Do you want me to test the new version on my system?

rhvgoyal:
@mwilck Yes, please do test thew new version.

rhvgoyal:
Closing it. We don't support old data and metadata mode.

PR #155, opened 2016-08-22
[merged] Detect that overlay2 is currently configured driver

rhvgoyal:
Right now even if overlay2 is currently configured driver, dss detects it
as overlay (and not overlay2).

This truncation happens due the way we have written regular expression
for matching pattern. It works for driver name as long as it contains
all small letters but not numbers. This patch fixes it.

Now currently configured driver is detected correctly and unnecessary
errors are avoided.


rhvgoyal:
cc @rhatdan @cgwalters

cgwalters:
@rh-atomic-bot r+ e30eb87

PR #156, opened 2016-08-29
[merged] Do not try to manage thin pool not created by dss

rhvgoyal:
Fixes #153 

In general /etc/sysconfig/docker-storage is managed by dss and user
should not edit it directly.

In the past there have been cases where this was not a strict
requirement. And looks like over upgrade we ran into a case
where a thin pool configured by user is not in same format
as dss expects it to be and various assumptions are broken
and errors happen.

Try to parse thin pool name and make sure it is in same
format as generated by dss and if it is not, exit early
and stop doing further processing.

This should allow user to continue with its setup.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -413,7 +413,15 @@ setup_lvm_thin_pool () {
   tpool=`get_configured_thin_pool`
 
   if [ -n "$tpool" ]; then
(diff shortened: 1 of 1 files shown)


rhatdan:
@rh-atomic-bot r+

rhvgoyal:
Just noticed a typo. Instead of "seem" in error message there is "see". m is missing. Will generate another PR to fix it.

PR #157, opened 2016-08-29
[merged] Fix error message

rhvgoyal:
There is a typo in error message. It should be "seem" instead of "see".


rhatdan:
@rh-atomic-bot r+

PR #159, opened 2016-09-28
[merged] Enable dm.use_deferred_deletion only if it is supported by the underlying kernel.

shishir-a412ed:

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -63,6 +63,10 @@ STORAGE_DRIVERS="devicemapper overlay overlay2"
 
 DOCKER_METADATA_DIR="/var/lib/docker"
 
+PIPE1=/run/dss-fifo1
+PIPE2=/run/dss-fifo2
+TEMPDIR=$(mktemp --tmpdir -d)
+
 # DEVS can have device names without absolute path. Convert these to absolute
 # paths and save in ABS_DEVS and use in rest of the code.
 DEVS_ABS=""
@@ -111,10 +115,49 @@ get_deferred_deletion_string() {
   [ -z "$major" ] && return 0
   [ -z "$minor" ] && return 0
 
-  # docker 1.9 onwards supports deferred device deletion. Enable it.
-  if [ $major -gt 1 ] || ([ $major -eq 1 ] && [ $minor -ge 9 ]);then
-    echo "--storage-opt dm.use_deferred_deletion=true"
+  if should_enable_deferred_deletion $major $minor; then
+     echo "--storage-opt dm.use_deferred_deletion=true"
   fi
+}
+
+should_enable_deferred_deletion() {
+   # docker 1.9 onwards supports deferred device deletion. Enable it.
+   local major=$1
+   local minor=$2
+   if [ $major -lt 1 ] || ([ $major -eq 1 ] && [ $minor -lt 9 ]);then
+      return 1
+   fi
+   if platform_supports_deferred_deletion; then
+      return 0
+   fi
+   return 1
+}
+
+platform_supports_deferred_deletion() {
+        local deferred_deletion_supported=1
+        trap cleanup_pipes EXIT
+        mkfifo $PIPE1
+        mkfifo $PIPE2
+        mount -o bind $TEMPDIR $TEMPDIR
+        if [ ! -x "/usr/lib/docker-storage-setup/dss-child-read-write" ];then
+           return 1
+        fi
+        unshare -m /usr/lib/docker-storage-setup/dss-child-read-write $PIPE1 $PIPE2 &
+        read -t 10 n <>$PIPE1
+        if [ "$n" != "start" ];then
+	   return 1
+        fi
+        umount $TEMPDIR
+        rmdir $TEMPDIR > /dev/null 2>&1
+        deferred_deletion_supported=$?
+        echo "finish" > $PIPE2
+        return $deferred_deletion_supported
+}
+
+cleanup_pipes(){
+    rm -f $PIPE1
+    rm -f $PIPE2
+    rmdir $TEMPDIR 2>/dev/null
 }
 
 extra_options_has_dm_fs() {
--- a/docker-storage-setup.spec
+++ b/docker-storage-setup.spec
@@ -12,6 +12,7 @@ Source1:        docker-storage-setup.service
 Source2:        docker-storage-setup.conf
 Source3:        docker-storage-setup-override.conf
 Source4:        libdss.sh
+Source5:        dss-child-read-write.sh
 
 BuildRequires:  pkgconfig(systemd)
 
@@ -39,6 +40,7 @@ install -d %{buildroot}%{_sysconfdir}/sysconfig/
 install -p -m 644 %{SOURCE3} %{buildroot}%{_sysconfdir}/sysconfig/docker-storage-setup
 install -d %{buildroot}/%{dsslibdir}
 install -p -m 755 %{SOURCE4} %{buildroot}/%{dsslibdir}/libdss.sh
+install -p -m 755 %{SOURCE5} %{buildroot}/%{dsslibdir}/dss-child-read-write
 
 %post
 %systemd_post %{name}.service
--- a/dss-child-read-write.sh
+++ b/dss-child-read-write.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+# This is a helper script which is called by docker-storage-setup.sh (d-s-s).
+# This script helps in providing synchronization primitives to d-s-s so that
+# d-s-s can determine whether deferred deletion is supported by the underlying
+# kernel or not.
+
+# $1 is named FIFO pipe.
+# This helper script will write to $1 to signal d-s-s that unshare has been completed successfully.
+echo "start" > $1
+# $2 is another named FIFO pipe.
+# This helper script will read from $2. The write for this pipe would come from d-s-s to indicate
+# that helper script can terminate now.
+read -t 10 n <>$2


shishir-a412ed:
ping @rhvgoyal

docker-storage-setup.sh
```diff
@@ -111,10 +111,48 @@ get_deferred_deletion_string() {
...
+}
+
+should_enable_deferred_deletion() {
+   # docker 1.9 onwards supports deferred device deletion. Enable it.
+   if [ $major -gt 1 ] || ([ $major -eq 1 ] && [ $minor -ge 9 ]);then
+      if platform_supports_deferred_deletion; then
+	 return 0
+      fi
+   fi
+   return 1
```
rhvgoyal: How about getting rid of nested if. Instead first check if platform supports deletion or not. If not, return immediately. If platform does check, then check the docker version

docker-storage-setup.sh
```diff
@@ -111,10 +111,48 @@ get_deferred_deletion_string() {
...
-  fi
+  if should_enable_deferred_deletion; then
+     echo "--storage-opt dm.use_deferred_deletion=true"
+  fi
+}
+
+should_enable_deferred_deletion() {
+   # docker 1.9 onwards supports deferred device deletion. Enable it.
+   if [ $major -gt 1 ] || ([ $major -eq 1 ] && [ $minor -ge 9 ]);then
+      if platform_supports_deferred_deletion; then
```
rhvgoyal: Variables major and minor are local to parent function. How does this check work here?

shishir-a412ed: https://gist.github.com/shishir-a412ed/7f294d66e8b461e76ba558becf01502b

rhvgoyal: In your example,  variables are not "local". But these variables are local to parent function in docker-storage-setup.sh.

`local version major minor`

docker-storage-setup.sh
```diff
@@ -111,12 +111,51 @@ get_deferred_deletion_string() {
...
+      return 1
+   fi
+   if platform_supports_deferred_deletion; then
+      return 0
+   fi
+   return 1
+}
+
+platform_supports_deferred_deletion() {
+        local deferred_deletion_not_supported=false
```
rhvgoyal: how about shorter name, deferred_deletion_supported.

docker-storage-setup.sh
```diff
@@ -111,12 +111,51 @@ get_deferred_deletion_string() {
...
+        local pipe1=/tmp/fifo1
+        local pipe2=/tmp/fifo2
+        mkfifo $pipe1
```
rhvgoyal: lets use mktemp to generate names

docker-storage-setup.sh
```diff
@@ -111,12 +111,51 @@ get_deferred_deletion_string() {
...
+        fi
+        echo "pipe2" > $pipe2
+        trap cleanup EXIT
+        if [ "$deferred_deletion_not_supported" = true ];then
+            return 1
+        else
+	    return 0
+        fi
+}
+
```
rhvgoyal: Can we simplify structure a bit. May be something like as follows.

local deferred_removal_supported

rmdir /tmp/foo > /dev/null
deferred_removal_supported = $?

<Do the cleanup>

return $deferred_removal_supported

docker-storage-setup.sh
```diff
@@ -111,12 +111,48 @@ get_deferred_deletion_string() {
...
+        umount $tmpDir
+        rmdir $tmpDir > /dev/null 2>&1
+        deferred_deletion_supported=$?
+        echo "finish" > $pipe2
+        trap cleanup_pipes EXIT
+        return $deferred_deletion_supported
+}
+
+cleanup_pipes(){
+    rm -f /tmp/dss-fifo1
```
rhvgoyal: Let us make /tmp/dss-fifo1 and /tmp/dss-fifo2 global variables and use both in cleanup_pipes and platform_supports_deferred_deletion(). This will avoid hardcoding file name in two functions.

Also let us make tmpDIR global "TEMPDIR".

docker-storage-setup.sh
```diff
@@ -111,12 +111,48 @@ get_deferred_deletion_string() {
...
+        mkfifo $pipe2
+        tmpDir=$(mktemp --tmpdir -d)
+        mount -o bind $tmpDir $tmpDir
+        unshare -m ./dss-child-read-write.sh $pipe1 $pipe2 &
+        read n < $pipe1
+        umount $tmpDir
+        rmdir $tmpDir > /dev/null 2>&1
+        deferred_deletion_supported=$?
+        echo "finish" > $pipe2
+        trap cleanup_pipes EXIT
```
rhvgoyal: Should we declare this trap early. Say right before mkfifo?

rhatdan:
Put your fifos on /run not on /tmp and make sure they are only writable by root.

docker-storage-setup.sh
shishir-a412ed: Fixed.

docker-storage-setup.sh
shishir-a412ed: Fixed.

docker-storage-setup.sh
shishir-a412ed: Fixed.

docker-storage-setup.sh
shishir-a412ed: Fixed.

docker-storage-setup.sh
shishir-a412ed: Fixed.

docker-storage-setup.sh
shishir-a412ed: Fixed. Also moved named pipes to `/run` as @rhatdan suggested.

docker-storage-setup.sh
shishir-a412ed: Fixed.

shishir-a412ed:
@rhvgoyal PTAL.

docker-storage-setup.sh
```diff
@@ -111,12 +115,45 @@ get_deferred_deletion_string() {
...
+   return 1
+}
+
+platform_supports_deferred_deletion() {
+        local deferred_deletion_supported=1
+        trap cleanup_pipes EXIT
+        mkfifo $PIPE1
+        mkfifo $PIPE2
+        mount -o bind $TEMPDIR $TEMPDIR
+        unshare -m ./dss-child-read-write.sh $PIPE1 $PIPE2 &
```
rhvgoyal: I think this patch is almost there. One minor nit. I don't think we should hardcode the path of the dss-child-read-write.sh. May be let it be found in $PATH. So just call it directly witout `./`

unshare -m dss-child-read-write.sh $PIPE1 $PIPE2 &.

Another concern is what if this program is not installed or missing. We probably want to capture error code. Otherwise next "read" call will block infinitely.

shishir-a412ed: @rhvgoyal we can do a `stat` on the file (`dss-child-read-write.sh`). if the return code (`$?`) is `0` i.e file exists, then we `unshare`, else we can just return.

rhvgoyal:
@shishir-a412ed We aready have a test for /usr/bin/growpart in the dss script. I think lets do a similar test.

[  ! -x /usr/bin/dss-child-read-write.sh ]  && return 1

rhvgoyal:
@shishir-a412ed Also bash `read` seems to support the timeout option. `-t`. Let us use a timeout of say 10 seconds. If we don't hear back from child in 10 seconds, return error (instead of blocking infinitely).

shishir-a412ed:
@rhvgoyal PTAL.

docker-storage-setup.sh
```diff
@@ -111,12 +115,48 @@ get_deferred_deletion_string() {
...
+        mount -o bind $TEMPDIR $TEMPDIR
+        if [ ! -x "/usr/bin/dss-child-read-write.sh" ];then
+           return 1
+        fi
+        unshare -m /usr/bin/dss-child-read-write.sh $PIPE1 $PIPE2 &
+        read -t 10 n <>$PIPE1
+        umount $TEMPDIR
+        rmdir $TEMPDIR > /dev/null 2>&1
+        deferred_deletion_supported=$?
+        echo "finish" > $PIPE2
```
rhvgoyal: Given we introduced timeout logic, we will have to differentiate between whether we timed out or child sent us an message. If we timed out we want to return "1". Current code will return 0.

May be read the value of the variable after read and if it is "start", then we know child sent us a valid message.

shishir-a412ed: Fixed.

shishir-a412ed:
@rhvgoyal PTAL.

docker-storage-setup.sh
```diff
@@ -111,12 +115,51 @@ get_deferred_deletion_string() {
...
+   return 1
+}
+
+platform_supports_deferred_deletion() {
+        local deferred_deletion_supported=1
+        trap cleanup_pipes EXIT
+        mkfifo $PIPE1
+        mkfifo $PIPE2
+        mount -o bind $TEMPDIR $TEMPDIR
+        if [ ! -x "/usr/bin/dss-child-read-write.sh" ];then
```
rhvgoyal: One more minor nit. We probably should get rid of extension ".sh" while calling this. Most likely we will install it as /usr/bin/dss-child-read-write (like other binaries there)

shishir-a412ed: Fixed.

rhvgoyal: @cgwalters That's a good point. Putting this in /usr/lib/docker-storage-setup/ makes more sense.

rhvgoyal:
@shishir-a412ed can we also update docker-storage-setup.spec file to accomodate new binary dss-child-read-write.sh

shishir-a412ed:
@rhvgoyal PTAL.

docker-storage-setup.spec
```diff
@@ -39,6 +40,7 @@ install -d %{buildroot}%{_sysconfdir}/sysconfig/
 install -p -m 644 %{SOURCE3} %{buildroot}%{_sysconfdir}/sysconfig/docker-storage-setup
 install -d %{buildroot}/%{dsslibdir}
 install -p -m 755 %{SOURCE4} %{buildroot}/%{dsslibdir}/libdss.sh
+install -p -m 755 %{SOURCE5} %{buildroot}/%{dsslibdir}/dss-child-read-write.sh
```
rhvgoyal: I think this should be 

`install -p -m 755 %{SOURCE5} %{buildroot}/%{dsslibdir}/dss-child-read-write`

shishir-a412ed: Fixed.

rhvgoyal:
LGTM. Thanks Shishir.

rhvgoyal:
@rh-atomic-bot r+

PR #160, opened 2016-10-11
[merged] Cleanup handling of Overlay2 file systems

rhatdan:


rhatdan:
@rhvgoyal PTAL

docker-storage-setup.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 #--
-# Copyright 2014-2015 Red Hat, Inc.
+# Copyright 2014-2016 Red Hat, Inc.
```
rhvgoyal: Should we update existing line or add another one for 2016. I am not sure typically what's the convention.

docker-storage-setup.sh
```diff
@@ -240,12 +236,8 @@ write_storage_config_file () {
...
       return 1
     fi
-  elif [ "$STORAGE_DRIVER" == "overlay" ];then
-    if ! storage_options=$(get_overlay_config_options); then
-      return 1
-    fi
-  elif [ "$STORAGE_DRIVER" == "overlay2" ];then
-    if ! storage_options=$(get_overlay2_config_options); then
+  else
+    if ! storage_options=$(get_config_options $STORAGE_DRIVER); then
```
rhvgoyal: This seems to be partial cleanup. We have another function get_devicemapper_config_options() and get_config_options() seems to imply that it works for all graph drivers.

So either create get_config_options() which works for all graph drivers, or rename this function to something which represents that this is overlay specific.

docker-storage-setup.sh
```diff
@@ -895,8 +883,8 @@ setup_storage() {
   # Set up lvm thin pool LV
   if [ "$STORAGE_DRIVER" == "devicemapper" ]; then
     setup_lvm_thin_pool
-  elif [[ "$STORAGE_DRIVER" == "overlay" || "$STORAGE_DRIVER" == "overlay2" ]];then
-    setup_overlay
+  else
```
rhvgoyal: I liked that original elif condition. That is we write config file only if graph driver is overlay or overlay2. Now we are putting blindly in else loop so it will be true for any value other than devicemapper. While it is true as of now, it might not be true in future when we support more graph drivers.

rhatdan: I look at the devicemapper as the one off and everything else uses the generic code. At least that was what I was going for.  If we add more graphdrivers, then either they can use the default or we make a graphdriver specific patch.

docker-storage-setup.sh
```diff
@@ -940,30 +927,32 @@ if [ -e /etc/sysconfig/docker-storage-setup ]; then
...
-# Read mounts
-ROOT_DEV=$( awk '$2 ~ /^\/$/ && $1 !~ /rootfs/ { print $1 }' /proc/mounts )
-if ! ROOT_VG=$(lvs --noheadings -o vg_name $ROOT_DEV 2>/dev/null);then
-  Info "Volume group backing root filesystem could not be determined"
-  ROOT_VG=
-else
-  ROOT_VG=$(echo $ROOT_VG | sed -e 's/^ *//' -e 's/ *$//')
-fi
+if [ "$STORAGE_DRIVER" == "devicemapper" ]; then
+    # Read mounts
```
rhvgoyal: There are bunch of functionalities this script does which can be independent of driver. For example, growing pvs, growing rootfs, adding devices to volume group etc.

I am especially worried about the use case of growing pv, and rootfs with atomic image in cloud. I would say that let us keep all this in a separate PR and discuss there what can we do.

I think proper solution to this issue is to move all this growing root functionality out of docker-sotrage-setup (may be in cloud-init) and then dss is only responsible for setting up storage for docker and things will become easy.

For now, If you can just retain overlay cleanup related changes, that would be fine.

rhatdan:
@rhvgoyal Only changed Overlay code.

rhvgoyal:
LGTM

rhvgoyal:
@rh-atomic-bot r+

PR #161, opened 2016-10-24
[merged] systemd: Order After=cloud-init.service, not cloud-final

cgwalters:
See https://lists.fedoraproject.org/archives/list/cloud@lists.fedoraproject.org/thread/TRT6A5ZCCMEC2LNNAAEGUZYJV4P7MEB7/

At least in recent cloud-init, the previous situation created
an ordering cycle.  We really just want to run after the primary
configuration run, not after all scripts.

Should fix Fedora 25 Atomic Host startup.

### DIFF
--- a/docker-storage-setup.service
+++ b/docker-storage-setup.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=Docker Storage Setup
-After=cloud-final.service
+After=cloud-init.service
 Before=docker.service
 
 [Service]


cgwalters:
The ordering here is like how `cloud-init.service` is `Before=sshd.service`.  We're doing the same thing basically.

rhvgoyal:
I am curious to know what changed. Why was it working so far and broke suddenly.

I am not sure what happens between cloud-init and cloud-final and what it can possibly break due to change of this dependency.

rhvgoyal:
Not sure where is official code for cloud-init. I found one source tree at `git.launchpad.net/cloud-init` and following commit there changed cloud-final.service to create dependency on multi-user.target. And that seems to have created this issue.

commit 34a26f7f59f2963691e36ca0476bec9fc9ccef63
Author: Scott Moser <smoser@brickies.net>
Date:   Thu Sep 8 13:17:37 2016 -0400

    systemd: Better support package and upgrade.

cgwalters:
Yeah, I think you're right.  That cloud-init change looks...hacky at best, but regardless of the merits of that change, I think this one is correct to do as well.

rhvgoyal:
Ok, lets make this change. I am little concerned that it will break something. We might have to revisit this if somebody complains.

rhvgoyal:
@rh-atomic-bot r+

rhatdan:
@lsm5 make sure on the next docker build you update docker-storage-setup from master

PR #163, opened 2016-11-01
[merged] Cleanup temporary mount point upon failure

### LINKED ISSUE #162: Mount leak in platform_supports_deferred_deletion()
Line 141:

    mount -o bind $TEMPDIR $TEMPDIR

...but lines 143 & 148 return without umounting. System is then left with stuck mounts:

    /dev/vda1 on /tmp/tmp.vBynTZ78If type xfs (rw,relatime,seclabel,attr2,inode64,noquota)

rhvgoyal:
Fixes #162 

platform_supports_deferred_deletion() mounts tempdir and then can exit
due to error conditions and never unmount that tempdir.  Do an unmount
during exit to do the cleanup.

Also make a error check little early. First do the check and then
create pipes and mount point.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -136,18 +136,16 @@ should_enable_deferred_deletion() {
 platform_supports_deferred_deletion() {
         local deferred_deletion_supported=1
         trap cleanup_pipes EXIT
-        mkfifo $PIPE1
-        mkfifo $PIPE2
-        mount -o bind $TEMPDIR $TEMPDIR
         if [ ! -x "/usr/lib/docker-storage-setup/dss-child-read-write" ];then
            return 1
         fi
-        unshare -m /usr/lib/docker-storage-setup/dss-child-read-write $PIPE1 $PIPE2 &
+        mkfifo $PIPE1
+        mkfifo $PIPE2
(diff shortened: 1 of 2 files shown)


docker-storage-setup.sh
```diff
@@ -136,12 +136,14 @@ should_enable_deferred_deletion() {
...
-        mount -o bind $TEMPDIR $TEMPDIR
+
         if [ ! -x "/usr/lib/docker-storage-setup/dss-child-read-write" ];then
            return 1
         fi
+
+        mkfifo $PIPE1
+        mkfifo $PIPE2
+        mount -o bind $TEMPDIR $TEMPDIR
         unshare -m /usr/lib/docker-storage-setup/dss-child-read-write $PIPE1 $PIPE2 &
```
cgwalters: We could rely on the kernel to umount if we did the mount inside the unshare, no?

rhvgoyal: That's a good point. We should not have to do mounting in host mount namespace. We should be able to create new mount namespace, do the mount and then signal to parent that try to remove directory from host mount namespace. 

Only extra overhead will be that we will have to pass the path to the tmp directory to child process. That should be fine. 

Will try it.

rhvgoyal:
@cgwalters PTAL. Now I am creating bind mount inside child process mount namespace.

rhatdan:
LGTM

cgwalters:
@rh-atomic-bot r+ 697d6ff

PR #164, opened 2016-11-02
[merged] Add Makefile

rhatdan:


edsantiago:
LGTM but you might want to remove the `Source0 - Source5` declarations in the specfile

rhatdan:
Removed Source*

rhvgoyal:
This should install binaries in /usr/local/ instead of /usr ?

Who defines DESTDIR

rhatdan:
It is defined in the RPM.  Or anyone who installs it.

Whether it installs to /usr/local versus /usr is debatable.  Especially where the /etc files are shred.

Makefile
```diff
@@ -0,0 +1,27 @@
...
+
+.PHONY: install
+install:
+
+	install -D -m 755 docker-storage-setup.sh ${BINDIR}/docker-storage-setup
+	install -D -m 644 docker-storage-setup.service ${UNITDIR}/docker-storage-setup.service
+	install -D -m 644 docker-storage-setup.conf ${DSSLIBDIR}/docker-storage-setup
+	install -D -m 644 docker-storage-setup-override.conf ${SYSCONFDIR}/docker-storage-setup
+	install -D -m 755 libdss.sh ${DSSLIBDIR}/libdss.sh
+	install -D -m 755 dss-child-read-write.sh ${DSSLIBDIR}/dss-child-read-write
```
rhatdan: Most likely not going to matter, since this is more about packaging in an RPM which is not going to preserve timestamps.  I don't think these are usually preserved in a Makefile.

rhvgoyal:
Looks like shishir forgot to add `dss-child-read-write` to `%files` section in `docker-storage-setup.spec` file. We will have to fix it. Either as part of this PR or a separate PR.

Makefile
```diff
@@ -0,0 +1,27 @@
...
+clean:
+	-rm -rf *~ \#* .#*
+
+.PHONY: install
+install:
+
+	install -D -m 755 docker-storage-setup.sh ${BINDIR}/docker-storage-setup
+	install -D -m 644 docker-storage-setup.service ${UNITDIR}/docker-storage-setup.service
+	install -D -m 644 docker-storage-setup.conf ${DSSLIBDIR}/docker-storage-setup
+	install -D -m 644 docker-storage-setup-override.conf ${SYSCONFDIR}/docker-storage-setup
```
rhvgoyal: Don't like the idea of overwriting user's config file `/etc/sysconfig/docker-storage-setup`. Can't we do this only if /etc/sysconfig/docker-storage-setup is not present.

rhatdan:
That can be done in the spec file  Not usually done in the Makefile.

rhvgoyal:
spec file already has it.

`%config(noreplace) %{_sysconfdir}/sysconfig/docker-storage-setup`

I think it would be a real surprise to overwrite user's config file. I don't expect `make install` to overwrite my config files. Do you have any examples of other packages which do it.

rhatdan:
Do you have any examples of packages that don't?

rhatdan:
We can add a -b to the install line which will backup the older version.

rhvgoyal:
I think something like this will make sense. Overwrite /etc/sysconfig/docker-storage-setup only if it does not exist.

https://github.com/openssh/openssh-portable/blob/master/Makefile.in#L346
`

rhatdan:
@rhvgoyal PTAL

docker-storage-setup.spec
```diff
@@ -54,9 +38,12 @@ install -p -m 755 %{SOURCE5} %{buildroot}/%{dsslibdir}/dss-child-read-write
 %files
 %{_unitdir}/docker-storage-setup.service
 %{_bindir}/docker-storage-setup
-%{dsslibdir}/docker-storage-setup
+%config(noreplace) %{_sysconfdir}/sysconfig/docker-storage-setup-override
```
rhvgoyal: I thought %files section is supposed to keep a list of files which are installed by package on system. And `docker-storage-setup-override` is not installed by package on the host.

docker-storage-setup.spec
```diff
@@ -54,9 +38,12 @@ install -p -m 755 %{SOURCE5} %{buildroot}/%{dsslibdir}/dss-child-read-write
...
 %config(noreplace) %{_sysconfdir}/sysconfig/docker-storage-setup
+%dir %{dsslibdir}
```
rhvgoyal: Why do we need `%dir`. Following webpage seems to say, 

`if a directory is specified in the %files list, the contents of that directory, and the contents of every directory under it, will automatically be included in the package`

http://www.rpm.org/max-rpm-snapshot/s1-rpm-inside-files-list-directives.html

And we don't want that. We are mentioning each file individually.

docker-storage-setup.spec
```diff
@@ -54,9 +38,12 @@ install -p -m 755 %{SOURCE5} %{buildroot}/%{dsslibdir}/dss-child-read-write
 %files
 %{_unitdir}/docker-storage-setup.service
 %{_bindir}/docker-storage-setup
-%{dsslibdir}/docker-storage-setup
```
rhvgoyal: Why are we removing above. We put default config options in /usr/lib/docker-storage-setup/ and want to overwrite it on all package upgrades.

docker-storage-setup.spec
```diff
@@ -30,17 +24,7 @@ as the root logical volume and partition table.
 %build
 
 %install
-install -d %{buildroot}%{_bindir}
```
rhvgoyal: There is no equivalent of `_bindir` though as we are hardcoding it to `bin`

docker-storage-setup.spec
```diff
@@ -30,17 +24,7 @@ as the root logical volume and partition table.
...
-install -p -m 755 %{SOURCE0} %{buildroot}%{_bindir}/docker-storage-setup
-install -d %{buildroot}%{_unitdir}
-install -p -m 644 %{SOURCE1} %{buildroot}%{_unitdir}
```
rhvgoyal: If we are getting rid of macros like _unitdir, are we not going to be broken if tomorrow systemd changes location of unit files and this macro starts expanding to a different value.

rhvgoyal:
LGTM

rhvgoyal:
@rh-atomic-bot r+

PR #165, opened 2016-11-02
Add dss-child-read-write to %files section

rhvgoyal:
Looks like we missed adding dss-child-read-write file to %files section
in spec file.


rhvgoyal:
cc @rhatdan @cgwalters

rhatdan:
I have added this already to my patch, and this is not the spec file that docker-storage-setup actually uses.

rhvgoyal:
Other PR, made this redundant.

PR #166, opened 2016-11-02
[merged] Makefile: add docker var

runcom:
This change is needed for `docker-latest` pkgs in Fedora and RHEL since we're now using `make install` (we'll rename files to `docker-latest-*` and need a way to install with `docker-latest-*` prefix as well):
```sh
make install DESTDIR=%{buildroot} DOCKER=%{name} # where name == docker-latest
```
@rhatdan @rhvgoyal PTAL

Signed-off-by: Antonio Murdaca <runcom@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+DOCKER ?= docker
 # Installation directories.
 PREFIX ?= $(DESTDIR)/usr
 BINDIR ?= $(PREFIX)/bin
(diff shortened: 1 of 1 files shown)


rhvgoyal:
LGTM

rhvgoyal:
@rh-atomic-bot r+

PR #167, opened 2016-11-08
Fix docker-storage-setup to run within a SPC container

rhatdan:
If $HOST is turned on, run LVM commands in with the chroot $HOST command.
This allows us to run docker-storage-setup in a docker container, but
guarantees that we use the hosts commands to prevent version mismatch.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -22,6 +22,41 @@
 
 set -e
 
+export GROWPART=/usr/bin/growpart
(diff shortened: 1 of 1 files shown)


rhvgoyal:
@rhatdan tests are failing.

rhatdan:
https://github.com/projectatomic/docker-storage-setup/pull/173 Replaces this, and is a lot simpler.

PR #168, opened 2016-11-09
[merged] Follow device symlinks in DEVS

ianabc:
This patch just checks if the content of DEVS is a symlink and if it is, follows it.

I'm adding this to avoid problems I've had with devices being renamed when they are reattached in openstack (but the problem is more general).

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -628,8 +628,16 @@ check_wipe_block_dev_sig() {
 canonicalize_block_devs() {
   local devs=$1 dev
   local devs_abs dev_abs
+  local dest_dev
 
   for dev in ${devs}; do
+    # If the device name is a symlink, follow it and use the target
+    if [ -h "$dev" ];then
+      if ! dest_dev=$(readlink -e $dev);then
+        Fatal "Failed to resolve symbolic link $dev"
+      fi
+      dev=$dest_dev
+    fi
     # Looks like we allowed just device name (sda) as valid input. In
     # such cases /dev/$dev should be a valid block device.
     dev_abs=$dev
--- a/tests/009-test-follow-symlinked-devices.sh
+++ b/tests/009-test-follow-symlinked-devices.sh
@@ -1,0 +1,48 @@
+source $SRCDIR/libtest.sh
+
+test_follow_symlinked_devices() {
+  local devs dev
+  local devlinks devlink
+  local test_status=1
+  local testname=`basename "$0"`
+  local vg_name="dss-test-foo"
+
+  # Create a symlink for a device and try to follow it
+  for dev in $TEST_DEVS; do
+    if [ ! -h $dev ]; then
+      devlink="/tmp/$(basename $dev)-test.$$"
+      ln -s $dev $devlink
+
+      dev=$devlink
+      devlinks="$devlinks $dev"
+    fi
+    devs="$devs $dev"
+    echo "Using symlinke devices: $dev -> $(readlink -e $dev)" >> $LOGS 
+  done
+
+  cat << EOF > /etc/sysconfig/docker-storage-setup
+DEVS="$devs"
+VG=$vg_name
+EOF
+  # Run docker-storage-setup
+  $DSSBIN >> $LOGS 2>&1
+
+  # Test failed.
+  if [ $? -ne 0 ]; then
+    cleanup_soft_links "$devlinks"
+    cleanup $vg_name "$TEST_DEVS"
+    return $test_status
+  fi
+
+  # Make sure volume group $VG got created.
+  if vg_exists "$vg_name"; then
+    test_status=0
+  fi
+
+  cleanup_soft_links "$devlinks"
+  cleanup $vg_name "$TEST_DEVS"
+  return $test_status
+}
+
+# Make sure symlinked disk names are supported.
+test_follow_symlinked_devices
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -53,3 +53,11 @@ cleanup() {
   clean_config_files
   wipe_signatures "$devs"
 }
+
+cleanup_soft_links() {
+  local dev devs=$1
+
+  for dev in $devs; do
+    rm $dev
+  done
+}


rhatdan:
bot, add author to whitelist

rhatdan:
bot, test pull request

rhatdan:
LGTM

rhatdan:
@rhvgoyal PTAL

docker-storage-setup.sh
```diff
@@ -632,6 +631,9 @@ canonicalize_block_devs() {
   for dev in ${devs}; do
     # Looks like we allowed just device name (sda) as valid input. In
     # such cases /dev/$dev should be a valid block device.
+    if [ -h "$dev" ];then
+      dev=$(readlink -f $dev)
```
rhvgoyal: Why not use `readlink -e $dev` ?

rhvgoyal:
Can you please also add a test for making sure symlink to devices work.

ianabc:
I just copied one of the existing tests (004) and modified it slightly to create symlinks to the test devices. I'm doing cleanup inside the test for those symlinks

docker-storage-setup.sh
```diff
@@ -632,7 +632,7 @@ canonicalize_block_devs() {
     # Looks like we allowed just device name (sda) as valid input. In
     # such cases /dev/$dev should be a valid block device.
     if [ -h "$dev" ];then
-      dev=$(readlink -f $dev)
+      dev=$(readlink -e $dev)
```
rhvgoyal: What happens if redlink fails. If destination device does not exist. Will I get a proper error message about this.

rhvgoyal: Also you can modify your existing patch and do a force push. Don't have to create incremental patch for taking care of review feedback.

tests/009-test-follow-symlinked-devices.sh
```diff
@@ -0,0 +1,50 @@
...
+  cleanup $vg_name "$TEST_DEVS"
+  for dev in $devs; do
+    if [ -h $dev ];then
+      rm $dev
+    fi 
+  done
+  return $test_status
+}
+
+# Make sure non-absolute disk names are supported. Ex. sdb.
```
rhvgoyal: This comment is stale and will need change.

docker-storage-setup.sh
ianabc: Im not really sure how to modify the existing patch and force push.

docker-storage-setup.sh
```diff
@@ -627,9 +626,18 @@ check_wipe_block_dev_sig() {
...
   local devs=$1 dev
+  local linktarget
   local devs_abs dev_abs
 
   for dev in ${devs}; do
+    # If the device name is a symlink, follow it and use the target
+    if [ -h "$dev" ];then
+      linktarget=$(readlink -e $dev)
+      if [ $? -ne 0 ] || [ ! -b "$linktarget" ] ; then
+        Fatal "$linktarget doesn't link to a valid block device."
```
rhvgoyal: We already do block device checks on dev_abs so no need to do it again. How about something like.

if ! dest_dev=$(readlink -e $dev);then
    Fatal "readlink failed to resolve device path $dev"
fi

tests/009-test-follow-symlinked-devices.sh
```diff
@@ -0,0 +1,50 @@
...
+VG=$vg_name
+EOF
+  # Run docker-storage-setup
+  $DSSBIN >> $LOGS 2>&1
+
+  # Test failed.
+  if [ $? -ne 0 ]; then
+    cleanup $vg_name "$TEST_DEVS"
+    for dev in $devs; do
+      if [ -h $dev ];then
```
rhvgoyal: We don't require this `-h` test again. We know all $dev are soft links as we have created these just now.

tests/009-test-follow-symlinked-devices.sh
```diff
@@ -0,0 +1,50 @@
...
+  if vg_exists "$vg_name"; then
+    test_status=0
+  fi
+
+  cleanup $vg_name "$TEST_DEVS"
+  for dev in $devs; do
+    if [ -h $dev ];then
+      rm $dev
+    fi 
+  done
```
rhvgoyal: This repetition of cleanup logic feels odd. Can we just define a helper function and call that function at both the places.

cleanup_soft_links()

tests/009-test-follow-symlinked-devices.sh
rhvgoyal: Also it probably is good to cleanup symlinks first and then call `cleanup()`

ianabc:
I added cleanup_soft_links based on other routines in libtest.sh. dev isn't defined as local, but that seems to be the convention for the other subroutines in there so I just followed it.

docker-storage-setup.sh
```diff
@@ -628,8 +628,16 @@ check_wipe_block_dev_sig() {
 canonicalize_block_devs() {
   local devs=$1 dev
   local devs_abs dev_abs
+  local dest_dev
 
   for dev in ${devs}; do
+    # If the device name is a symlink, follow it and use the target
+    if [ -h "$dev" ];then
+      if ! dest_dev=$(readlink -e $dev);then
+        Fatal "$dest_dev failed to resolve linked device path $dev"
```
rhvgoyal: When error happens $dest_dev will be empty string. Right? So it will not help using dest_dev in error message. So how about just keeping it simple as.

`Fatal "Failed to resolve symbolic link $dev"`

rhvgoyal:
I do think `dev` should be local in all routines of libtest.sh. This is not intentional. I think we just could not spot it. If you like you can open another PR to clean it up. Or atleast make dev local in the function you are introducing.

docker-storage-setup.sh
ianabc: Yep, you're right. I've made that change and pushed it now.

rhvgoyal:
I tested it and automated test does not seem to be cleaning up symlink properly. If I pass in two devices, one of the symlinks seems to be left behind.

rhvgoyal:
These tests are assuming that device passed in TEST_DEVS is a regular device. Can we check device. And if it is not symlink, then setup a symlink for test otherwise not. That way, One can pass in symlinks in DEVS= and still be able to run automated tests.

ianabc:
Yes, I can take a look at that. For cleanup I guess we want to cleanup any symlinks that were created by the test, but leave any which existed before (in TEST_DEVS).

rhvgoyal:
@ianabc Looking at test 004, looks like it is assuming input of format "/dev/<foo-disk>". So for now, let us keep it simple. That is for tests, we will continue to pass in devices (and not symlinks). We can have another look at it later in a separate PR.

tests/009-test-follow-symlinked-devices.sh
```diff
@@ -0,0 +1,48 @@
...
+  local devlinks devlink
+  local test_status=1
+  local testname=`basename "$0"`
+  local vg_name="dss-test-foo"
+
+  # Create a symlink for a device and try to follow it
+  for dev in $TEST_DEVS; do
+    if [ ! -h $dev ]; then
+      devlink="$dev-test.$$"
+      ln -s $dev $devlink
```
rhvgoyal: Another minor nit. Can we create these soft links in some dir like /tmp. Right now they seem to be created in cwd of process and if cleanup fails for some reason, it will just leave persistent links behind.

ianabc:
I've reverted the change in 004 and moved the symlinks into /tmp. For the moment I haven't removed the symlink check in 009 but if it is out of place I can get rid of it.

tests/009-test-follow-symlinked-devices.sh
```diff
@@ -0,0 +1,48 @@
...
+    cleanup $vg_name "$TEST_DEVS"
+    return $test_status
+  fi
+
+  # Make sure volume group $VG got created.
+  if vg_exists "$vg_name"; then
+    test_status=0
+  fi
+
+  cleanup_soft_links $devlinks
```
rhvgoyal: double quotes around `$devlinks`

rhvgoyal:
LGTM

rhvgoyal:
@rh-atomic-bot r+

PR #171, opened 2016-11-14
[merged] Let lvm create metadata volume automatically

### LINKED ISSUE #169: error when trying to use all free space for DATA
On Fedora 25 atomic host I'm able to get errors from docker storage setup if I specify to use all free space in the VG. 

```
-bash-4.3# rpm-ostree status
State: idle
Deployments:
● fedora-atomic:fedora-atomic/25/x86_64/docker-host
       Version: 25.40 (2016-11-08 10:20:02)
        Commit: b02be04acde2a04a96d9e969edf3b73753967c06811fc51d65c2b2b4985dff4a
        OSName: fedora-atomic
-bash-4.3# rpm -q docker
docker-1.12.2-5.git8f1975c.fc25.x86_64
-bash-4.3# cat /etc/sysconfig/docker-storage-setup 
GROWPART=true
DATA_SIZE=100%FREE
-bash-4.3# journalctl -u docker-storage-setup.service 
-- Logs begin at Thu 2016-11-10 21:45:09 UTC, end at Thu 2016-11-10 21:52:42 UTC. --
Nov 10 21:45:15 cloudhost.localdomain systemd[1]: Starting Docker Storage Setup...
Nov 10 21:45:15 cloudhost.localdomain docker-storage-setup[966]: CHANGED: partition=2 start=616448 old: size=11966464 end=12582912 new: size=20355072,end=20971520
Nov 10 21:45:15 cloudhost.localdomain docker-storage-setup[966]:   Physical volume "/dev/sda2" changed
Nov 10 21:45:15 cloudhost.localdomain docker-storage-setup[966]:   1 physical volume(s) resized / 0 physical volume(s) not resized
Nov 10 21:45:15 cloudhost.localdomain docker-storage-setup[966]:   WARNING: D-Bus notification failed: The name com.redhat.lvmdbus1 was not provided by any .service files
Nov 10 21:45:15 cloudhost.localdomain docker-storage-setup[966]:   Rounding up size to full physical extent 12.00 MiB
Nov 10 21:45:16 cloudhost.localdomain docker-storage-setup[966]:   Logical volume "docker-poolmeta" created.
Nov 10 21:45:16 cloudhost.localdomain docker-storage-setup[966]:   WARNING: D-Bus notification failed: The name com.redhat.lvmdbus1 was not provided by any .service files
Nov 10 21:45:16 cloudhost.localdomain docker-storage-setup[966]:   Logical volume "docker-pool" created.
Nov 10 21:45:16 cloudhost.localdomain docker-storage-setup[966]:   WARNING: D-Bus notification failed: The name com.redhat.lvmdbus1 was not provided by any .service files
Nov 10 21:45:16 cloudhost.localdomain docker-storage-setup[966]:   WARNING: Converting logical volume atomicos/docker-pool and atomicos/docker-poolmeta to pool's data and metadata volumes.
Nov 10 21:45:16 cloudhost.localdomain docker-storage-setup[966]:   THIS WILL DESTROY CONTENT OF LOGICAL VOLUME (filesystem etc.)
Nov 10 21:45:16 cloudhost.localdomain docker-storage-setup[966]:   Volume group "atomicos" has insufficient free space (0 extents): 3 required.
Nov 10 21:45:16 cloudhost.localdomain systemd[1]: docker-storage-setup.service: Main process exited, code=exited, status=5/NOTINSTALLED
Nov 10 21:45:16 cloudhost.localdomain systemd[1]: Failed to start Docker Storage Setup.
Nov 10 21:45:16 cloudhost.localdomain systemd[1]: docker-storage-setup.service: Unit entered failed state.
Nov 10 21:45:16 cloudhost.localdomain systemd[1]: docker-storage-setup.service: Failed with result 'exit-code'.
```

I end up with a half set up system:

```
-bash-4.3# lsblk
NAME                        MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
loop1                         7:1    0    2G  0 loop 
└─docker-253:0-4431057-pool 253:1    0  100G  0 dm   
sdb                           8:16   0   10G  0 disk 
loop0                         7:0    0  100G  0 loop 
└─docker-253:0-4431057-pool 253:1    0  100G  0 dm   
sdc                           8:32   0  366K  0 disk 
sda                           8:0    0   10G  0 disk 
├─sda2                        8:2    0  9.7G  0 part 
│ └─atomicos-root           253:0    0    3G  0 lvm  /sysroot
└─sda1                        8:1    0  300M  0 part /boot
-bash-4.3# lvs
  LV              VG       Attr       LSize  Pool Origin Data%  Meta%  Move Log Cpy%Sync Convert
  docker-pool     atomicos -wi-------  6.76g                                                    
  docker-poolmeta atomicos -wi------- 12.00m                                                    
  root            atomicos -wi-ao----  2.93g
```

rhvgoyal:
Fixes #169 

Right now we create lvm metadata volume first and then data volume and then
pass it to lvm to convert it into a thin pool (lvconvert). This means lvm
internally creates an spare logical volume (used during repair) and needs
free space in volume group for that spare volume. But lvm does not tell
how big an spare volume will be so caller does not know how much free space
to leave in volume group. And that means if user passes in DATA_SIZE=100%FREE
by the time we call lvconvert, there is no free space left and lvconvert
fails. And we leave the system in uncleaned state (metadata and data volumes
behind).

Dusty came up with idea that why not let lvm create metadata, data and spare
volume. That way docker-storage-setup does not have to deal with how much
free space to leave. Also use option --poolmetadatasize option to specify
size of metadata, so that we retain control on how big metadata volume
should be. 

This should allow users to pass in DATA_SIZE as 100%FREE. Though lvm
developers say it is not a good idea to fill up volume group completely
and leave some space free so that later either metadata, data or spare
volume can grow as need be. 

This patch changes behavior of MIN_DATA_SIZE. Previously we will make sure
free space in VG is more than MIN_DATA_SIZE before creating data volume
(and after crating metadata volume). Now we do the same check before
creating metadata volume. Given metadata volume is very small (.1% of data
size). This can be considered sort of round off error and should not
be a real problem.

Reported-by: Dusty Mabe <dmabe@redhat.com>

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -55,8 +55,6 @@ set -e
 # Note: lvm2 version should be same or higher than lvm2-2.02.112 for lvm
 #       thin pool functionality to work properly.
 POOL_LV_NAME="docker-pool"
-DATA_LV_NAME=$POOL_LV_NAME
-META_LV_NAME="${POOL_LV_NAME}meta"
 
 DOCKER_STORAGE="/etc/sysconfig/docker-storage"
 STORAGE_DRIVERS="devicemapper overlay overlay2"
@@ -209,7 +207,7 @@ get_devicemapper_config_options() {
   # docker expects device mapper device and not lvm device. Do the conversion.
   eval $( lvs --nameprefixes --noheadings -o lv_name,kernel_major,kernel_minor $VG | while read line; do
     eval $line
-    if [ "$LVM2_LV_NAME" = "$DATA_LV_NAME" ]; then
+    if [ "$LVM2_LV_NAME" = "$POOL_LV_NAME" ]; then
       echo POOL_DEVICE_PATH=/dev/mapper/$( cat /sys/dev/block/${LVM2_LV_KERNEL_MAJOR}:${LVM2_LV_KERNEL_MINOR}/dm/name )
     fi
   done )
@@ -244,27 +242,6 @@ $storage_options
 EOF
 
   mv -Z $DOCKER_STORAGE.tmp $DOCKER_STORAGE
-}
-
-create_metadata_lv() {
-  # If metadata lvm already exists (failures from previous run), then
-  # don't create it.
-  # TODO: Modify script to cleanup meta and data lvs if failure happened
-  # later. Don't exit with error leaving partially created lvs behind.
-
-  if lvs -a $VG/${META_LV_NAME} --noheadings &>/dev/null; then
-    Info "Metadata volume $META_LV_NAME already exists. Not creating a new one."
-    return 0
-  fi
-
-  # Reserve 0.1% of the free space in the VG for docker metadata.
-  # Calculating the based on actual data size might be better, but is
-  # more difficult do to the range of possible inputs.
-  VG_SIZE=$( vgs --noheadings --nosuffix --units s -o vg_size $VG )
-  META_SIZE=$(( $VG_SIZE / 1000 + 1 ))
-  if [ ! -n "$META_LV_SIZE" ]; then
-    lvcreate -y -L ${META_SIZE}s -n $META_LV_NAME $VG
-  fi
 }
 
 convert_size_in_bytes() {
@@ -368,9 +345,13 @@ check_min_data_size_condition() {
   fi
 }
 
-create_data_lv() {
+create_lvm_thin_pool () {
+  if [ -z "$DEVS_ABS" ] && [ -z "$VG_EXISTS" ]; then
+    Fatal "Specified volume group $VG does not exist, and no devices were specified"
+  fi
+
   if [ ! -n "$DATA_SIZE" ]; then
-    Fatal "Data volume creation failed. No DATA_SIZE specified"
+    Fatal "DATA_SIZE not specified."
   fi
 
   if ! check_data_size_syntax $DATA_SIZE; then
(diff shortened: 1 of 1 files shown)


rhvgoyal:
@cgwalters @rhatdan @shishir-a412ed PTAL

docker-storage-setup.sh
```diff
@@ -379,27 +360,25 @@ create_data_lv() {
...
-
-create_lvm_thin_pool () {
-  if [ -z "$DEVS_ABS" ] && [ -z "$VG_EXISTS" ]; then
-    Fatal "Specified volume group $VG does not exist, and no devices were specified"
+  # Calculate size of metadata lv. Reserve 0.1% of the free space in the VG
+  # for docker metadata.
+  VG_SIZE=$(vgs --noheadings --nosuffix --units s -o vg_size $VG)
+  META_SIZE=$(( $VG_SIZE / 1000 + 1 ))
+  if [ -z "$META_SIZE" ];then
+    Fatal "Failed to calculate size of metadata volume"
```
dustymabe: we don't need to do any of the above. let LVM do this for us as it will automatically create the metadata LV for us. 

If we want to allow the user to specify META_SIZE in the future, then we can by simply exposing the `--poolmetadatasize` argument to them.

rhvgoyal: I am doing this so that I can give very clear error message to user if we screw up calculating the size of metadata volume. I don't know what error message will lvm emit if META_SIZE="". I can try though.

rhvgoyal: I get following from lvm if META_SIZE="". 

`Invalid argument for --poolmetadatasize: s`

I guess not bad. Will get rid of above check.

rhvgoyal:
Got rid of META_SIZE check. Also started using `--type=thin-pool` in lvconvert command instead of `--thin`. Suggested by lvm developers.

docker-storage-setup.sh
dustymabe: no, what I am saying is that we don't need any of the lines. We don't need to calculate `META_SIZE` at all and we don't need to pass that information in to the `lvcreate` command.

rhvgoyal: @dustymabe I don't want to change existing behavior of the script until and unless we have a good reason to do so. As of now script calculates the metadata size and there is no reason to change it?

rhvgoyal: @dustymabe We can easily change it if our calculations become a problem. We had a long conversation in the past about how big a metadata lvm should be and we did not have a good answer. So we decided on .1%. Given it seems to be working reasonably well, I don't see an incentive to change it.

Even if we have to do it, it should be done in a separate PR and this PR should not be blocked because of that.

dustymabe: > We had a long conversation in the past about how big a metadata lvm should be and we did not have a good answer.

Even more of a reason to let the tool decide.

> Even if we have to do it, it should be done in a separate PR and this PR should not be blocked because of that.

I can accept that. if we leave it like this then we should probably add the 'if statement' back

rhvgoyal: I am not sure why do you think that lvm will necessarily do a better job of figuring out metadata volume size better. So I don't agree with `Even more of a reason to let the tool decide`. You are assuming that lvm always makes good decisions on this.

Ok, I can add back META_SIZE being empty check.

rhvgoyal:
@dustymabe Added META_SIZE check back. PTAL.

rhvgoyal:
@rhatdan @cgwalters PTAL

cgwalters:
@rh-atomic-bot r+ c93c2e3

cgwalters:
This is obviously a lot nicer!

PR #173, opened 2016-11-17
We want to execute docker-storage-setup on the host system

rhatdan:
If docker-storage-setup is run from a Super Privileged Container (SPC)
we want to execute the script on the host machine.

$HOST indicates the mountpoint of the host root fs of the container.
This patch will chroot to this rootfs and execute the script.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -54,12 +54,20 @@ set -e
 #
 # Note: lvm2 version should be same or higher than lvm2-2.02.112 for lvm
 #       thin pool functionality to work properly.
+
+# Check if the script is being run inside of a SPC container by examining $HOST
+# If set, PATH to use $HOST/usr/bin and $HOST/usr/sbin for host helper applications.
+if [[ ! -e $HOST ]]; then
+    export PATH=$HOST/usr/bin:$HOST/usr/sbin:$PATH
+fi
+
 POOL_LV_NAME="docker-pool"
 
-DOCKER_STORAGE="/etc/sysconfig/docker-storage"
+DOCKER_STORAGE="${HOST}/etc/sysconfig/docker-storage"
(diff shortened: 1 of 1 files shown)


rhatdan:
@rhvgoyal PTAL

rhvgoyal:
This assumes that all the /usr/lib/docker-storage-setup files are installed on host. That means docker-storage-setup and all dependencies are installed on host. But AFAIK, we are not doing any of that?

rhatdan:
Yes you are right.  I thought the script was all encompassing.

rhatdan:
Another version, this one just adds the $HOST/usr/bin and $HOST/usr/sbin to $PATH.
Also changes a couple of other small parts of script.  Probably need to look at other helper apps.

docker-storage-setup.sh
```diff
@@ -54,12 +54,20 @@ set -e
 #
 # Note: lvm2 version should be same or higher than lvm2-2.02.112 for lvm
 #       thin pool functionality to work properly.
+
+# Check if the script is being run inside of a SPC container by examining $HOST
+# If set, PATH to use $HOST/usr/bin and $HOST/usr/sbin for host helper applications.
```
cgwalters: Yeah, but we'll end up loading shared libraries from the host.  And any program that uses absolute paths for helper binaries (not uncommon) will end up using them from the container, not the host...leading to potential confusion.  This seems very possible with lvm, for example in a quick `git grep /sbin` in the LVM git I see:

```
configure.in:                   THIN_CHECK_CMD=/usr/sbin/thin_check
```

The only reliable and sane way to run host binaries to to execute them in the host, i.e. we use the "generate rpm" path for this.

rhatdan: I think for the time being we are just going to not run commands from the host.  The container needs to match the host as much as possible.  We need to see if this is a real problem.

rhatdan:
We have decided not to do this for now, closing.

PR #175, opened 2016-11-30
[merged] Options {DOCKER_ROOT_VOLUME, DOCKER_ROOT_VOLUME_SIZE} for setting up separate logical volume for docker root.

shishir-a412ed:

### DIFF
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -98,6 +98,37 @@ WIPE_SIGNATURES:
       overwritten as suitable. This also means that if there is any
       data on disk, it will be lost.
 
+DOCKER_ROOT_VOLUME:
+      By default no new volume and filesystem will be setup
+      for docker root dir. Docker creates /var/lib/docker/ on
+      top of underlying filesystem for storing images and containers.
+
+      To carve out a separate logical volume for storing docker
+      images/containers/volumes data, set DOCKER_ROOT_VOLUME=yes
+      NOTE: devicemapper stores images/containers on thin pool.
+
+      docker-storage-setup creates a logical volume with an XFS
+      filesystem mounted on docker root directory
+      (default: /var/lib/docker).
+
+DOCKER_ROOT_VOLUME_SIZE:
+      Specify the desired size for docker root volume. It defaults
+      to 40% of all free space.
+
+      DOCKER_ROOT_VOLUME_SIZE can take values acceptable to
+      "lvcreate -L" as well as some values acceptable to to 
+      "lvcreate -l". If user intends to pass values acceptable 
+      to "lvcreate -l", then only those values which contains "%" 
+      in syntax are acceptable.  If value does not contain "%" it 
+      is assumed value is suitable for "lvcreate -L".
+
+      NOTE: If both STORAGE_DRIVER=devicemapper and 
+      DOCKER_ROOT_VOLUME=yes is set, docker-storage-setup would set
+      up the thin pool for devicemapper first, followed by docker 
+      root volume. e.g if free space in the volume group is 10G, 
+      devicemapper thin pool size would be 4G (40% of 10G) and docker
+      root volume would be 2.4G (40% of 6G).
+
 The options below should be specified as values acceptable to 'lvextend -L':
 
 ROOT_SIZE: The size to which the root filesystem should be grown.
--- a/docker-storage-setup.conf
+++ b/docker-storage-setup.conf
@@ -1,7 +1,6 @@
 # Specify storage driver one wants to use with docker. Default is devicemapper.
 # Other possible options are overlay, overlay2 and "". Empty string means do
 # not do any storage setup.
-#
 STORAGE_DRIVER=devicemapper
 
 # A set of extra options that should be appended to the generated
@@ -84,3 +83,31 @@ DEVICE_WAIT_TIMEOUT=60
 # by setting WIPE_SIGNATURES=true. Be careful before using this option
 # as this means if there was any leftover data on disk, it will be lost.
 WIPE_SIGNATURES=false
+
+# By default no new volume and filesystem will be setup for docker
+# root dir. Docker creates /var/lib/docker/ on top of underlying filesystem
+# for storing images and containers.
+#
+# To carve out a separate logical volume for storing docker
+# images/containers/volumes data, set DOCKER_ROOT_VOLUME=yes
+# NOTE: devicemapper stores images/containers on thin pool.
+#
+# docker-storage-setup creates a logical volume with an XFS filesystem mounted
+# on docker root directory (default: /var/lib/docker).
+DOCKER_ROOT_VOLUME=no
+
+# Specify the desired size for docker root volume. It defaults to 40% of all
+# free space.
+#
+# DOCKER_ROOT_VOLUME_SIZE can take values acceptable to "lvcreate -L" as well
+# as some values acceptable to to "lvcreate -l". If user intends to pass
+# values acceptable to "lvcreate -l", then only those values which
+# contains "%" in syntax are acceptable.  If value does not contain "%" it
+# is assumed value is suitable for "lvcreate -L".
+#
+# NOTE: If both STORAGE_DRIVER=devicemapper and DOCKER_ROOT_VOLUME=yes is
+# set, docker-storage-setup would set up the thin pool for devicemapper
+# first, followed by docker root volume. e.g if free space in the volume 
+# group is 10G, devicemapper thin pool size would be 4G (40% of 10G)
+# and docker root volume would be 2.4G (40% of 6G).
+DOCKER_ROOT_VOLUME_SIZE=40%FREE
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -55,6 +55,8 @@ set -e
 # Note: lvm2 version should be same or higher than lvm2-2.02.112 for lvm
 #       thin pool functionality to work properly.
 POOL_LV_NAME="docker-pool"
+DOCKER_ROOT_LV_NAME="docker-root-lv"
+DOCKER_ROOT_DIR="/var/lib/docker"
 
 DOCKER_STORAGE="/etc/sysconfig/docker-storage"
 STORAGE_DRIVERS="devicemapper overlay overlay2"
@@ -409,6 +411,43 @@ check_docker_storage_metadata() {
   Info "Docker state can be reset by stopping docker and by removing ${DOCKER_METADATA_DIR} directory. This will destroy existing docker images and containers and all the docker metadata."
   exit 1
 }
+
+systemd_escaped_filename () {
+  local escaped_path filename path=$1
+  escaped_path=$(echo ${path}|sed 's|-|\\x2d|g')
+  filename=$(echo ${escaped_path}.mount|sed 's|/|-|g' | cut -b 2-)
+  echo $filename
+}
+
+reset_docker_root_volume () {
+  local mp filename
+  if docker_root_volume_exists; then
+    mp=$(docker_root_lv_mountpoint)
+    if [ -n "$mp" ];then
+      if ! umount $mp >/dev/null 2>&1; then
+        Fatal "Failed to unmount $mp"
+      fi
+    fi
+    lvchange -an $VG/${DOCKER_ROOT_LV_NAME}
+    lvremove $VG/${DOCKER_ROOT_LV_NAME}
+  else
+    return 0
+  fi
+  # If the user has manually unmounted ${DOCKER_ROOT_DIR}, mountpoint (mp)
+  # will be empty. Extract ${mp} from /etc/sysconfig/docker in that case.
+  if [ -z "$mp" ];then
+    if ! get_docker_root_dir; then
+      return 1
+    fi
+    mp=${DOCKER_ROOT_DIR}
+  fi
+  filename=$(systemd_escaped_filename $mp)
+  if [ -f /etc/systemd/system/$filename ];then
+    systemctl disable $filename >/dev/null 2>&1
+    systemctl daemon-reload
+    rm /etc/systemd/system/$filename >/dev/null 2>&1
+  fi
+} 
 
 reset_lvm_thin_pool () {
   if lvm_pool_exists; then
@@ -839,6 +878,102 @@ get_existing_storage_driver() {
   return 1
 }
 
+docker_root_volume_exists() {
+  lvs $VG/$DOCKER_ROOT_LV_NAME > /dev/null 2>&1 && return 0
+  return 1
+}
+
+# This returns the mountpoint of $DOCKER_ROOT_LV_NAME
+docker_root_lv_mountpoint() {
+  local mounts
+  mounts=$(findmnt -n -o TARGET --first-only --source /dev/$VG/$DOCKER_ROOT_LV_NAME)
+  echo $mounts
+}
+
+# Mount docker root volume on $DOCKER_ROOT_DIR.
+# drop a systemd mount unit configuration file at /etc/systemd/system.
+# This would allow the $DOCKER_ROOT_DIR mountpoint to persist across reboots.
+mount_docker_root_volume() {
+  local filename
+  # filename must match $DOCKER_ROOT_DIR path. e.g if $DOCKER_ROOT_DIR is /var/lib/docker
+  # then filename will be var-lib-docker.mount
+  filename=$(systemd_escaped_filename ${DOCKER_ROOT_DIR})
+  cat <<EOF > /etc/systemd/system/$filename
+[Unit]
+Description=Mount docker-root-lv on docker root directory.
+Before=docker-storage-sethup.service
+
+[Mount]
+What=/dev/$VG/$DOCKER_ROOT_LV_NAME
+Where=${DOCKER_ROOT_DIR}
+Type=xfs
+Options=defaults
+
+[Install]
+WantedBy=docker-storage-setup.service
+EOF
+  systemctl enable $filename >/dev/null 2>&1
+  systemctl start $filename 
+}
+
+# Create a logical volume of size specified by first argument. Name of the
+# volume is specified using second argument.
+create_lv() {
+  local data_size=$1
+  local data_lv_name=$2
+
+  # TODO: Error handling when data_size > available space.
+  if [[ $data_size == *%* ]]; then
+    lvcreate -y -l $data_size -n $data_lv_name $VG || return 1
+  else
+    lvcreate -y -L $data_size -n $data_lv_name $VG || return 1
+  fi
+return 0
+}
+
+setup_docker_root_volume() {
+  if [ -z "$DOCKER_ROOT_VOLUME_SIZE" ]; then
+    Fatal "Specify a valid value for DOCKER_ROOT_VOLUME_SIZE."
+  fi
+
+  if ! check_data_size_syntax $DOCKER_ROOT_VOLUME_SIZE; then
+    Fatal "DOCKER_ROOT_VOLUME_SIZE value $DOCKER_ROOT_VOLUME_SIZE is invalid."
+  fi
+
+  if ! create_lv $DOCKER_ROOT_VOLUME_SIZE $DOCKER_ROOT_LV_NAME; then
+    Fatal "Failed to create volume $DOCKER_ROOT_VOLUME_SIZE of size ${DOCKER_ROOT_VOLUME_SIZE}."
+  fi
+
+  if ! mkfs -t xfs /dev/$VG/$DOCKER_ROOT_LV_NAME > /dev/null; then
+    Fatal "Failed to create filesystem on /dev/$VG/${DOCKER_ROOT_LV_NAME}."
+  fi
+
+  if ! mount_docker_root_volume; then
+    Fatal "Failed to mount docker root volume ${DOCKER_ROOT_LV_NAME} on ${DOCKER_ROOT_DIR}"
+  fi
+
+  # setup right selinux label first time fs is created. Mount operation
+  # changes the label of directory to reflect the label on root inode
+  # of mounted fs.
+  if ! restore_selinux_context $DOCKER_ROOT_DIR; then
+    return 1
+  fi
+}
+
+setup_docker_root_lv_fs() {
+  [ "$DOCKER_ROOT_VOLUME" != "yes" ] && return 0
+  if ! setup_docker_root_dir; then
+    return 1
+  fi
+  if docker_root_volume_exists; then
+    return 0
+  fi
+  # Docker root volume does not exist. Create one.
+  if ! setup_docker_root_volume; then
+    Fatal "Failed to setup logical volume for docker root."
+  fi
+}
+
 setup_storage() {
   local current_driver
 
@@ -865,15 +1000,77 @@ setup_storage() {
    Fatal "Storage is already configured with ${current_driver} driver. Can't configure it with ${STORAGE_DRIVER} driver. To override, remove $DOCKER_STORAGE and retry."
   fi
 
-  # Set up lvm thin pool LV
+  # If a user decides to setup both (a) and (b):
+  # a) lvm thin pool for devicemapper.
+  # b) a separate volume for docker root.
+  # (a) will have a higher priority than (b) and will be setup first.
+
+  # Set up lvm thin pool LV.
   if [ "$STORAGE_DRIVER" == "devicemapper" ]; then
     setup_lvm_thin_pool
   else
       write_storage_config_file
   fi
+
+  # If docker root is on a separate volume, setup that.
+  if ! setup_docker_root_lv_fs; then
+    Error "Failed to setup docker root volume."
+    return 1
+  fi
+}
+
+restore_selinux_context() {
+  local dir=$1
+
+  if ! restorecon -R $dir; then
+    Error "restorecon -R $dir failed."
+    return 1
+  fi
+}
+
+get_docker_root_dir(){
+    local flag=false path
+    options=$(grep -e "^OPTIONS" /etc/sysconfig/docker|cut -d"'" -f 2)
+    for opt in $options
+    do
+        if [ "$flag" = true ];then
+           path=$opt
+           flag=false
+           continue
+        fi
+	case "$opt" in
+            "-g"|"--graph")
+                flag=true
+                ;;
+            -g=*|--graph=*)
+		path=$(echo $opt|cut -d"=" -f 2)
+                ;;
+            *)
+                ;;
+        esac
+    done
+    if ! DOCKER_ROOT_DIR=$(realpath -m $path);then
+      Fatal "realpath failed on $path"
+    fi
+}
+
+setup_docker_root_dir() {
+  if ! get_docker_root_dir; then
+    return 1
+  fi
+
+  [ -d "$DOCKER_ROOT_DIR" ] && return 0
+
+  # Directory does not exist. Create one.
+  mkdir -p $DOCKER_ROOT_DIR
+  return $?
 }
 
 reset_storage() {
+  
+    if [ "$DOCKER_ROOT_VOLUME" == "yes" ];then
+        reset_docker_root_volume
+    fi      	
     if [ "$STORAGE_DRIVER" == "devicemapper" ]; then
 	reset_lvm_thin_pool
     fi


shishir-a412ed:
ping @rhvgoyal

docker-storage-setup.sh
```diff
@@ -873,6 +987,82 @@ setup_storage() {
...
+    return 1
+  fi
+}
+
+# Inherit the same selinux context as /var/lib/docker.
+setup_selinux_context_root_dir() {
+  local dir=$1
+  local policy
+
+  [ "$dir" == "/var/lib/docker" ] && return 0
```
rhatdan: Shouldn't this be ${DOCKER_ROOT_DIR}

docker-storage-setup.sh
```diff
@@ -873,6 +987,82 @@ setup_storage() {
...
+
+  # If there is already a labeling policy for the dir, don't try to add one.
+  policy=`semanage fcontext -l -C $dir | grep "^$dir =" | wc -l`
```
rhatdan: Shouldn't this be?

```
policy=`semanage fcontext -l -C | grep "^$dir =" | wc -l`
```

docker-storage-setup.sh
```diff
@@ -873,6 +987,82 @@ setup_storage() {
...
+    if ! semanage fcontext -a -e /var/lib/docker $dir; then
+      echo "semanage fcontext -a -e /var/lib/docker $dir failed."
+      return 1
+    fi
+  fi
+}
+
+get_docker_root_dir(){
+    local flag=false
+    options=$(grep -e "^OPTIONS" /etc/sysconfig/docker|cut -d"'" -f 2)
```
rhatdan: How about something like this

 sh -c ". /etc/sysconfig/docker; echo \$OPTIONS"
--selinux-enabled --log-driver=journald --add-registry atomic-registry.usersys.redhat.com

docker-storage-setup.sh
```diff
@@ -873,6 +987,82 @@ setup_storage() {
...
+    for opt in $options
```
rhatdan: I think we can match the regular expression using something like awk or grep rather then building our own version.

rhatdan: Do we handle the case where we have docker-latest installed?

rhvgoyal:
Can we provide better changelog for the patch and explain what problem these new options are solving.

docker-storage-setup.1
```diff
@@ -63,6 +63,21 @@ GROWPART:
       disabled. Use GROWPART=true to enable automatic partition
       table resizing.
 
+DOCKER_ROOT_VOLUME:
+      If set to "yes", then a separate logical volume will be setup
+      for docker root and it will be mounted at docker root directory.
+      By default, docker root is located at "/var/lib/docker", unless a
+      user start docker daemon using "-g" or "--graph" option and provides
+      a custom docker root location.  
+
```
rhvgoyal: Probably it is a good idea to mention that we are going to scan /etc/sysconfig/docker
file to figure out -g option. If found it will be used as docker root dir otherwise /var/lib/docker is default.

docker-storage-setup.conf
```diff
@@ -1,7 +1,28 @@
...
+# values # acceptable to "lvcreate -l", then only those values which
+# contains "%" in syntax are acceptable.  If value does not contain "%" it
+# is assumed # value is suitable for "lvcreate -L".
+DOCKER_ROOT_VOLUME_SIZE=40%FREE
+
 # Specify storage driver one wants to use with docker. Default is devicemapper.
 # Other possible options are overlay, overlay2 and "". Empty string means do
 # not do any storage setup.
-#
+
```
rhvgoyal: No need for this extra line.

docker-storage-setup.conf
```diff
@@ -1,7 +1,28 @@
...
+# The desired size for the docker root volume. It defaults using 40% of all
+# free space.
+#
+# DOCKER_ROOT_VOLUME_SIZE can take values acceptable to "lvcreate -L" as well
+# as some values acceptable to to "lvcreate -l". If user intends to pass
+# values # acceptable to "lvcreate -l", then only those values which
+# contains "%" in syntax are acceptable.  If value does not contain "%" it
+# is assumed # value is suitable for "lvcreate -L".
+DOCKER_ROOT_VOLUME_SIZE=40%FREE
+
```
rhvgoyal: These are new options. Instead of adding them in the beginning of file, how about adding them towards the end of the file?

docker-storage-setup.conf
```diff
@@ -1,7 +1,28 @@
+# By default no new volume and fs will be setup for docker root dir. Docker
+# will create /var/lib/docker/ and use underlying fs for storing image and
+# containers.
+# 
+# If one wants to carve out a separate logical volume for storing docker
+# metadata and images/containers/volumes, then set DOCKER_ROOT_VOLUME=yes
+#
+# This will try to create a logical volume, create filesystem on that volume
+# and mount on $DOCKER_ROOT_DIR.
```
rhvgoyal: I think very soon we will have request for allowing passing in filesystem to create and options to pass in. May be we can create two more options at that time. Say,

DOCKER_ROOT_FS="xfs"
DOCKER_ROOT_FS_OPTS="-o discard"

This is just an example for future extension. For now, I think we can just mention that by default xfs filesystem is created.

docker-storage-setup.conf
```diff
@@ -1,7 +1,28 @@
...
+DOCKER_ROOT_VOLUME=yes
+
```
rhvgoyal: By default we don't want to create docker root volume. So this should be.

DOCKER_ROOT_VOLUME=no

docker-storage-setup.conf
```diff
@@ -1,7 +1,28 @@
...
+# The desired size for the docker root volume. It defaults using 40% of all
+# free space.
+#
+# DOCKER_ROOT_VOLUME_SIZE can take values acceptable to "lvcreate -L" as well
+# as some values acceptable to to "lvcreate -l". If user intends to pass
+# values # acceptable to "lvcreate -l", then only those values which
```
rhvgoyal: Extra `#` in the middle of line.

docker-storage-setup.conf
```diff
@@ -1,7 +1,28 @@
...
+# contains "%" in syntax are acceptable.  If value does not contain "%" it
+# is assumed # value is suitable for "lvcreate -L".
```
rhvgoyal: Extra `#` in the middle of line.

docker-storage-setup.sh
shishir-a412ed: Fixed based on nalin's comments.

shishir-a412ed: docker-latest will be handled in the spec file, not part of this PR.

docker-storage-setup.sh
shishir-a412ed: @rhvgoyal is okay with `grep`. Leaving as it is for now.

shishir-a412ed:
@rhvgoyal @rhatdan PTAL.

rhvgoyal:
shishir, lets keep changelog line lenth to around 60-70. Right now, do git log on your terminal and chanelog lines are much longer and these are wrapping and look pretty odd.

docker-storage-setup.conf
```diff
@@ -84,3 +83,24 @@ DEVICE_WAIT_TIMEOUT=60
 # by setting WIPE_SIGNATURES=true. Be careful before using this option
 # as this means if there was any leftover data on disk, it will be lost.
 WIPE_SIGNATURES=false
+
+# By default no new volume and filesystem (fs) will be setup for docker root dir.
+# Docker will create /var/lib/docker/ and use underlying fs for storing image and
```
rhvgoyal: line more than 80 char.

docker-storage-setup.conf
```diff
@@ -84,3 +83,24 @@ DEVICE_WAIT_TIMEOUT=60
 # by setting WIPE_SIGNATURES=true. Be careful before using this option
 # as this means if there was any leftover data on disk, it will be lost.
 WIPE_SIGNATURES=false
+
+# By default no new volume and filesystem (fs) will be setup for docker root dir.
```
rhvgoyal: line more than 80 char

docker-storage-setup.sh
```diff
@@ -865,6 +969,16 @@ setup_storage() {
    Fatal "Storage is already configured with ${current_driver} driver. Can't configure it with ${STORAGE_DRIVER} driver. To override, remove $DOCKER_STORAGE and retry."
   fi
 
+  if [ "$STORAGE_DRIVER" == "devicemapper" ] && [ "$DOCKER_ROOT_VOLUME" == "yes" ];then
+   Fatal "STORAGE_DRIVER=${STORAGE_DRIVER} and DOCKER_ROOT_VOLUME=${DOCKER_ROOT_VOLUME} are mutually exclusive options. Use either one of them, not both."
+  fi
```
rhvgoyal: shishir, I am changing my mind on this now. There is no real technical reason that why root volume can't be allowed with devicemapper storage driver. So lets get rid of this restriction.

docker-storage-setup.sh
```diff
@@ -865,6 +969,16 @@ setup_storage() {
...
+
+  # If docker root is on a separate volume, setup that first.
+  if ! setup_docker_root_lv_fs; then
+    echo "Failed to setup docker root volume."
+    return 1
+  fi
+
```
rhvgoyal: I guess we will have to define in documentation very clearly order of creation of root lv and devicemapper lv. Reason being that they both can pass syntax like 50%FREE. And order of operation will decide which gets what size.

To make sure we don't get surprises, how about creating devicemapper volume first and then root volume.

rhvgoyal: Also lets document this in man page and also in docker-storage-setup.conf

shishir-a412ed:
@rhvgoyal PTAL.

rhvgoyal:
When i do `man ./docker-storage-setup.1`, description of DOCKER_ROOT_VOLUME and DOCKER_ROOT_VOLUME_SIZE is not properly formatted. Some of the lines are too long.

rhvgoyal:
In man page, can we add description of new options after `WIPE_SIGNATURES`. This will be consistent with docker-storage-setup.conf

docker-storage-setup.sh
rhvgoyal: Can we define path to docker file in a global variable instead of hardcoding `/etc/sysconfig/docker` here.

rhvgoyal: I am wondering if we can use /usr/bin/getopt utility to parse -g/--graph-driver parameter? An example is provided in `/usr/share/doc/util-linux/getopt-parse.bash`

rhvgoyal:
We will have to change "reset_storage" as well I think to remove docker root volume. Also we probably will have to unmount volume first and then remove it.

rhvgoyal:
Can we add some automated test cases to test this feature.

docker-storage-setup.sh
```diff
@@ -865,14 +969,102 @@ setup_storage() {
...
+    options=$(grep -e "^OPTIONS" /etc/sysconfig/docker|cut -d"'" -f 2)
+    for opt in $options
+    do
+        if [ "$flag" = true ];then
+           DOCKER_ROOT_DIR=$opt
+           flag=false
+           continue
+        fi
+        if [[ "$opt" == --graph* || "$opt" == -g* ]];then
+           if [[ "$opt" == *"="* ]];then
```
rhvgoyal: This also accepts `--graph-driver foo` as valid option.

docker-storage-setup.sh
```diff
@@ -865,14 +969,102 @@ setup_storage() {
...
+           else
+             flag=true
+             continue
+           fi
+        fi
+    done
+}
+
+check_docker_root_volume_options() {
+
```
rhvgoyal: This function name does not seem to match what we are doing in that function. Basically we are parsing and setting up docker root dir. May be call this function `setup_docker_root_dir`

docker-storage-setup.sh
```diff
@@ -839,6 +841,108 @@ get_existing_storage_driver() {
...
+    fi
+
+    if ! mount_docker_root_volume; then
+      return 1
+    fi
+
+    # setup right selinux label first time fs is created. Mount operation
+    # changes the label of directory to reflect the label on root inode
+    # of mounted fs.
+    if ! restore_selinux_context $DOCKER_ROOT_DIR; then
```
rhatdan: You could do a mkdir -Z when creating the directory, But this will not create it with the correct label if the directory is /host/var/lib/docker.  I just submitted some changes to container-selinux that would make the creation of this directory work for certain process types.  We probably can get it to work correctly for docker-storage-setup, if we know which domain it is going to run as.  If it runs as docker_t these directories will get created with the correct label,

docker-storage-setup.sh
shishir-a412ed: Which `docker version` do you see this flag (`--graph-driver`). I am not able to find it.

docker-storage-setup.sh
rhvgoyal: @rhatdan, so you control the label of docker-storage-setup in selinux policy right? So you can control that to whatever you want and make sure directory is created with right label?

docker-storage-setup.sh
rhvgoyal: This is not a valid option. I just set invalid option --graph-driver in /etc/sysconfig/docker and this logic still took it.

shishir-a412ed:
@rhvgoyal @rhatdan PTAL.

Shishir

docker-storage-setup.conf
```diff
@@ -84,3 +83,28 @@ DEVICE_WAIT_TIMEOUT=60
...
 WIPE_SIGNATURES=false
+
+# By default no new volume and filesystem (fs) will be setup for docker
+# root dir. Docker will create /var/lib/docker/ and use underlying fs for
+# storing image and containers.
+#
+# If one wants to carve out a separate logical volume for storing docker
+# metadata and images/containers/volumes, then set DOCKER_ROOT_VOLUME=yes
+#
+# This will try to create a logical volume, create a XFS filesystem on
```
rhatdan: docker-storage-setup will try to create ...

docker-storage-setup.conf
```diff
@@ -84,3 +83,28 @@ DEVICE_WAIT_TIMEOUT=60
...
+# that volume and mount on $DOCKER_ROOT_DIR.
+#
+# NOTE: If both STORAGE_DRIVER=devicemapper and DOCKER_ROOT_VOLUME=yes is
+# passed, docker-storage-setup would set up the thin pool for devicemapper
```
rhatdan: s/passed/set

docker-storage-setup.conf
```diff
@@ -84,3 +83,28 @@ DEVICE_WAIT_TIMEOUT=60
...
+# first, followed by docker root volume.
+DOCKER_ROOT_VOLUME=no
+
+# The desired size for the docker root volume. It defaults using 40% of all
```
rhatdan: Shouldn't this start out with a 
DOCKER_ROOT_VOLUME_SIZE

Is this the Percentage of free space available 
The default size for DOCKER_ROOT_VOLUME_SIZE

rhatdan: I think this whole description section we should work on together, if this is the section an admin would be reading.

rhatdan: We can probably merge the this patch and then sit down in a room and go over the description line by line,

docker-storage-setup.sh
```diff
@@ -839,6 +853,141 @@ get_existing_storage_driver() {
...
+  [ -z "$mounts" ] && return 1
+  # If root volume is mounted, it should be mounted on $DOCKER_ROOT_DIR
+  # otherwise it is an error.
+  target=`echo $mounts | cut -d " " -f1`
+  [ "$target" == "$DOCKER_ROOT_DIR" ] && return 0
+    echo "Docker root volume /dev/$VG/$DOCKER_ROOT_LV_NAME is mounted on $target instead of ${DOCKER_ROOT_DIR}. Unmount it and try again."
+    exit 1
+}
+
+# drop a systemd mount file at /etc/systemd/system.
```
rhatdan: Mount unit configuration file

docker-storage-setup.sh
```diff
@@ -839,6 +853,141 @@ get_existing_storage_driver() {
...
+# This would allow the $DOCKER_ROOT_DIR mountpoint to persist across reboots.
+drop_systemd_mount_file() {
+  IFS='/ '
+  # filename must match $DOCKER_ROOT_DIR path. e.g if $DOCKER_ROOT_DIR is /var/lib/docker
+  # then filename will be var-lib-docker.mount
+  read -r -a elements <<< $DOCKER_ROOT_DIR
```
rhatdan: filename=echo ${DOCKER_ROOT_DIR}.mount /var/lib/docker | sed 's|/|-|g' | cut -b 2-

docker-storage-setup.sh
```diff
@@ -839,6 +853,141 @@ get_existing_storage_driver() {
...
+  [ "$target" == "$DOCKER_ROOT_DIR" ] && return 0
+    echo "Docker root volume /dev/$VG/$DOCKER_ROOT_LV_NAME is mounted on $target instead of ${DOCKER_ROOT_DIR}. Unmount it and try again."
+    exit 1
+}
+
+# drop a systemd mount file at /etc/systemd/system.
+# This would allow the $DOCKER_ROOT_DIR mountpoint to persist across reboots.
+drop_systemd_mount_file() {
+  IFS='/ '
+  # filename must match $DOCKER_ROOT_DIR path. e.g if $DOCKER_ROOT_DIR is /var/lib/docker
```
rhvgoyal: If you are setting IFS, then you should save previous IFS and restore it at the end?

shishir-a412ed:
@rhatdan @rhvgoyal PTAL.

Shishir

rhatdan:
@shishir-a412ed Lets meet with @rhvgoyal and go over the docs before we merge.  Meet after standup?

shishir-a412ed:
@rhvgoyal @rhatdan PTAL.

One last thing I wanted to check: Should we cleanup the mount file on `/usr/bin/docker-storage-setup --reset`? i.e in `reset_docker_root_volume` function, should we do:
```
systemctl disable $filename`
systemctl daemon-reload
rm /etc/systemd/system/$filename
```
Shishir

rhvgoyal:
@shishir-a412ed I think cleaning up systemd .mount file on reset makes sense. Anyway we will delete the root logical volume and mounting will fail. So yes, lets clean it up.

rhatdan:
Make sure you eat errors of $filename is not enabled.

docker-storage-setup.1
```diff
@@ -93,11 +93,39 @@ DEVICE_WAIT_TIMEOUT:
 WIPE_SIGNATURES:
       Wipe any signatures found on disk. Valid values are
       true/false and default value is false. By default if any
-      signatures are found on disk operation is aborted. If this value
-      is set to true, then signatures will either be wiped or
+      signatures are found on disk operation is aborted. If this
+      value is set to true, then signatures will either be wiped or
```
rhvgoyal: Why are we changing this? This is not related to this patch.

docker-storage-setup.1
```diff
@@ -93,11 +93,39 @@ DEVICE_WAIT_TIMEOUT:
...
+      By default no new volume and filesystem will be setup for docker
+      root dir. Docker creates /var/lib/docker/ on top of underlying
+      filesystem for storing images and containers.
+
+      To carve out a separate logical volume for storing docker
+      images/containers/volumes data, set DOCKER_ROOT_VOLUME=yes
+      NOTE: devicemapper stores images/containers on thin pool.
+
+      docker-storage-setup creates a logical volume with an XFS
+      filesystem mounted on docker root directory (default: /var/lib/docker).
```
rhvgoyal: Do man `./docker-storage-setup.1` and make sure all the lines look fine. This line seems to wrap.

docker-storage-setup.1
```diff
@@ -93,11 +93,39 @@ DEVICE_WAIT_TIMEOUT:
...
+
+DOCKER_ROOT_VOLUME_SIZE:
+      Specify the desired size for docker root volume. It defaults to
+      40% of all free space.
+
+      DOCKER_ROOT_VOLUME_SIZE can take values acceptable to "lvcreate -L"
```
rhvgoyal: Many of the lines in this section are wrapping.

docker-storage-setup.sh
```diff
@@ -410,6 +412,18 @@ check_docker_storage_metadata() {
   exit 1
 }
 
+reset_docker_root_volume () {
+  if docker_root_volume_exists; then
+     if ! get_docker_root_dir; then
+        return 1
+     fi
```
rhvgoyal: We should not have to look at docker root dir again. We can just check if root volume is mounted and where it is mounted. Just unmount it and delete volume.

docker-storage-setup.sh
```diff
@@ -839,6 +853,123 @@ get_existing_storage_driver() {
...
+    lvs $VG/$DOCKER_ROOT_LV_NAME > /dev/null 2>&1 && return 0
+    return 1
+}
+
+# Check if docker root volume is already mounted.
+# This return 0 if volume is already mounted on $DOCKER_ROOT_DIR. Returns 1
+# if volumes needs mounting. On all other error conditions it exits.
+
+docker_root_lv_mounted() {
+  local mounts target
```
rhvgoyal: I think we can modify this function to return location of mount point of docker volume. It will return 1 in case of error. Return 0 in case of success or if volume is not mounted. If volume is not mounted, will return 0 and echo empty string.

rhvgoyal: That way it can easily be used in more than one context.

docker-storage-setup.sh
```diff
@@ -839,6 +853,123 @@ get_existing_storage_driver() {
...
+  if ! mounts=$(findmnt -n --first-only --source /dev/$VG/$DOCKER_ROOT_LV_NAME); then
+    return 1
+  fi
+  [ -z "$mounts" ] && return 1
+  # If root volume is mounted, it should be mounted on $DOCKER_ROOT_DIR
+  # otherwise it is an error.
+  target=`echo $mounts | cut -d " " -f1`
+  [ "$target" == "$DOCKER_ROOT_DIR" ] && return 0
+    Error "Docker root volume /dev/$VG/$DOCKER_ROOT_LV_NAME is mounted on $target instead of ${DOCKER_ROOT_DIR}. Unmount it and try again."
+    exit 1
```
rhvgoyal: This logic of giving error if volume is not mounted on correct directory can go outside the funciton at the call site.

rhvgoyal: If you want to exit immediately, you can use `Fatal` instead of `Error`

docker-storage-setup.sh
```diff
@@ -839,6 +853,123 @@ get_existing_storage_driver() {
...
+  Description=Mount docker-root-lv on docker root directory.
+
+  [Mount]
+  What=/dev/$VG/$DOCKER_ROOT_LV_NAME
+  Where=${DOCKER_ROOT_DIR}
+  Type=xfs
+  Options=defaults
+
+  [Install]
+  WantedBy=multi-user.target
```
rhvgoyal: I did `systemd list-dependencies --all docker.service` and it does not list `multi-user.target`. That means there is no guarantee that this mount unit will finish execution before docker starts.

How about doing WantedBy=docker.service and that should ensure that this activates

rhvgoyal: I am thinking this probably should be `RequiredBy` dependency. That is if we fail to mount this device, then docker should fail.

rhatdan: I would prefer not to tie this script to docker if we can help it.  I want to start looking at converting the script to container-storage-setup and being able to use it with OCID.

I am pretty sure that the mount unit will have to be mounted before Multi-User Target and Docker.service or OCID.service will not be started until multi-user target is started.  No need to link them.

rhatdan: If you look at docker.service, you will see that it is 
[Install]
WantedBy=multi-user.target

Which means when you enable the docker,service it gets started when systemd starts the multi-user-target.

rhvgoyal: @rhatdan IIUC, so multi-user.target wants docker.service. And it also wants docker-var-lib.mount. That means systemd can activate docker.service and docker-var-lib.mount in any order. There is no guarantee that docker-var-lib.mount will finish execution before docker.service. (Atleast multi-user.target does not seem to guarantee that)

rhvgoyal: Should we use `WantedBy=local-fs.target` instead? And possibly add `Before=local-fs.target` as well. 

man `systemd.special` says systemd-fstab-generator automatically adds dependencies for entries in /etc/fstab file. Given this will not be in fstab file, that probably means we will have to add dependencies ourselves.

rhatdan: Yes I was thinking the same thing. Should it be INSTALL local-fs.target?

rhatdan:
Systemd guys say to add

 Before=local-fs.target

docker-storage-setup.sh
rhvgoyal: @shishir-a412ed my comment of using a variable for `/etc/sysconfig/docker` is yet to be addressed.

docker-storage-setup.sh
```diff
@@ -865,15 +990,74 @@ setup_storage() {
...
+    do
+        if [ "$flag" = true ];then
+           DOCKER_ROOT_DIR=$opt
+           flag=false
+           continue
+        fi
+	case "$opt" in
+            "-g"|"--graph")
+                flag=true
+                ;;
```
rhvgoyal: Does this handle the case of two or more spaces -g and value. For example `-g    /var/lib/foo`

docker-storage-setup.sh
```diff
@@ -410,6 +412,19 @@ check_docker_storage_metadata() {
   exit 1
 }
 
+reset_docker_root_volume () {
+  local mp
+  if docker_root_volume_exists; then
+     if ! mp=$(docker_root_lv_mountpoint); then
+        return 1
```
rhvgoyal: It will be a good idea to give an error message here. Say

`Failed to determine where /dev/$VG/$DOCKER_ROOT_LV is mounted`

docker-storage-setup.sh
```diff
@@ -839,6 +854,103 @@ get_existing_storage_driver() {
...
 }
 
+docker_root_volume_exists() {
+    lvs $VG/$DOCKER_ROOT_LV_NAME > /dev/null 2>&1 && return 0
+    return 1
+}
+
+# This returns the mountpoint of $DOCKER_ROOT_LV_NAME
+docker_root_lv_mountpoint() {
+  local mounts target
```
rhvgoyal: local variable `target` is unused now.

docker-storage-setup.sh
rhvgoyal: @shishir-a412ed Change this file to add "Before=local-fs.target" in [Unit] section. And I think we can retain WantedBy=local-fs.target in [Install] section. Though I am not sure if WantedBy can go in [Unit] section or not.

rhvgoyal:
@shishir-a412ed I am testing this patch now.

After creating docker-root-lv, I tried "atomic storage reset" and it asks me whether I really want to remove an active volume. I don't think we need it.  Like lvm thin pool, lets first deactivate the docker root volume and then remove it to get rid of this message.

# atomic storage reset
Do you really want to remove active logical volume fedora/docker-root-lv? [y/n]: y
Removed /etc/systemd/system/local-fs.target.wants/var-lib-docker.mount.

Also I see this "Removed var-lib-docker.mount" message. We don't need to see it. I would say send stdout to /dev/null/

rhvgoyal:
After doing "atomic storage reset" once, I did it again and got following.

atomic storage reset
Failed to disable unit: No such file or directory

We should not have assumption that unit file will be there necessarily.

rhvgoyal:
Also, I unmounted /var/lib/docker manually and did "atomic storage reset" and it complains failed to determine mount point. That should not happen. we should not enforce that file is mounted.

atomic storage reset
ERROR: Failed to determine where /dev/fedora/docker-root-lv is mounted

rhvgoyal:
@rhatdan I removed device from my VM and rebooted. System waits for 1 minute 30 seconds for device to come up. Till then boot stops. 

After 1 minute 30 second timeout, boot continues.

rhvgoyal:
Well, another problem is that docker starts and sets up new overlay2 on /var/lib/docker/. So if disk shows up during next reboot, then docker will switch to old view again.

And all this is happening because docker does not fail if docker-storage-setup fails. This automatic creation of loop devices need is creating lot of problems.

rhatdan:
Is docker-storage-setup a systemd unit file?  If yes put the requirement there?  That the mount has to succeed.

rhvgoyal:
Problem is not docker-storage-setup but docker. Docker unit file has Wants= dependency on docker-storage-setup and not Requires= dependency. That means docker will continue to start even if docker-storage-setup failed.

rhvgoyal:
@shishir-a412ed had a conversation with dan and here is the summary.

- Change dependency in mount unit file to docker-storage-setup instead of local-fs.target

  Before=docker-storage-setup.service
  WantedBy=docker-storage-setup.service

  This should help with issue of whole boot hanging for 90 seconds if device or volume group or logical volume does not show up.

- We will soon have default of overlay2 in fedora. At that time we will modify docker.service file to Requires=docker-storage-setup as well. That will make sure if docker-storage-setup fails, then
docker fails as well. That way we will not run into odd situations if underlying disk or volume group does not come up.

rhvgoyal:
@shishir-a412ed I tested it with `-g foo` in /etc/sysconfig/docker and it gives following error.


./docker-storage-setup.sh 
INFO: Device /dev/vdd is already partitioned and is part of volume group docker-vg
  Wiping xfs signature on /dev/docker-vg/docker-root-lv.
  Logical volume "docker-root-lv" created.
Job for oo.mount failed.
See "systemctl status oo.mount" and "journalctl -xe" for details.
ERROR: Failed to mount docker root volume docker-root-lv on foo

docker-storage-setup.sh
```diff
@@ -410,6 +412,36 @@ check_docker_storage_metadata() {
...
+  fi
+  # If the user has manually unmounted ${DOCKER_ROOT_DIR}, mountpoint (mp)
+  # will be empty. Extract ${mp} from /etc/sysconfig/docker in that case.
+  if [ -z "$mp" ];then
+    if ! get_docker_root_dir; then
+      return 1
+    fi
+    mp=${DOCKER_ROOT_DIR}
+  fi
+  filename=$(echo ${mp}.mount|sed 's|/|-|g' | cut -b 2-)
```
rhvgoyal: I think here you need to make sure that mp starts with /.

rhvgoyal:
I tried passing `-g /var/lib/foo/` and that fails too.

shishir-a412ed:
@rhvgoyal I tried this "I tried passing -g /var/lib/foo/ and that fails too."
For me, d-s-s is able to handle this case, remove the docker root volume and exit gracefully.

I will come to your desk, and we can check this out.

Shishir

rhvgoyal:
./docker-storage-setup.sh 
INFO: Device /dev/vdd is already partitioned and is part of volume group docker-vg
  Wiping xfs signature on /dev/docker-vg/docker-root-lv.
  Logical volume "docker-root-lv" created.
Failed to start var-lib-foo-.mount: Unit var-lib-foo-.mount is not loaded properly: Invalid argument.
See system logs and 'systemctl status var-lib-foo-.mount' for details.
ERROR: Failed to mount docker root volume docker-root-lv on /var/lib/foo/

rhvgoyal:
Lets use `realpath -m <path>` to canonicalize path and then use sed on this.

rhvgoyal:
LGTM

I have done lot of local testing and it seems to work fine for me. Once jenkins is working again, I will merge this PR.

rhvgoyal:
Once this gets merged we need two things

- some automated test to test this.
- facility to save local metadata so that we can save filename of mount unit file.

rhvgoyal:
@rh-atomic-bot retest

rhvgoyal:
@rh-atomic-bot retest this please

rhvgoyal:
@rh-atomic-bot try

rhvgoyal:
@rh-atomic-bot r+

PR #176, opened 2016-12-21
get_docker_root_dir: use realpath only when path is not empty.

shishir-a412ed:
If a user has not specified `-g` or `--graph` for $DOCKER_ROOT_DIR
in `/etc/sysconfig/docker`, then $DOCKER_ROOT_DIR will default to
`/var/lib/docker`. In that case, $path will be empty, and `realpath`
would fail. This PR addresses that issue.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -1049,6 +1049,9 @@ get_docker_root_dir(){
                 ;;
         esac
     done
+    if [ -z "$path" ];then
+      return
+    fi
     if ! DOCKER_ROOT_DIR=$(realpath -m $path);then
       Fatal "realpath failed on $path"
     fi


shishir-a412ed:
ping @rhatdan @rhvgoyal

docker-storage-setup.sh
```diff
@@ -1049,8 +1049,10 @@ get_docker_root_dir(){
                 ;;
         esac
     done
-    if ! DOCKER_ROOT_DIR=$(realpath -m $path);then
-      Fatal "realpath failed on $path"
+    if [ ! -z "$path" ];then
```
rhatdan: Won't this end up with DOCKER_ROOT_DIR not defined.

Why not just do 
 if [ ! -z "$path" ];then
          path = $DEFAULT_CONTAINER_PATH
fi

rhatdan: Ok simplify the patch to 

if [ ! z ${path} ]; then
     return
fi

shishir-a412ed: @rhatdan Sorry I am not following. If $path is not empty, why would we return ? If $path is not empty, we should get the realpath: $(realpath -m $path) ?

rhatdan:
Sorry code was wrong, check 
```
if [ -z ${path} ]; then
return
fi
```

shishir-a412ed:
@rhatdan Fixed.

rhatdan:
LGTM

PR #177, opened 2017-01-04
Test-1: Tests for PR # 175

shishir-a412ed:

### DIFF
--- a/tests/010-test-use-devs-to-create-docker-root-volume.sh
+++ b/tests/010-test-use-devs-to-create-docker-root-volume.sh
@@ -1,0 +1,69 @@
+source $SRCDIR/libtest.sh
+
+# Test DOCKER_ROOT_VOLUME= directive. Returns 0 on success and 1 on failure.
+test_docker_root_volume() {
+  local devs=$TEST_DEVS
+  local test_status=1
+  local testname=`basename "$0"`
+  local vg_name="dss-test-foo"
+  local docker_root_lv_name="docker-root-lv"
+
+  # Error out if any pre-existing volume group vg named dss-test-foo
+  if vg_exists "$vg_name"; then
+    echo "ERROR: $testname: Volume group $vg_name already exists." >> $LOGS
+    return $test_status
+  fi
+
+  # Create config file
+  cat << EOF > /etc/sysconfig/docker-storage-setup
+DEVS="$devs"
+VG=$vg_name
+DOCKER_ROOT_VOLUME=yes
+EOF
+
+ # Run docker-storage-setup
+ $DSSBIN >> $LOGS 2>&1
+
+ # Test failed.
+ if [ $? -ne 0 ]; then
+    echo "ERROR: $testname: $DSSBIN Failed." >> $LOGS
+    cleanup $vg_name "$devs"
+    return $test_status
+ fi
+
+  # Make sure $DOCKER_ROOT_VOLUME {docker-root-lv} got created
+  # successfully.
+  if ! lv_exists "$vg_name" "$docker_root_lv_name"; then
+    echo "ERROR: $testname: Logical Volume $docker_root_lv_name does not exist." >> $LOGS
+    cleanup $vg_name "$devs"
+    return $test_status
+  fi
+
+  # Make sure $DOCKER_ROOT_VOLUME {docker-root-lv} is
+  # mounted on /var/lib/docker
+  local mnt
+  mnt=$(findmnt -n -o TARGET --first-only --source /dev/${vg_name}/${docker_root_lv_name})
+  if [ "$mnt" != "/var/lib/docker" ];then
+   echo "ERROR: $testname: Logical Volume $docker_root_lv_name is not mounted on /var/lib/docker." >> $LOGS
+   cleanup $vg_name "$devs"
+   return $test_status
+  fi
+
+  cleanup_container_root_volume $vg_name $docker_root_lv_name $mnt
+  cleanup $vg_name "$devs"
+  return 0
+}
+
+cleanup_container_root_volume(){
+  local vg_name=$1
+  local lv_name=$2
+  local mount_path=$3
+
+  umount $mount_path >> $LOGS 2>&1
+  lvchange -an $vg_name/${lv_name} >> $LOGS 2>&1
+  lvremove $vg_name/${lv_name} >> $LOGS 2>&1
+
+  cleanup_mount_file $mount_path
+}
+
+test_docker_root_volume
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -10,7 +10,14 @@ vg_exists() {
     fi
   done
   return 1
-  
+}
+
+# Tests if the logical volume lv_name exists
+lv_exists() {
+  local vg_name=$1
+  local lv_name=$2
+  lvs $vg_name/$lv_name > /dev/null 2>&1 && return 0
+  return 1
 }
 
 clean_config_files() {
@@ -74,6 +81,17 @@ cleanup() {
   udevadm settle
   clean_config_files
   wipe_signatures "$devs"
+}
+
+cleanup_mount_file() {
+  local mount_path=$1
+  local mount_filename=$(echo $mount_path|sed 's/\//-/g'|cut -c 2-)
+
+  if [ -f "/etc/systemd/system/$mount_filename.mount" ];then
+    systemctl disable $mount_filename.mount >/dev/null 2>&1
+    rm /etc/systemd/system/$mount_filename.mount >/dev/null 2>&1
+    systemctl daemon-reload
+  fi
 }
 
 cleanup_soft_links() {


rhvgoyal:
@shishir-a412ed can we put some description in commit, explaining what this test is testing.

tests/010-test-use-devs-to-create-docker-root-volume.sh
```diff
@@ -0,0 +1,64 @@
...
+
+   # Make sure $DOCKER_ROOT_VOLUME {docker-root-lv} is
+   # mounted on /var/lib/docker
+   local mnt
+   mnt=$(findmnt -n -o TARGET --first-only --source /dev/${vg_name}/${docker_root_lv_name})
+   if [ "$mnt" == "/var/lib/docker" ];then
+      test_status=0
+   fi
+
+  $DSSBIN --reset >> $LOGS 2>&1
```
rhvgoyal: I think docker storage reset should be a separate test. First test should just test if DOCKER_ROOT_VOLUME works fine or not.

tests/010-test-use-devs-to-create-docker-root-volume.sh
```diff
@@ -0,0 +1,64 @@
...
+    echo "ERROR: $testname: Volume group $vg_name already exists." >> $LOGS
+    return $test_status
+  fi
+
+  # Create config file
+  cat << EOF > /etc/sysconfig/docker-storage-setup
+DEVS="$devs"
+VG=$vg_name
+DOCKER_ROOT_VOLUME=yes
+DOCKER_ROOT_VOLUME_SIZE=40%FREE
```
rhvgoyal: we should not have to specify size. Let it be default. Want this test to take default path where users will simply specify DOCKER_ROOT_VOLUME=yes.

tests/010-test-use-devs-to-create-docker-root-volume.sh
```diff
@@ -0,0 +1,64 @@
...
+ if [ $? -ne 0 ]; then
+    echo "ERROR: $testname: $DSSBIN Failed." >> $LOGS
+    cleanup $vg_name "$devs"
+    return $test_status
+ fi
+
+  # Make sure $DOCKER_ROOT_VOLUME {docker-root-lv} got created
+  # successfully.
+  if lv_exists "$docker_root_lv_name"; then
+    test_status=0
```
rhvgoyal: Don't set test_status=0 yet. As test will be successful only when you have done mountpoint check as well. You can instead check if lv exists or not and if not, then do the cleanup and return.

tests/010-test-use-devs-to-create-docker-root-volume.sh
rhvgoyal: I think we already have test for devmapper and overlay cleanup. We probably should write one for docker-root-volume cleanup as well.

tests/libtest.sh
```diff
@@ -10,7 +10,18 @@ vg_exists() {
...
   return 1
-  
+}
+
+# Tests if the logical volume lv_name exists
+lv_exists() {
+  local lv lv_name="$1"
+
+  for lv in $(lvs --noheadings -o lv_name); do
+    if [ "$lv" == "$lv_name" ]; then
```
rhvgoyal: Do we really have to search for lv name in output. Can we just pass the expected lv to lvs and that should tell us if it exists or not?

lvs docker-vg/docker-root-lv ?

tests/010-test-use-devs-to-create-docker-root-volume.sh
shishir-a412ed: Yeah I saw those tests (devmapper and overlay cleanup) while developing this one.
After running this test, we would still need to unmount and remove DOCKER_ROOT_VOLUME.
I felt if I use `docker-storage-setup --reset`, it would test the `reset` functionality and do the cleanup at the same time. Otherwise it would be duplicating the effort.

tests/010-test-use-devs-to-create-docker-root-volume.sh
shishir-a412ed: Initially I did not specify `DOCKER_ROOT_VOLUME_SIZE=40%FREE` in the test. I was expecting docker-storage-setup to pick up the default value from `/usr/lib/docker-storage-setup/docker-storage-seutp`. However the CI started failing. I believe it was because CI version of that file did not have these new storage options `DOCKER_ROOT_VOLUME` and `DOCKER_ROOT_VOLUME_SIZE`.

tests/libtest.sh
shishir-a412ed: I used existing logic and structure of `vg_exists()` while developing `lv_exists()`. I wanted to keep `lv_exists()` as close as possible to existing code style. Does this work for you ? or should we change both (vg_exists, lv_exists) to incorporate this logic ?

tests/010-test-use-devs-to-create-docker-root-volume.sh
shishir-a412ed: Fixed.

shishir-a412ed:
@rhvgoyal PTAL.

tests/010-test-use-devs-to-create-docker-root-volume.sh
```diff
@@ -0,0 +1,66 @@
...
+    cleanup $vg_name "$devs"
+    return $test_status
+  fi
+
+   # Make sure $DOCKER_ROOT_VOLUME {docker-root-lv} is
+   # mounted on /var/lib/docker
+   local mnt
+   mnt=$(findmnt -n -o TARGET --first-only --source /dev/${vg_name}/${docker_root_lv_name})
+   if [ "$mnt" == "/var/lib/docker" ];then
+      test_status=0
```
rhvgoyal: I think test_status should be set to 0 only once when we have done all the checks.

tests/010-test-use-devs-to-create-docker-root-volume.sh
```diff
@@ -0,0 +1,66 @@
...
+   fi
+
+  $DSSBIN --reset >> $LOGS 2>&1
+  # Test failed.
+  if [ $? -eq 0 ]; then
+     if [ ! -e /etc/sysconfig/docker-storage ]; then
```
rhvgoyal: I don't think we need to check for presence of this file. here. We should write a separate test which tests the reset functionality. And then we should make sure logical volume went away, var-lib-docker.mount file went away and
/etc/sysconfig/docker-storage went away.

tests/010-test-use-devs-to-create-docker-root-volume.sh
```diff
@@ -0,0 +1,66 @@
...
+          test_status=0
```
rhvgoyal: Ok, how about following. Write two tests.
- First tests to test volume creation funcitonality
- second test to test reset funcitonality.

In first test you can still use --reset to cleanup but don't use its return code to determine test status. This will be something like `cleanup $vg_name`. It can fail too but we don't determine success/failure of tests based on cleanup.

So first test will ignore the --reset results. It will PASS/FAIL based on whether volume creation was successful or not.

Second test will fail if reset did not work properly.

shishir-a412ed: Based on our discussion on irc, I am not using `--reset` for volume creation tests. I am writing a manual methods for doing cleanups.

shishir-a412ed:
@rhvgoyal PTAL.

tests/libtest.sh
rhvgoyal: shishir, now you have code in extra_volume_exists(), in docker-storage-setup.sh. It should be much simpler to use that.

tests/libtest.sh
```diff
@@ -76,6 +87,23 @@ cleanup() {
   wipe_signatures "$devs"
 }
 
+cleanup_container_root_volume() {
```
rhvgoyal: This function feels very specific. It probably is better in the test itself and not part of library.

Or, break this down into smaller pieces. That is have a helper function to cleanup mount unit file in libtest.sh and logic to remove lvm keep it in test itself.

shishir-a412ed:
@rhvgoyal PTAL.

rhvgoyal:
@shishir-a412ed On my local machine, once in a while test is failing. It says mkfs failed as device or resource is busy.

mkfs.xfs: cannot open /dev/dss-test-foo/docker-root-lv: Device or resource busy
ERROR: Failed to create filesystem on /dev/dss-test-foo/docker-root-lv

rhvgoyal:
I introduced "udevadm settle" after lvcreate and before mkfs and that seems to have fixed the issue. So looks like udev is keeping device busy. I will open another PR to fix it.

tests/010-test-use-devs-to-create-docker-root-volume.sh
```diff
@@ -0,0 +1,69 @@
...
+cleanup_container_root_volume(){
+  local vg_name=$1
+  local lv_name=$2
+  local mount_path=$3
+
+  umount $mount_path >> $LOGS 2>&1
+  lvchange -an $vg_name/${lv_name} >> $LOGS 2>&1
+  lvremove $vg_name/${lv_name} >> $LOGS 2>&1
+
+  cleanup_mount_file $mountpath
```
rhvgoyal: This should be `$mount_path` and not `$mountpath`. As a result, right now /etc/systemd/system/var-lib-docker.mount is not being cleaned up on the system after the tests.

rhvgoyal:
I used `udevadm settle` and problem went away. Now I removed it and still I can't see the problem. I think this is a race which does not happen all the time. For now, fix the mount_path thing and I will merge it. If problem happens, will create separate PR to add udevadm settle.

shishir-a412ed:
@rhvgoyal PTAL.

rhvgoyal:
LGTM

rhvgoyal:
@rh-atomic-bot r+

rhvgoyal:
@rh-atomic-bot r+ 4f2ce20

PR #180, opened 2017-01-10
REMOVE ALL DOCKER_ references

rhatdan:
This is part of the effort to convert docker-storage-setup to
container-storage-setup

First we want to remove all references to DOCKER in environment variables.

Users could have set
DOCKER_ROOT_VOLUME, DOCKER_ROOT_VOLUME_SIZE and EXTRA_DOCKER_STORAGE_OPTIONS

so we add the following to check for these.
if [ -z "$ROOT_VOLUME" ]; then
    ROOT_VOLUME="$DOCKER_ROOT_VOLUME"
fi
if [ -z "$ROOT_VOLUME_SIZE" ]; then
    ROOT_VOLUME_SIZE="$DOCKER_ROOT_VOLUME_SIZE"
fi
if [ -z "$EXTRA_STORAGE_OPTIONS" ]; then
    EXTRA_STORAGE_OPTIONS="$EXTRA_DOCKER_STORAGE_OPTIONS"
fi


rhatdan:
@shishir-a412ed PTAL

rhatdan:
@rhvgoyal PTAL

docker-storage-setup.1
```diff
@@ -36,7 +36,7 @@ STORAGE_DRIVER:
       overlay2 and "". Empty string tells docker-storage-setup to
       not do any storage setup.
 
-EXTRA_DOCKER_STORAGE_OPTIONS:
+EXTRA_STORAGE_OPTIONS:
       A set of extra options that should be passed to the Docker
       daemon.
```
shishir-a412ed: Should we call this "Container daemon" ?

docker-storage-setup.conf
```diff
@@ -4,9 +4,9 @@
 STORAGE_DRIVER=devicemapper
 
 # A set of extra options that should be appended to the generated
-# DOCKER_STORAGE_OPTIONS string. These options will be passed to the Docker
+# STORAGE_OPTIONS string. These options will be passed to the Docker
 # daemon as-is and should be valid Docker storage options.
```
shishir-a412ed: These options will be passed to the "Container daemon" as-is and should be valid "Container" storage options.

shishir-a412ed:
@rhatdan 
a) Need to fix the tests.
b) When we move from `docker` to `container` in `d-s-s`, I think we also need to think about how it would affect `docker-latest`. Right now @lsm5 has a script that replaces all occurrences of keyword `docker` to `docker-latest` when it builds `docker-storage-setup` for `docker-latest`. I don't think this patch affects that right now, but we need to keep that in mind while moving forward.

Shishir

rhatdan:
Right, the true fix to this is to change the script so that it works with multiple config, so that we have one script but different configs, which is what we are working towards.

docker-storage-setup.sh
```diff
@@ -1112,6 +1112,15 @@ fi
 if [ -e /etc/sysconfig/docker-storage-setup ]; then
   source /etc/sysconfig/docker-storage-setup
 fi
+if [ -z "$ROOT_VOLUME" ]; then
+    ROOT_VOLUME="$DOCKER_ROOT_VOLUME"
+fi
```
shishir-a412ed: Since you have already sourced `/usr/lib/docker-storage-setup/docker-storage-setup`: `ROOT_VOLUME=no` (default value) has been set. So If I as a user set `DOCKER_ROOT_VOLUME=yes` in `/etc/sysconfig/docker-storage-setup`, `ROOT_VOLUME` will still be `no` and this condition will never get executed.

shishir-a412ed:
@rhatdan `/usr/bin/docker-storage-setup --reset` is failing in the tests {05, 06, 07, 08}.
You can try to run `/usr/bin/docker-storage-setup` manually first and setup the storage.
And then try to reset it manually `/usr/bin/docker-storage-setup --reset`. 

I am also looking into why `--reset` is failing.

Shishir

shishir-a412ed:
@rhatdan 
These IF conditions
https://github.com/projectatomic/docker-storage-setup/pull/180#issue-199947956
should be changed to this:
````
if [ -v "$DOCKER_ROOT_VOLUME" ]; then
    ROOT_VOLUME="$DOCKER_ROOT_VOLUME"
fi
if [ -v "$DOCKER_ROOT_VOLUME_SIZE" ]; then
    ROOT_VOLUME_SIZE="$DOCKER_ROOT_VOLUME_SIZE"
fi
if [ -v "$EXTRA_DOCKER_STORAGE_OPTIONS" ]; then
    EXTRA_STORAGE_OPTIONS="$EXTRA_DOCKER_STORAGE_OPTIONS"
````

Shishir

File: docker-storage-setup.sh [hunk 4]
```diff
@@ -1062,22 +1062,22 @@ setup_docker_root_dir() {
...
   # Directory does not exist. Create one.
-  mkdir -p $DOCKER_ROOT_DIR
+  mkdir -p $ROOT_DIR
   return $?
 }
 
 reset_storage() {
-  
-    if [ "$DOCKER_ROOT_VOLUME" == "yes" ];then
+    [ "$ROOT_VOLUME" != "yes" ] && return 0
```
shishir-a412ed: Please remove this line. This is causing the tests to fail. Since the default value of `$DOCKER_ROOT_VOLUME` or `$ROOT_VOLUME` is `no`. This will just return and will never execute `reset_storage()`.

rhatdan: Thanks I have fixed this now.

docker-storage-setup.sh
rhatdan: Good catch, I have reversed the logic.  If a user sets a DOCKER_ variable it will override.

docker-storage-setup.conf
rhatdan: I plan on other pull requests to change the Documentation and function names,  Want to take it one step at a time.

rhvgoyal:
If we are changing input environment variable names we should do couple of things.

- At run time we should warn that DOCKER_* is being deprecated and recommend to use the new variable. That way sometime down the line we can start erroring out if we see DOCKER_* vaiables
in input.

- Also we should probably modify man page to retain old varibable name and mention old variable is being deprecated.

docker-storage-setup.sh
```diff
@@ -55,13 +55,13 @@ set -e
...
-DOCKER_ROOT_DIR="/var/lib/docker"
+ROOT_LV_NAME="docker-root-lv"
+ROOT_DIR="/var/lib/docker"
 
-DOCKER_STORAGE="/etc/sysconfig/docker-storage"
+STORAGE="/etc/sysconfig/docker-storage"
 STORAGE_DRIVERS="devicemapper overlay overlay2"
 
-DOCKER_METADATA_DIR="/var/lib/docker"
+METADATA_DIR="/var/lib/docker"
```
rhvgoyal: This probably can continue to be DOCKER_METADATA_DIR because that's what this variable will continue to represent.

docker-storage-setup.conf
shishir-a412ed: Makes sense

shishir-a412ed:
@rhatdan I don't see the new changes. Have you pushed the changes to github ?

Shishir

docker-storage-setup.sh
rhatdan: No because we are going to want to pass in the METADATA_DIR via config.  /var/lib/containers will be the METADATA_DIR for OCID.

rhatdan:
Down to one failure

docker-storage-setup.sh
rhvgoyal: The whole purpose of this METADATA_DIR was to check if docker already has /var/lib/dokcer/devicemapper dir which tells it that docker has been using loop devices and it errors out in this case.

This is another case of docker-storage-setup probing metadata of caller which is completely wrong. We just did it for better user experience. So in long term I would like to get rid of this functionality and not expose this knob to OCID and other tools at all. Before they call docker-storage-setup, they need to make sure they are not using loop devices. Its none of the docker-storage-setup business.

rhatdan:
@shishir-a412ed @rhvgoyal All Tests pass PTAL

rhvgoyal:
@rhatdan did you see my comment about giving out message that old environment variables are being deprecated. And to make transition smooth over upgrade, I think we need to keep reference to old variables in man page as well?

docker-storage-setup.1
```diff
@@ -36,9 +36,10 @@ STORAGE_DRIVER:
       overlay2 and "". Empty string tells docker-storage-setup to
       not do any storage setup.
 
-EXTRA_DOCKER_STORAGE_OPTIONS:
+EXTRA_STORAGE_OPTIONS:
       A set of extra options that should be passed to the Docker
```
rhvgoyal: I think docker-storage-setup will have to know about the environment for which storage is being generated. Because output file format will depend on that. For example, --storage-opt key=value is docker format and other run time might have different format.

docker-storage-setup.1
```diff
@@ -36,9 +36,10 @@ STORAGE_DRIVER:
...
       daemon.
+      NOTE: This option replaces EXTRA_DOCKER_STORAGE_OPTIONS which has been deprecated
```
rhvgoyal: How about following format for options in man page.

EXTRA_DOCKER_STORAGE_OPTIONS: (DEPRECATED)
EXTRA_STORAGE_OPTIONS:
       Description of option....

docker-storage-setup.1
```diff
@@ -98,36 +99,38 @@ WIPE_SIGNATURES:
       overwritten as suitable. This also means that if there is any
       data on disk, it will be lost.
 
-DOCKER_ROOT_VOLUME:
+ROOT_VOLUME:
```
rhvgoyal: Retain 

DOCKER_ROOT_VOLUME: (DEPRECATED)

docker-storage-setup.1
```diff
@@ -98,36 +99,38 @@ WIPE_SIGNATURES:
...
       "lvcreate -L" as well as some values acceptable to to 
       "lvcreate -l". If user intends to pass values acceptable 
       to "lvcreate -l", then only those values which contains "%" 
       in syntax are acceptable.  If value does not contain "%" it 
       is assumed value is suitable for "lvcreate -L".
 
       NOTE: If both STORAGE_DRIVER=devicemapper and 
-      DOCKER_ROOT_VOLUME=yes is set, docker-storage-setup would set
+      ROOT_VOLUME=yes is set, docker-storage-setup would set
       up the thin pool for devicemapper first, followed by docker
```
rhvgoyal: I really don't like renaming DOCKER_ROOT_VOLUME to ROOT_VOLUME because it is strongly tied to bind mounting lv on /var/lib/docker.

Can we decouple that first. How about create a new variable ROOT_VOLUME_MOUNT_DIR.  By default it will scan /etc/sysconfig/docker for -g otherwise use /var/lib/docker. Otherwise user can override it. Now ROOT_VOLUME will be mounted on ROOT_VOLUME_MOUNT_DIR.

@shishir-a412ed can you create a separate PR to take care of it.  Once this change is in, it decouples ROOT_VOLUME with mount point location and renaming DOCKER_ROOT_VOLUME to ROOT_VOLUME is much more reasonable.

docker-storage-setup.1
rhatdan: I don't think we should make them that prominent.

docker-storage-setup.1
rhatdan: Lets discuss this.  I don't see the strong link.

docker-storage-setup.1
rhatdan: But these tools will set environment for different tools, so the admin/packager will know the difference.   OCID-storage-setup will have defaults that explains these flags.

rhatdan:
@rh-atomic-bot retest

shishir-a412ed:
@rhatdan By looking at the changes, it looks like you are using `sed` for the substitutions. I don't think that would work here (This would require some manual work). I will give you 1-2 examples. 

a) In `setup_docker_root_lv_fs` changing `DOCKER_ROOT_VOLUME_SIZE` to `CONTAINER_ROOT_LV_SIZE`. `DOCKER_ROOT_VOLUME_SIZE` continue to exist as we have agreed on maintaining backward compatibility.

b) Same thing in test 011, where `DOCKER_ROOT_VOLUME` is replaced with `ROOT_VOLUME`. `ROOT_VOLUME` is not even a valid option :)

I think a lot of this is already fixed in master branch, however rebase might be bit tricky as there are many conflicts. Opening a new PR against the current master would be easier.

Shishir

rhatdan:
I believe A is fixed in the first few lines of code, whereas B has been fixed in the test code.

tests/010-test-use-devs-to-create-docker-root-volume.sh
```diff
@@ -19,6 +19,7 @@ test_docker_root_volume() {
 DEVS="$devs"
 VG=$vg_name
 DOCKER_ROOT_VOLUME=yes
+DOCKER_ROOT_VOLUME_SIZE=40%FREE
```
shishir-a412ed: Why are we setting `DOCKER_ROOT_VOLUME_SIZE=40%FREE`? I believe this should be picked by default.

rhatdan:
@shishir-a412ed @rhvgoyal Tests pass, please start the review.

tests/011-test-docker-root-volume-cleanup.sh
```diff
@@ -21,6 +21,7 @@ test_reset_docker_root_volume() {
 DEVS="$devs"
 VG=$vg_name
 DOCKER_ROOT_VOLUME=yes
+DOCKER_ROOT_VOLUME_SIZE=20%FREE
```
shishir-a412ed: Any specific reason we are setting this to `20%FREE`? The default value should automatically get picked.

docker-storage-setup.conf
```diff
@@ -84,63 +85,33 @@ DEVICE_WAIT_TIMEOUT=60
 # as this means if there was any leftover data on disk, it will be lost.
 WIPE_SIGNATURES=false
 
-# By default no new volume and filesystem will be setup for docker
-# root dir. Docker creates /var/lib/docker/ on top of underlying filesystem
-# for storing images and containers.
+# By default no new volume and filesystem will be setup for container runtime
+# root dir. For example the docker engin creates /var/lib/docker/ on top of
```
shishir-a412ed: s/engin/engine

docker-storage-setup.sh
```diff
@@ -1199,6 +1197,17 @@ fi
 if [ -e /etc/sysconfig/docker-storage-setup ]; then
   source /etc/sysconfig/docker-storage-setup
 fi
+if [ ! -z "$DOCKER_ROOT_VOLUME" ]; then
+    Info "DOCKER_ROOT_VOLUME is deprecated, and will be removed soon. Use CONTAINER_ROOT_LV_MOUNT_PATH instead."
+fi
```
shishir-a412ed: This check `if [ ! -z "$DOCKER_ROOT_VOLUME" ]` is already handled in `check_storage_options()`.
https://github.com/projectatomic/docker-storage-setup/blob/master/docker-storage-setup.sh#L1032

tests/010-test-use-devs-to-create-docker-root-volume.sh
shishir-a412ed: I see, you have removed `DOCKER_ROOT_VOLUME_SIZE=40%FREE` from the .conf file. So it won't be picked up by default. so that means if the user wants to use `DOCKER_ROOT_VOLUME` he has to explicitly pass both these options.

docker-storage-setup.1
```diff
@@ -98,56 +100,23 @@ WIPE_SIGNATURES:
...
+     Creates a logical volume named CONTAINER_ROOT_LV_NAME and mounts
+     it at the specified path. By default no new logical volume will
+     be created. For example:
+     \f[B]CONTAINER_ROOT_LV_MOUNT_PATH=/var/lib/containers\f[]
+     would carve out a logical volume, format it with an XFS filesystem
+     and mount it on /var/lib/containers. 
+
+     Note: DOCKER_ROOT_VOLUME is deprecated. Specifying
+     DOCKER_ROOT_VOLUME and CONTAINER_ROOT_LV_MOUNT_PATH
+     is not allowed.
```
shishir-a412ed: This seems a little bit confusing to me. "Specifying DOCKER_ROOT_VOLUME and CONTAINER_ROOT_LV_MOUNT_PATH is not allowed." . This seems to imply that specifying these two options are not allowed, which is not the case. You can specify both DOCKER_ROOT_VOLUME and CONTAINER_ROOT_LV_MOUNT_PATH however you cannot use them together. Following is the description in current master.

https://github.com/projectatomic/docker-storage-setup/blob/master/docker-storage-setup.conf#L97-L102

tests/010-test-use-devs-to-create-docker-root-volume.sh
rhatdan: We are removing this from defaults, since we don't want people using it any longer.

shishir-a412ed:
I tried to run this by setting the following in /etc/sysconfig/docker-storage-setup.

````
DEVS=/dev/vdb
VG=atomicos
DOCKER_ROOT_VOLUME=yes
````
Output: http://pastebin.test.redhat.com/450875
Since `DOCKER_ROOT_VOLUME_SIZE=40%FREE` is no more the default i.e we have removed this field from `/usr/lib/docker-storage-setup/docker-storage-setup`.  If the user does not specify both,  we should error out. Right now it is leaving docker-storage-setup in a messy state which require a cleanup in this case.

File: tests/011-test-docker-root-volume-cleanup.sh (hunk 11 above)
Comment (rhatdan) [inline tests/011-test-docker-root-volume-cleanup.sh]: We dropped DOCKER_ROOT_VOLUME_SIZE as a default, and I thought it better to try multiple defaults.  We probably should have some failed cases.  But we really need cases that tests that this works the new prescribed way.

File: docker-storage-setup.sh (hunk 13 above)
Comment (rhatdan) [inline docker-storage-setup.sh]: Thanks removed.

File: docker-storage-setup.1 (hunk 14 above)
Comment (rhatdan) [inline docker-storage-setup.1]: changing to

     Note: DOCKER_ROOT_VOLUME is deprecated. Specifying
     DOCKER_ROOT_VOLUME and CONTAINER_ROOT_LV_MOUNT_PATH at the same time
     is not allowed.

**Comment by rhatdan** [comment]
It must be getting this default from elsewhere, since this was never specified in the config tests.  We can leave it in the default config with an error.  @rhvgoyal would this cause an error if the user has already run the script previously?

**Comment by rhatdan** [comment]
Once this patch is merged, I want to rename the actual executables and start adding --input and --output options, so that the packager can provide his own 

FOOBAR-storage-setup and FOOBAR-storage-setup.conf.

File: docker-storage-setup.sh [hunk 15]
```diff
@@ -58,7 +58,7 @@ POOL_LV_NAME="docker-pool"
 DOCKER_ROOT_LV_NAME="docker-root-lv"
 DOCKER_ROOT_DIR="/var/lib/docker"
 
-DOCKER_STORAGE="/etc/sysconfig/docker-storage"
+STORAGE="/etc/sysconfig/docker-storage"
```
rhvgoyal: Out about calling this variable `STORAGE_OUT_FILE`. This will fit nicely with your follow up patches as well. Just using `STORAGE` is too generic.

rhatdan: Sure, I will make the change.

docker-storage-setup.sh
```diff
@@ -215,11 +215,11 @@ get_devicemapper_config_options() {
   done )
 
   if extra_options_has_dm_fs; then
-    # dm.fs option defined in ${EXTRA_DOCKER_STORAGE_OPTIONS}
+    # dm.fs option defined in ${EXTRA_STORAGE_OPTIONS}
     dm_fs=""
   fi
 
-  storage_options="DOCKER_STORAGE_OPTIONS=\"--storage-driver devicemapper ${dm_fs} --storage-opt dm.thinpooldev=$POOL_DEVICE_PATH $(get_deferred_removal_string) $(get_deferred_deletion_string) ${EXTRA_DOCKER_STORAGE_OPTIONS}\""
+  storage_options="STORAGE_OPTIONS=\"--storage-driver devicemapper ${dm_fs} --storage-opt dm.thinpooldev=$POOL_DEVICE_PATH $(get_deferred_removal_string) $(get_deferred_deletion_string) ${EXTRA_STORAGE_OPTIONS}\""
```
rhvgoyal: This will break docker? Docker file parses for DOCKER_STORAGE_OPTIONS and not STORAGE_OPTIONS.

rhvgoyal:
I would say that in first pass, let us just change the name of internal variables. And then whatever is left we can have a closer look and decide what to do with these now.

rhvgoyal:
I think we should set a goal that we will not be breaking working docker. And changes will be fully backward compatible.

docker-storage-setup.sh
rhatdan: Then we should just use the transition name like we did at the bottom.  If user specifies DOCKER_STORAGE_OPTIONS we want and assign to STORAGE_OPTIONS.

rhatdan: I thought we only offered EXTRA_DOCKER_STORAGE_OPTIONS?

docker-storage-setup.sh
```diff
@@ -978,7 +976,7 @@ setup_extra_volume() {
 }
 
 setup_extra_lv_fs() {
-  [ -z "$CONTAINER_ROOT_LV_MOUNT_PATH" ] && [ -z "$CONTAINER_ROOT_LV_NAME" ] && return 0
+  [ -z "$RESOLVED_MOUNT_DIR_PATH" ] && [ -z "$CONTAINER_ROOT_LV_NAME" ] && return 0
```
rhvgoyal: What are we gaining by this change? Checking for CONTAINER_ROOT_LV_MOUNT_PATH is simpler in my opinion as it is well documented. RESOLVVED_MOUNT_DIR_PATH is just an internal variable.

rhatdan: You are buying that someone later coming and reading this code will know what is going on.
The very next line uses RESOLVED_MOUNT_DIR_PATH.  No where in the function is the CONTAINER_ROOT_LV_MOUNT_PATH used.

rhatdan:
I agree that we can not break docker.   So I believe that none of these changes will break docker at this point.

rhatdan:
The only external names that this patch is modifying is

DOCKER_ROOT_VOLUME_SIZE
EXTRA_DOCKER_STORAGE_OPTIONS

And those are warned about and stored in their new names.

rhatdan:
The next patch will have to handle 

if [ -e /usr/lib/docker-storage-setup/docker-storage-setup ]; then
  source /usr/lib/docker-storage-setup/docker-storage-setup
fi

Which should only be executed if your INFILE IS docker-storage-setup

docker-storage-setup.sh
rhvgoyal: That's input. In output file DOCKER_STORAGE_OPTIONS is set which is parsed by docker unit file.

ExecStart=/usr/bin/dockerd-current \
          --add-runtime oci=/usr/libexec/docker/docker-runc-current \
          --default-runtime=oci \
          --containerd /run/containerd.sock \
          --exec-opt native.cgroupdriver=systemd \
          --userland-proxy-path=/usr/libexec/docker/docker-proxy-current \
          $OPTIONS \
          $DOCKER_STORAGE_OPTIONS \

rhatdan:
@rhvgoyal Broke the patches into a series, and left most tests alone so they can test with the old "docker" way.

Once these patches are merged, I will create a patch to handle input and output file and make copy of all tests to test with input and output files, no Docker builtins.

docker-storage-setup.1
```diff
@@ -98,56 +100,23 @@ WIPE_SIGNATURES:
...
-     mount it on /var/lib/containers.
+     Creates a logical volume named CONTAINER_ROOT_LV_NAME and mounts
+     it at the specified path. By default no new logical volume will
+     be created. For example:
+     \f[B]CONTAINER_ROOT_LV_MOUNT_PATH=/var/lib/containers\f[]
+     would carve out a logical volume, format it with an XFS filesystem
+     and mount it on /var/lib/containers. 
+
+     Note: DOCKER_ROOT_VOLUME is deprecated. Specifying
+     DOCKER_ROOT_VOLUME and CONTAINER_ROOT_LV_MOUNT_PATH at the same time
```
rhvgoyal: `time` is not formatted properly when I read this man page using man. It needs to be on next line. Current line is too long.

docker-storage-setup.1
```diff
@@ -172,30 +141,28 @@ The options below should be specified as values acceptable to 'lvextend -L':
...
-           Value should be a number followed by a optional suffix.
-           "bBsSkKmMgGtTpPeE" are valid suffixes. If no suffix is
-           specified then value will be considered as megabyte unit.
-
-           Both upper and lower case suffix represent same unit of
-           size. Use suffix B for Bytes, S for sectors as 512 bytes, K
-           for kibibytes (1024 bytes), M for mebibytes
-           (1024 kibibytes), G for gibibytes, T for tebibytes, P for
-           pebibytes and E for exbibytes.
+DATA_SIZE: The desired size for the container runtime data LV.
```
rhvgoyal: This now specifies size of thin pool data LV. Should we make it explicit by adding key word `thin pool` in description.

docker-storage-setup.1
```diff
@@ -172,30 +141,28 @@ The options below should be specified as values acceptable to 'lvextend -L':
...
+	Defaults: 40% free space in the VG after the root LV and container
+	runtime metadata LV have been allocated/grown.
+
+	DATA_SIZE can take values acceptable to "lvcreate -L" as well as
+	some values acceptable to "lvcreate -l". If user intends to pass
+	values acceptable to "lvcreate -l", then only those values which
+	contains "%" in syntax are acceptable.  If value does not contain
+	"%" it is assumed value is suitable for "lvcreate -L".
+
+MIN_DATA_SIZE: Specifies the minimum size of the data volume. If
```
rhvgoyal: same here. Will `thin pool data LV` be better instead of `data volume` ?

docker-storage-setup.sh
```diff
@@ -17,7 +17,7 @@
 #++
 
 #  Purpose:  This script grows the root filesystem and sets up LVM volumes
-#            for docker metadata and data.
+#            for container runtime metadata and data.
```
rhvgoyal: Can we change it. We have deprecated old mode of setting up metadata and data volumes. So we probably can say `sets up storage for container runtime`

docker-storage-setup.sh
```diff
@@ -28,22 +28,22 @@ set -e
...
 # The options below should be specified as values acceptable to 'lvextend -L':
 #
 # ROOT_SIZE: The size to which the root filesystem should be grown.
 #
-# DATA_SIZE: The desired size for the docker data LV.  Defaults to using all
-#            free space in the VG after the root LV and docker metadata LV
-#            have been allocated/grown.
+# DATA_SIZE: The desired size for the container runtime data LV.
+#            Defaults to using all free space in the VG after the root LV and
+#            container runtime metadata LV have been allocated/grown.
```
rhvgoyal: I think this information needs to be updated. First of all now it is effectively size of thin pool. Also it defaults to using 40% of FREE space and not 100% of FREE space.

docker-storage-setup.sh
rhvgoyal: I think this is confusing. Given we have already verified that both CONTAINER_ROOT_MOUNT_DIR_PATH and CONTAINER_ROOT_LV_NAME have been passed in, why not make just a single check.

`[ -z $RESOLVED_MOUNT_DIR_PATH ] && return 0`

rhvgoyal: Ideally this change should be in a separate patch as it has nothing to do with renaming of variables. Even if you leave it here, I am fine.

docker-storage-setup.1
rhatdan: Fixed

docker-storage-setup.1
rhatdan: Fixed

docker-storage-setup.1
rhatdan: Fixed

docker-storage-setup.sh
rhatdan: Fixed.

docker-storage-setup.sh
rhatdan: Should we just refer to the man page for an explanation of the fields, rather then having to maintain them in two places?

docker-storage-setup.sh
rhatdan: Fixed

docker-storage-setup.sh
rhvgoyal: makes sense. if we keep information at multiple places, it becomes harder to keep it up to date.

tests/006-test-overlay-cleanup.sh
```diff
@@ -23,9 +23,11 @@ EOF
...
     # Test failed.
     test_status=1
+    echo $DSSBIN --reset "Failed" >> $LOGS 2>&1
  elif [ -e /etc/sysconfig/docker-storage ]; then
     # Test failed.
     test_status=1
+    echo "$DSSBIN --reset Failed: /etc/sysconfig/docker-storage still exist" >> $LOGS 2>&1
  fi
   if [ ${test_status} -eq 1 ]; then
     echo "ERROR: $testname: $DSSBIN --reset Failed." >> $LOGS
```
rhvgoyal: Given you have added specific message for errors. may be lets change this message to just inform that test failed. Anyways this message is little misleading now.

`ERROR: $testname Failed`

tests/007-test-setting-extra-opts.sh
```diff
@@ -45,7 +45,11 @@ EOF
   echo $DOCKER_STORAGE_OPTIONS | grep -q -- "$extra_options"
   
   # Successful appending to DOCKER_STORAGE_OPTIONS
-  [ $? -eq 0 ] && test_status=0
+  if [ $? -eq 0 ]; then
+      test_status=0
+  else
+      echo "ERROR: $testname: $extra_options != $DOCKER_STORAGE_OPTIONS." >> $LOGS
```
rhvgoyal: We are doing grep. So DOCKER_STORAGE_OPTIONS should contain extra_options and does not have to be same.

Can we change error message to reflect that. Maybe.

` ERROR: $testname: DOCKER_STORAGE_OPTIONS does not contain extra_options`

docker-storage-setup.sh
```diff
@@ -1169,6 +1169,11 @@ if [ -e /etc/sysconfig/docker-storage-setup ]; then
   source /etc/sysconfig/docker-storage-setup
 fi
 
+if [ ! -z "$EXTRA_DOCKER_STORAGE_OPTIONS" ]; then
+    Info "EXTRA_DOCKER_STORAGE_OPTIONS is deprecated, please use EXTRA_STORAGE_OPTIONS"
+    EXTRA_STORAGE_OPTIONS=${EXTRA_DOCKER_STORAGE_OPTIONS}
+fi
```
rhvgoyal: We also need to have a check to make sure user can't specify both old and new options. Otherwise with current code, old option will overwrite new option.

We already have a function called check_storage_options() and I think these checks can go in there?

tests/006-test-overlay-cleanup.sh
```diff
@@ -23,11 +23,9 @@ EOF
  if [ $? -ne 0 ]; then
     # Test failed.
     test_status=1
-    echo $DSSBIN --reset "Failed" >> $LOGS 2>&1
  elif [ -e /etc/sysconfig/docker-storage ]; then
     # Test failed.
     test_status=1
-    echo "$DSSBIN --reset Failed: /etc/sysconfig/docker-storage still exist" >> $LOGS 2>&1
```
rhvgoyal: You added these messages in previous patch and now you got rid of these?

tests/006-test-overlay-cleanup.sh
```diff
@@ -37,7 +35,7 @@ EOF
  return $test_status
 }
 
-# Create a overlay backend and then make sure the
```
rhvgoyal: This change also seems a revert from previous patch. What's going on.

tests/007-test-setting-extra-opts.sh
```diff
@@ -45,11 +45,7 @@ EOF
   echo $DOCKER_STORAGE_OPTIONS | grep -q -- "$extra_options"
   
   # Successful appending to DOCKER_STORAGE_OPTIONS
-  if [ $? -eq 0 ]; then
-      test_status=0
-  else
-      echo "ERROR: $testname: $extra_options != $DOCKER_STORAGE_OPTIONS." >> $LOGS
-  fi
```
rhvgoyal: This is a revert too.

tests/008-test-overlay2-setup-cleanup.sh
```diff
@@ -41,7 +41,7 @@ EOF
  return $test_status
 }
 
-# Create a overlay2 backend and then make sure the
```
rhvgoyal: revert from previous patch.

tests/run-tests.sh
```diff
@@ -22,11 +22,11 @@ check_docker_active() {
...
-  local devmapper_meta_dir="$METADATA_DIR/devicemapper/metadata/"
+  local docker_devmapper_meta_dir="$DOCKER_METADATA_DIR/devicemapper/metadata/"
   
-  [ ! -d "$devmapper_meta_dir" ] && return 0
+  [ ! -d "$docker_devmapper_meta_dir" ] && return 0
 
-  echo "ERROR: ${METADATA_DIR} directory exists and contains old metadata. Remove it." >&2
+  echo "ERROR: /var/lib/docker directory exists and contains old metadata. Remove it." >&2
   exit 1
 }
```
rhvgoyal: This whole change is revert from previous patch.

docker-storage-setup.sh
rhatdan: Fixed

rhvgoyal:
Hi Dan,

We will also need tests for new option EXTRA_STORAGE_OPTIONS. What's the plan w.r.t adding new automated tests.

docker-storage-setup.1
```diff
@@ -98,67 +100,34 @@ WIPE_SIGNATURES:
...
-      is assumed value is suitable for "lvcreate -L".
-
-      Note: If both STORAGE_DRIVER=devicemapper and
-      DOCKER_ROOT_VOLUME=yes is set, docker-storage-setup would set
-      up the thin pool for devicemapper first, followed by docker 
-      root volume. e.g if free space in the volume group is 10G, 
-      devicemapper thin pool size would be 4G (40% of 10G) and docker
-      root volume would be 2.4G (40% of 6G).
-
 CONTAINER_ROOT_LV_NAME:
```
rhvgoyal: We are getting rid of DOCKER_ROOT_VOLUME_SIZE. That means we need to output an Info message saying this is deprecated. I don't see that in code yet.

rhvgoyal: We probably also need to add a Note in man page under CONTAINER_ROOT_LV_SIZE saying DOCKER_ROOT_LV_SIZE is deprecated.

rhatdan: Fixed

rhatdan:
Extra tests will come with the next pull requests when we add INPUTFILE and OUTPUTFILE.

Fixed tests scripts to have new behaviour.

docker-storage-setup.conf
```diff
@@ -27,7 +28,7 @@ STORAGE_DRIVER=devicemapper
 #
 # ROOT_SIZE=8G
 
-# The desired size for the docker data LV.  It defaults using 60% of all
+# The desired size for the container runtime data LV.  Defaults using 60% of all
```
rhvgoyal: Defaults to 40% of FREE space.

docker-storage-setup.1
rhvgoyal: Can we move these checks into check_storage_options(). Much better to have all the checks together in a function.

rhvgoyal: Also we need to deprecated `DOCKER_ROOT_VOLUME_SIZE`. I don't know where did `DOCKER_ROOT_LV_SIZE` come from.

docker-storage-setup.sh
```diff
@@ -1169,6 +1169,18 @@ if [ -e /etc/sysconfig/docker-storage-setup ]; then
   source /etc/sysconfig/docker-storage-setup
 fi
 
+if [ ! -z "$DOCKER_ROOT_LV_SIZE" ]; then
+    Info " DOCKER_ROOT_LV_SIZE is deprecated, please use CONTAINER_ROOT_LV_SIZE"
```
rhvgoyal: This should be `DOCKER_ROOT_VOLUME_SIZE`. There is no variable called `DOCKER_ROOT_LV_SIZE`.

Also can we move all these checks in check_storage_options().

Also this check does not really belong in this patch. Given you moved DOCKER_ROOT_VOLUME_SIZE in docker-storage-setup.sh in patch3, that's where it should be.

docker-storage-setup.sh
```diff
@@ -1032,6 +1008,19 @@ check_storage_options(){
   if [ "$DOCKER_ROOT_VOLUME" == "yes" ];then
      Info "DOCKER_ROOT_VOLUME is deprecated, and will be removed soon. Use CONTAINER_ROOT_LV_MOUNT_PATH instead."
   fi
+
+  if [ ! -z "$DOCKER_ROOT_VOLUME_SIZE" ]; then
+      Info " DOCKER_ROOT_VOLUME_SIZE is deprecated, please use CONTAINER_ROOT_LV_SIZE"
+  fi
+
```
shishir-a412ed: Why not use `-n` instead of ` ! -z` ? That would be more inline with the existing checks and less characters.

docker-storage-setup.sh
```diff
@@ -1032,6 +1008,19 @@ check_storage_options(){
...
+  if [ ! -z "$EXTRA_DOCKER_STORAGE_OPTIONS" ]; then
```
shishir-a412ed: Use `-n`. Same as above.

docker-storage-setup.sh
```diff
@@ -1032,6 +1008,19 @@ check_storage_options(){
...
+      Info "EXTRA_DOCKER_STORAGE_OPTIONS is deprecated, please use EXTRA_STORAGE_OPTIONS"
+      if [ ! -z "$EXTRA_STORAGE_OPTIONS" ]; then
```
shishir-a412ed: Use `-n`. Same as above.

docker-storage-setup.1
```diff
@@ -98,67 +100,34 @@ WIPE_SIGNATURES:
...
-     no new logical volume will be created. e.g. Specifying
-     CONTAINER_ROOT_LV_MOUNT_PATH=/var/lib/containers will carve
-     out a logical volume, create a filesystem on it and
-     mount it on /var/lib/containers.
+     Creates a logical volume named CONTAINER_ROOT_LV_NAME and mounts
+     it at the specified path. By default no new logical volume will
+     be created. For example:
+     \f[B]CONTAINER_ROOT_LV_MOUNT_PATH=/var/lib/containers\f[]
+     would carve out a logical volume, format it with an XFS filesystem
+     and mount it on /var/lib/containers.
```
shishir-a412ed: Small nit :) There is an extra space after `/var/lib/containers` that needs to be removed.

rhatdan: fixed

docker-storage-setup.sh
rhatdan: fixed

docker-storage-setup.sh
```diff
@@ -16,53 +16,25 @@
...
-# Note: lvm2 version should be same or higher than lvm2-2.02.112 for lvm
-#       thin pool functionality to work properly.
+# Read man page for a description of currently supported options:
+# 'man docker-storage-setup'
+
 POOL_LV_NAME="docker-pool"
 DOCKER_ROOT_LV_NAME="docker-root-lv"
 DOCKER_ROOT_DIR="/var/lib/docker"
+DOCKER_METADATA_DIR="/var/lib/docker"
+DOCKER_ROOT_VOLUME_SIZE=40%FREE
```
shishir-a412ed: Why are we setting `DOCKER_ROOT_VOLUME_SIZE=40%FREE` in the code? This should be set by the user.

docker-storage-setup.sh
rhatdan: fixed

docker-storage-setup.sh
rhatdan: fixed

docker-storage-setup.sh
rhatdan: This was being set originally by /usr/lib/docker-storage-setup/docker-storage-setup.conf, but we have removed it from their, since we don't want references to "docker" in container-storage-setup configuration.  We hard code it here so that on upgrade systems will continue to work

shishir-a412ed: The problem is everytime I run `docker-storage-setup` or `docker-storage-setup --reset`, I see this message 
`INFO:  DOCKER_ROOT_VOLUME_SIZE is deprecated, please use CONTAINER_ROOT_LV_SIZE`.

Even if I am not using these options. e.g. `cat /etc/sysconfig/docker-storage-setup`.
````
DEVS=/dev/vdb
VG=atomicos
WIPE_SIGNATURES=true
`````
Just trying to setup a thin pool under `VG=atomicos` and I see this message.
We did not had a deprecation message for `DOCKER_ROOT_VOLUME_SIZE` before since it was not set by the user. Now if we want a INFO message, we should print it when it is set by the user as opposed to set up the code.

File: docker-storage-setup.sh (hunk 34 above)
Comment (shishir-a412ed) [inline docker-storage-setup.sh]: Have you pushed the changes to github? I am not able to see the changes. It is still showing ` ! -z`

File: docker-storage-setup.sh (hunk 38 above)
Comment (rhatdan) [inline docker-storage-setup.sh]: Ok we should remove that check, since we are hard coding it to be there.

Comment (rhatdan) [inline docker-storage-setup.sh]: fixed

**Comment by rhatdan** [comment]
@shishir-a412ed @rhvgoyal All changes are in now.

File: docker-storage-setup.sh (hunk 38 above)
Comment (rhvgoyal) [inline docker-storage-setup.sh]: So you removed this check from check_storage_options()? In which patch? looks like it is still present in third patch of the series.

**Comment by rhvgoyal** [comment]
I am having a hard time now. 

- First we override CONTAINER_ROOT_LV_SIZE with DOCKER_ROOT_LV_SIZE.
- Then we introduced deprecation check for DOCKER_ROOT_LV_SIZE
- Then we fixed it to use DOCKER_ROOT_VOLUME_SIZE but removed the code which assigned default value to CONTAINER_ROOT_LV_SIZE
- Then we removed DOCKER_ROOT_VOLUME_SIZE check altogether.

**Comment by rhatdan** [comment]
We are hard coding DOCKER_ROOT_VOLUME_SIZE into the .sh, so the check will always fire.

**Comment by rhatdan** [comment]
Then we fixed it to use DOCKER_ROOT_VOLUME_SIZE but removed the code which assigned default value to CONTAINER_ROOT_LV_SIZE

I can do this if you want.  But we will not be able to tell the difference between the two.

**Comment by shishir-a412ed** [comment]
@rhvgoyal There was no deprecation check for `DOCKER_ROOT_VOLUME_SIZE` in the master. Only deprecation check for `DOCKER_ROOT_VOLUME`.
@rhatdan introduced that check (deprecation message for `DOCKER_ROOT_VOLUME_SIZE`) which I pointed out that it would fire up everytime since we are:

a) hard-coding `DOCKER_ROOT_VOLUME_SIZE=40%FREE` in `docker-storage-setup.sh`. We are doing this to maintain backward compatibility since `DOCKER_ROOT_VOLUME_SIZE` is removed from `/usr/lib/docker-storage-setup/docker-storage-setup`.

So Dan removed that check. No harm done, since we never had that check anyways. I have tested the current set of patches, and they work well with both `DOCKER_ROOT_VOLUME*` options and `CONTAINER_ROOT_LV*` options and provides full backward compatibility.

**Comment by rhvgoyal** [comment]
Ok, for now, let us not give any warning for DOCKER_ROOT_VOLUME_SIZE.

What's bothering me now is that in commit 8da47a60  we are introducing the check and in commit 77b35dd3b we are getting rid of check. If we dont' want this check, let us not introduce the check to begin with in earlier commit.

**Comment by rhvgoyal** [comment]
Similarly, in commit 8da47a6036, we are adding following.

[ -z "$RESOLVED_MOUNT_DIR_PATH" ] && [ -z "$CONTAINER_ROOT_LV_NAME" ] && return 0

And then modifying it again in commit 6f781edf

-  [ -z "$RESOLVED_MOUNT_DIR_PATH" ] && [ -z "$CONTAINER_ROOT_LV_NAME" ] && return 0
+  [ -z "$RESOLVED_MOUNT_DIR_PATH" ] && return 0


This change should only happen in last patch and commit 8da47a6036 should not be touching it. That patch is not related to this change at all.

**Comment by rhatdan** [comment]
Cleaned both of those up.

File: tests/006-test-overlay-cleanup.sh [hunk 39]
```diff
@@ -22,20 +22,19 @@ EOF
  $DSSBIN --reset >> $LOGS 2>&1
  if [ $? -ne 0 ]; then
     # Test failed.
+    echo "ERROR: $testname: $DSSBIN --reset Failed." >> $LOGS
     test_status=1
  elif [ -e /etc/sysconfig/docker-storage ]; then
     # Test failed.
+    echo "ERROR: $testname: $DSSBIN --reset Failed to create /etc/sysconfig/docker." >> $LOGS
```
rhvgoyal: This should be `cleanup` instead of `create` ? 

I wished more cleanups we had taken in another PR. More we add in single PR, more comments and rework it generates.

rhvgoyal:
@rhatdan Apart from that single comment, this series looks fine to me. Please make that change, and I will provide my ack and merge it.

rhatdan:
Fixed.

tests/006-test-overlay-cleanup.sh
rhatdan: fixed

rhvgoyal:
LGTM

rhvgoyal:
@rh-atomic-bot r+

rhvgoyal:
@rh-atomic-bot r+ 2dc19ad

PR #183, opened 2017-01-20
REMOVE ALL DOCKER_ references from man page and default config

rhatdan:
This is part of the effort to convert docker-storage-setup to
container-storage-setup

First we want to remove all references to DOCKER in environment variables.

### DIFF
--- a/docker-storage-setup.1
+++ b/docker-storage-setup.1
@@ -1,7 +1,7 @@
 .TH "DOCKER-STORAGE-SETUP" "1" "NOVEMBER 2014" "Helper Script for Docker Storage Setup" ""
 .SH NAME
 .PP
-docker\-storage\-setup - Grows the root filesystem and sets up storage for docker.
+docker\-storage\-setup - Grows the root filesystem and sets up storage for container runtimes.
 .SH SYNOPSIS
 .PP
 \f[B]docker-storage-setup\f[] [OPTIONS]
@@ -11,11 +11,13 @@ docker\-storage\-setup - Grows the root filesystem and sets up storage for docke
   Print usage statement
 
 \f[B]--reset\f[]
-  Reset your docker storage to init state. Reset does not try to remove volume groups or try to remove any of the disks added previously.
-  
-Note: The
-\f[B]--reset\f[]
-command is not sufficient to cleanup your docker environment.  This option is used by other tools. (\f[B]atomic storage reset\f[])
+  Reset your container storage to init state. Reset does not remove
+  volume groups or remove any of the disks added previously.
+
+  Note: The \f[B]--reset\f[]
+command is not always sufficient to cleanup your
+  container runtime environment. Other tools (\f[B]atomic storage reset\f[])
+  use this command to cleanup all storage.
 
 .SH EXAMPLES
 Run \f[B]docker-storage-setup\f[] after setting up your configuration in
@@ -31,14 +33,16 @@ functionality to work properly.
 \f[B]Supported options for the configuration file\f[]:
 
 STORAGE_DRIVER:
-      Specify a storage driver to be used with docker. Default
-      driver is "devicemapper". Other valid values are overlay,
-      overlay2 and "". Empty string tells docker-storage-setup to
-      not do any storage setup.
+      Specify a storage driver to be used with container runtime.
+      Default: "devicemapper".
+      Valid values are overlay, overlay2 and "".
+      "" tells docker-storage-setup to not perform any storage setup.
 
-EXTRA_DOCKER_STORAGE_OPTIONS:
-      A set of extra options that should be passed to the Docker
-      daemon.
+EXTRA_STORAGE_OPTIONS:
+      A set of extra options that should be passed to the container
+      runtime daemon.
+      Note: EXTRA_STORAGE_OPTIONS replaces EXTRA_DOCKER_STORAGE_OPTIONS
+      which has been deprecated
 
 DEVS: A quoted, space-separated list of devices to be used.
       If a drive is partitioned and contains a ${dev}1 partition,
@@ -47,14 +51,13 @@ DEVS: A quoted, space-separated list of devices to be used.
       is not specified, then use of the root disk's extra space
       is implied.
 
-VG:   The volume group to use for docker storage.  Defaults to the
+VG:   The volume group to use for container storage.  Defaults to the
       volume group where the root filesystem resides.  If VG is
       specified and the volume group does not exist, it will be
       created (which requires that "DEVS" be nonempty, since we don't
       currently support putting a second partition on the root disk).
 
-      lvm2 version should be same or higher than lvm2-2.02.112 for lvm
-      thin pool functionality to work properly.
+Note: lvm2 needs to be lvm2-2.02.112 or later for lvm thin pool functionality to work properly.
 
 GROWPART:
       One can use this option to enable/disable growing of partition
@@ -79,16 +82,15 @@ POOL_AUTOEXTEND_PERCENT:
       pool size.
 
 CHUNK_SIZE:
-      Controls the chunk size/block size of thin pool. Value of CHUNK_SIZE
-      be suitable to be passed to --chunk-size option of lvconvert.
+      Controls the chunk size/block size of thin pool. CHUNK_SIZE value
+      must be suitable for passing to \f[B]lvconvert --chunk-size\f[].
 
 DEVICE_WAIT_TIMEOUT:
-           Specifies a device wait timeout value in seconds. In
-           certain cases required devices might not be immediately
-           available and docker-storage-setup might decide to wait for
-           it. This timeout specifies how long one should wait for the
-           device. Default is 60 seconds. One can disable the wait by
-           specifying a value of 0.
+      Specifies a device wait timeout value in seconds. In certain
+      cases required devices might not be immediately available and
+      docker-storage-setup might decide to wait for it. This timeout
+      specifies how long one should wait for the device.
+      Default is 60 seconds. 0 disables wait.
 
 WIPE_SIGNATURES:
       Wipe any signatures found on disk. Valid values are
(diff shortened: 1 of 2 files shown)


docker-storage-setup.1
```diff
@@ -31,14 +31,11 @@ functionality to work properly.
 \f[B]Supported options for the configuration file\f[]:
 
 STORAGE_DRIVER:
-      Specify a storage driver to be used with docker. Default
-      driver is "devicemapper". Other valid values are overlay,
-      overlay2 and "". Empty string tells docker-storage-setup to
-      not do any storage setup.
+      Specify a storage driver to be used with container runtime. The default driver is "devicemapper". Valid values are overlay, overlay2 and "". Empty string tells docker-storage-setup to not perform any storage setup.
```
rhvgoyal: Do you like this new format. I mean when I do man, start of description line is aligned with the option this text is explaining.

I liked previous format better where all text was shifted to right.

docker-storage-setup.1
```diff
@@ -31,14 +31,11 @@ functionality to work properly.
...
 
-EXTRA_DOCKER_STORAGE_OPTIONS:
-      A set of extra options that should be passed to the Docker
-      daemon.
+EXTRA_STORAGE_OPTIONS:
+      A set of extra options that should be passed to the container runtime daemon.
+      NOTE: This option replaces EXTRA_DOCKER_STORAGE_OPTIONS which has been deprecated
```
rhvgoyal: Can we modify this along with a patch where you actually deprecate EXTRA_STORAGE_OPTIONS. May be we can have one patch for each option. In fact it would be good if we have a patch series where each patch deprecates one option. Easy to review and merge.

docker-storage-setup.1
```diff
@@ -47,14 +44,13 @@ DEVS: A quoted, space-separated list of devices to be used.
...
-VG:   The volume group to use for docker storage.  Defaults to the
+VG:   The volume group to use for container storage.  Defaults to the
       volume group where the root filesystem resides.  If VG is
       specified and the volume group does not exist, it will be
       created (which requires that "DEVS" be nonempty, since we don't
       currently support putting a second partition on the root disk).
 
-      lvm2 version should be same or higher than lvm2-2.02.112 for lvm
-      thin pool functionality to work properly.
+NOTE: lvm2 needs to be lvm2-2.02.112 or later for lvm thin pool functionality to work properly.
```
rhvgoyal: I told @shishir-a412ed to use `Note` instead of `NOTE`. If `NOTE` is better, then we will to change that to make man page uniform.

docker-storage-setup.1
```diff
@@ -79,16 +75,16 @@ POOL_AUTOEXTEND_PERCENT:
       pool size.
 
 CHUNK_SIZE:
-      Controls the chunk size/block size of thin pool. Value of CHUNK_SIZE
-      be suitable to be passed to --chunk-size option of lvconvert.
+      Controls the chunk size/block size of thin pool. Value of
+      CHUNK_SIZE be suitable to be passed to --chunk-size option
```
rhvgoyal: `should be suitable` ?

docker-storage-setup.1
```diff
@@ -98,42 +94,42 @@ WIPE_SIGNATURES:
...
-      Note: A user cannot specify DOCKER_ROOT_VOLUME and
-      CONTAINER_ROOT_LV_MOUNT_PATH together. They are mutually exclusive
-      options.
+      Note: DOCKER_ROOT_VOLUME is deprecated.
+      Specifying DOCKER_ROOT_VOLUME and CONTAINER_ROOT_LV_MOUNT_PATH
+      is not allowed.
 
-DOCKER_ROOT_VOLUME_SIZE:
-      Specify the desired size for docker root volume. It defaults
+      Specify the desired size for container root volume. It defaults
```
rhvgoyal: You forgot to add `CONTAINER_ROOT_LV_SIZE` ?

docker-storage-setup.1
```diff
@@ -98,42 +94,42 @@ WIPE_SIGNATURES:
       overwritten as suitable. This also means that if there is any
       data on disk, it will be lost.
 
-DOCKER_ROOT_VOLUME:
+CONTAINER_ROOT_LV_MOUNT_PATH:
```
rhvgoyal: CONTAINER_ROOT_LV_MOUNT_PATH is already there. I guess you should modify that section.

docker-storage-setup.1
```diff
@@ -172,30 +168,28 @@ The options below should be specified as values acceptable to 'lvextend -L':
...
-           "bBsSkKmMgGtTpPeE" are valid suffixes. If no suffix is
-           specified then value will be considered as megabyte unit.
-
-           Both upper and lower case suffix represent same unit of
-           size. Use suffix B for Bytes, S for sectors as 512 bytes, K
-           for kibibytes (1024 bytes), M for mebibytes
-           (1024 kibibytes), G for gibibytes, T for tebibytes, P for
-           pebibytes and E for exbibytes.
+DATA_SIZE: The desired size for the container runtime data LV.
+	Defaults: 40% free space in the VG after the root LV and container
```
rhvgoyal: can we do something about `root LV` and emphasize that it refers to system rootfs and not container root lv.

docker-storage-setup.1
```diff
@@ -98,42 +94,42 @@ WIPE_SIGNATURES:
...
+      Specifying DOCKER_ROOT_VOLUME and CONTAINER_ROOT_LV_MOUNT_PATH
+      is not allowed.
 
-DOCKER_ROOT_VOLUME_SIZE:
-      Specify the desired size for docker root volume. It defaults
+      Specify the desired size for container root volume. It defaults
       to 40% of all free space.
 
-      DOCKER_ROOT_VOLUME_SIZE can take values acceptable to
+      CONTAINER_ROOT_LV_SIZE can take values acceptable to
```
rhvgoyal: Now we have two entries for CONTAINER_ROOT_LV_SIZE. One need to go away.

docker-storage-setup.1
rhatdan: I am trying to cleanup the format.  I had lots of problems.  Take another look at the format now.

docker-storage-setup.1
rhatdan: Sure I just wanted to get agreement on the man page and config then I can add them back to the other pull request.  I just saw this as a lot of noice and wanted to take care of these first.

docker-storage-setup.1
rhatdan: I switched all NOTE->Note

docker-storage-setup.conf
```diff
@@ -1,20 +1,21 @@
...
+# Default is devicemapper.
 # Other possible options are overlay, overlay2 and "". Empty string means do
 # not do any storage setup.
 STORAGE_DRIVER=devicemapper
 
-# A set of extra options that should be appended to the generated
-# DOCKER_STORAGE_OPTIONS string. These options will be passed to the Docker
-# daemon as-is and should be valid Docker storage options.
-# EXTRA_DOCKER_STORAGE_OPTIONS="--storage-opt dm.fs=ext4"
+# Set extra options that will be appended to the generated STORAGE_OPTIONS
```
rhvgoyal: This will continue to be DOCKER_STORAGE_OPTIONS for a while till we figure out a way how to make output string generic as well as useful. May be user will have to pass in a prefix so that we could have DOCKER_STORAGE_OPTIONS in output?

Or always output STORAGE_OPTIONS and modify docker to accept that.

Anyway, this should be modified when associated code has been modified too. Possibly a separate patch for this.

docker-storage-setup.conf
```diff
@@ -84,43 +85,30 @@ DEVICE_WAIT_TIMEOUT=60
 # as this means if there was any leftover data on disk, it will be lost.
 WIPE_SIGNATURES=false
 
-# By default no new volume and filesystem will be setup for docker
-# root dir. Docker creates /var/lib/docker/ on top of underlying filesystem
-# for storing images and containers.
+# By default no new volume and filesystem will be setup for container runtime
+# root dir. For example docker creates /var/lib/docker/ on top of underlying
+# filesystem for storing images and containers.
 #
```
rhvgoyal: can we get rid of this extra line. Otherwise it feels description of new option has started. Right now convention is that new line is there only if we moved onto description of next option.

docker-storage-setup.conf
```diff
@@ -84,43 +85,30 @@ DEVICE_WAIT_TIMEOUT=60
...
-#
-# docker-storage-setup creates a logical volume with an XFS filesystem mounted
-# on docker root directory (default: /var/lib/docker).
-#
-# Note: DOCKER_ROOT_VOLUME is deprecated, and will be removed
-# soon. Use CONTAINER_ROOT_LV_MOUNT_PATH instead.
-#
-# Note: A user cannot specify DOCKER_ROOT_VOLUME and CONTAINER_ROOT_LV_MOUNT_PATH
-# together. They are mutually exclusive options.
-DOCKER_ROOT_VOLUME=no
```
rhvgoyal: Actually I was talking of this blank line and not the previous one.

docker-storage-setup.conf
rhvgoyal: ignore this comment. I was talking of the blank line below and not this one.

docker-storage-setup.conf
rhatdan: These are just flags in the unit file.  We will just have to modify the UNIT file to take both STORAGE_OPTIONS and DOCKER_STORAGE_OPTIONS.  container-storage-setup will only generate STORAGE_OPTIONS in the furture.

docker-storage-setup.1
rhatdan: I think we would need to go over this together.  There are a ton of words in this man page and default config already,  not sure you will ever be able to fully express what you mean in a manner people would understand.

rhatdan:
I think I have all of your comments.

docker-storage-setup.conf
rhvgoyal: ok, can we make this change when you actually introduce option EXTRA_STORAGE_OPTIONS. Right now we are modifying man page without actually code supporting what man page says.

docker-storage-setup.1
```diff
@@ -98,56 +100,33 @@ WIPE_SIGNATURES:
...
-     out a logical volume, create a filesystem on it and
-     mount it on /var/lib/containers.
+     Creates a logical volume named CONTAINER_ROOT_LV_NAME and mounts
+     it at the specified path. By default no new logical volume will
+     be created. For example:
+     \f[B]CONTAINER_ROOT_LV_MOUNT_PATH=/var/lib/containers\f[]
+     would carve out a logical volume, format it with an XFS filesystem
+     and mount it on /var/lib/containers. 
+
+     By default no new volume and filesystem will be setup for
```
rhvgoyal: We are repeating `By default no new volume will be created`. We said tht same thing in previous paragraph too.

I think next two paragraphs are not needed at all. May be keep it short and sweet. Just retain the Note section.

I think first paragraph covers it just fine?

docker-storage-setup.conf
rhvgoyal: We are using term `generated STROAGE_OPTIONS variable`. And code does not do right now what this text is saying. Code will generate DOCKER_STORAGE_OPTIONS

docker-storage-setup.conf
```diff
@@ -84,43 +85,29 @@ DEVICE_WAIT_TIMEOUT=60
...
+# Setting CONTAINER_ROOT_LV_MOUNT_PATH requires that CONTAINER_ROOT_LV_NAME be set
+# CONTAINER_ROOT_LV_NAME="container-root-lv"
 
-# Specify the desired size for docker root volume. It defaults to 40% of all
-# free space.
+# Specify the desired size for container root lv volume. It defaults to 40% of
+# all free space.
 #
-# DOCKER_ROOT_VOLUME_SIZE can take values acceptable to "lvcreate -L" as well
+# CONTAINER_ROOT_LV_SIZE can take values acceptable to "lvcreate -L" as well
```
rhvgoyal: Check your final docker-storage-setup.conf file. There is double description of CONTAINER_ROOT_LV_SIZE.

Are you using some script to make these changes? :-)

docker-storage-setup.conf
rhatdan: Yes once you are happy with the man page, I will put this patch into the other patch set and close this down,  Then we can make sure docker-storage-setup actually matches the man page.

We will not merge this patch until Code is generating STORAGE_OPTIONS.

rhatdan:
Moved this back into the original patch set.

https://github.com/projectatomic/docker-storage-setup/pull/180

PR #184, opened 2017-01-20
build: Do make install before running tests

rhvgoyal:
Do make install to make sure all config files have been installed at
proper place before tests run. Otherwise new PR can source old files
and not the new one and that can break tests.

Given new docker-storage-setup.sh will be installed in /usr/bin/, there
is no need to specify path of binary using environment variable. So
get rid of that.

### DIFF
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -12,12 +12,11 @@ context: fedora/25/atomic
 
 required: true
 
-env:
-  DOCKER_STORAGE_SETUP: ./docker-storage-setup.sh
-
 tests:
   - systemctl stop docker
   - rm -rf /var/lib/docker
+  - ostree admin unlock
+  - make install
   - rm -f /etc/sysconfig/docker-storage-setup
   - rm -f /etc/sysconfig/docker-storage
   - if findmnt /dev/vdb; then umount /dev/vdb; fi


rhvgoyal:
@jlebon PTAL cc @shishir-a412ed

rhvgoyal:
@jlebon it does not kick in any tests automatically. In the past it used to. Do I need to start tests manually?

rhvgoyal:
@rh-atomic-bot try

jlebon:
bot, retest this please

jlebon:
bot, retest this please

jlebon:
bot, retest this please

jlebon:
bot, retest this please

.redhat-ci.yml
```diff
@@ -13,11 +13,12 @@ context: fedora/25/atomic
 required: true
 
 env:
```
jlebon: You can also remove the `env:` key.

.redhat-ci.yml
```diff
@@ -13,11 +13,12 @@ context: fedora/25/atomic
...
-  DOCKER_STORAGE_SETUP: ./docker-storage-setup.sh
 
 tests:
   - systemctl stop docker
   - rm -rf /var/lib/docker
+  - ostree admin unlock
+  - make install
```
jlebon: I'm surprised this works. Didn't realize `make` was in Fedora AH. The way this workflow is done usually is to `make install DESTDIR=/host` from inside a build container. But since you don't actually need to compile anything and `make` is available... why not.

rhvgoyal:
@jlebon looks like parted is missing on atomic host? jenkins test passed but fedora/25/atomic one is failing.

In fact this should be new. Because on your PR same test passed. And my PR does not seem to do anything which seems to take parted away.

jlebon:
@rhvgoyal Indeed, there is no `parted` on Fedora 25 AH. That said, I don't think that's why the test is failing. This line:

```
tests/libtest.sh: line 34: parted: command not found
```

is present in both these logs and the logs from the other PR. It's from the cleanup procedure in `libtest.sh`. (In fact, even successful tests have that line).

The actual error happens before that it seems:

```
Running test tests/004-test-non-absolute-disk-name-support.sh
ERROR: Device /dev/vdb is already partitioned and cannot be added to volume group dss-test-foo
  Volume group "dss-test-foo" not found
  Cannot process volume group dss-test-foo
  No PV label found on /dev/vdb1.
  WARNING: D-Bus notification failed: The name com.redhat.lvmdbus1 was not provided by any .service files
```

which happened because the ioctl from the wipefs of the previous test failed:

```
/dev/vdb: calling ioctl to re-read partition table: Device or resource busy
```

So something is still holding the device busy. Seems like it's a racy test & cleanup procedure. Might help to retry the wipefs a few times before giving up.

rhvgoyal:
@jlebon BTW, we will need to install lvm2-dbusd package to get rid of all the following warnings.

WARNING: D-Bus notification failed: The name com.redhat.lvmdbus1 was not provided by any .service files

rhvgoyal:
@jlebon this race gets visible if parted is missing. I removed parted on my system and in the very first run I can see the race.

I suspect that some udev rule is executing. When parted is used to remove partition, then by the time wipefs executes, udev rule probably has finished.

But when parted is missing, then wipefs must triggering some notificaiton to udev while deleting partition signature. And that probably is causing race. 

IOW, race does not happen if parted is present on the system. We need parted on atomic host. In fact there was a card where we wanted to move to parted instead of sfdisk for docker-storage-setup.

So is it possible to start including parted in atomic host builds.

cgwalters:
@rhvgoyal File an issue at https://pagure.io/fedora-atomic

jlebon:
@rhvgoyal Can't we use `fdisk` instead here?

rhvgoyal:
@jlebon I am about to push a patch to use sfdisk if parted is not present on the system. That should get us going for now. In the long term, I think it is easier to use parted from script and it probably is good idea to ship parted with atomic host.

rhvgoyal:
https://github.com/projectatomic/docker-storage-setup/pull/185

rhvgoyal:
@cgwalters created an issue to consider shipping parted.

https://pagure.io/fedora-atomic/issue/50

rhvgoyal:
@cgwalters also opened an issue to install lvm2-dbusd to get rid of following warnings with every lvm2 command.

WARNING: D-Bus notification failed: The name com.redhat.lvmdbus1 was not provided by any .service files

rhvgoyal:
@jlebon hmm..., looks like even using sfdisk did not resolve the issue. It did resolve issue on my f25 VM though.

cgwalters:
By "F25" do you mean a yum-managed not Atomic Host system?

cgwalters:
Atomic Host is *also* Fedora so it helps to disambiguate.

rhvgoyal:
yes, dnf manage system and not atomic host. I probably should install f25 atomic-host vm and test.

cgwalters:
That seems like a good idea indeed.

rhvgoyal:
ok, I installed f25 atomic-host vm locally and I can't reproduce the issue there. All thests pass locally. And that's true both with using sfdisk and without using sfdisk.

So looks like this race condition is being hit in this test environment more easily for some reason. 

One thing I noticed that there are lvm2 warnings in my host. May be it is somehow related.

rhvgoyal:
@rh-atomic-bot retry

cgwalters:
If something isn't queued for a merge, it's Jenkins, not homu, so:

bot, retest this please

rhvgoyal:
bot, retest this please

rhvgoyal:
@cgwalters looks like retest worked for jenkins. but how to re-run fedora/25/atomic tests (without re-pushing)

rhvgoyal:
@rh-atomic-bot r+

rhvgoyal:
@jlebon Now this seems to be stuck in merging. another homu issue?

jlebon:
@rh-atomic-bot r+ 0286afc

PR #185, opened 2017-01-23
tests: Use sfdisk if parted is not present

rhvgoyal:
Right now atomic host does not seem to include parted. In that case use
sfdisk to remove partitions.

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -26,12 +26,30 @@ remove_pvs() {
   done
 }
 
+parted_del_partition() {
+  local dev=$1
+  parted ${dev} rm 1 >> $LOGS 2>&1
+}
+
+sfdisk_del_partition() {
+  local dev=$1
+  sfdisk --delete ${dev} 1 >> $LOGS 2>&1
+}
+
 remove_partitions() {
   local dev devs=$1
+  local use_parted=false
 
-  # Assume partition number 1 is to be removed.
+  if [ -x "/usr/sbin/parted" ]; then
+    use_parted=true
+  fi
+
   for dev in $devs; do
-    parted ${dev} rm 1 >> $LOGS 2>&1
+    if [ "$use_parted" == "true" ]; then
+      parted_del_partition "$dev"
+    else
+      sfdisk_del_partition "$dev"
+    fi
   done
 }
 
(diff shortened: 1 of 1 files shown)


rhvgoyal:
cc @jlebon

jlebon:
Strange indeed. I can't reproduce this when running on OpenStack just like the tester.
Could you try adding a `udevadm settle` in case it is indeed udev?

jlebon:
Ahh, I did get it to reproduce, but on another test. Will check if `udevadm settle` makes it more reliable (otherwise, a simple retry loop should do the trick).

jlebon:
This works for me:

```diff
diff --git a/tests/libtest.sh b/tests/libtest.sh
index bd407ae..cf55ce4 100644
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -56,6 +56,11 @@ remove_partitions() {
 # Wipe all signatures on devices
 wipe_signatures() {
   local dev devs=$1
+
+  # wipefs triggers a partition table reload,
+  # so let's make sure we're stable first
+  udevadm settle
+
   for dev in $devs; do
     wipefs -a $dev >> $LOGS 2>&1
   done
```

Would be nice to turn on `set -e` in these scripts. It's a bit tricky to debug otherwise.

rhvgoyal:
@jlebon, ok, using udevadm settble between remove_partition and wipefs seems reasonable. 

Or we could do.

- delete partition
- udevadm settle
- partprobe <disk>

But this again requires partprobe and that requires parted package which is missing on atomic host. 

So for now, it makes sense to let udev settle down before running wipefs. I will update my patch and test it and put your signed-off-by as well.

jlebon:
@rh-atomic-bot r+ 084a374

rhvgoyal:
@jlebon, I am not sure if this commit fixed the problem. I was looking at logs and got the feeling that wipefs was still finding the device busy.

rhvgoyal:
@jlebon is homu stuck? Generally it does not take this long to merge a PR.

jlebon:
@rhvgoyal It's not configured properly for this repo. Will fix.

jlebon:
@rh-atomic-bot retry

jlebon:
@rhvgoyal I think those "Device or resource busy" messages are not coming from `wipefs` but maybe e.g. deleting partitions. The `wipefs` calls themselves seem clean:

```
/dev/vdb: 2 bytes were erased at offset 0x000001fe (dos): 55 aa
/dev/vdb: calling ioctl to re-read partition table: Success
```

rhvgoyal:
@jlebon Right. if we have waited for udev to settle down after partition deletion, wipefs reload partition table operation should be success.

PR #191, opened 2017-01-26
Use parted (If available) to partition disk, else fallback to sfdisk.

shishir-a412ed:

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -706,9 +706,8 @@ scan_disks() {
   echo $new_disks
 }
 
-create_partition() {
+create_partition_sfdisk(){
   local dev="$1" size
-
   # Use a single partition of a whole device
   # TODO:
   #   * Consider gpt, or unpartitioned volumes
@@ -720,6 +719,21 @@ unit: sectors
 
 ${dev}1 : start=     2048, size=  ${size}, Id=8e
 EOF
+}
+
+create_partition_parted(){
+  local dev="$1"
+  parted $dev --script mklabel msdos mkpart primary 0% 100% set 1 lvm on
+}
+
+create_partition() {
+  local dev="$1"
+
+  if [ -x "/usr/sbin/parted" ]; then
+    create_partition_parted $dev
+  else
+    create_partition_sfdisk $dev
+  fi
 
   # Sometimes on slow storage it takes a while for partition device to
   # become available. Wait for device node to show up.


shishir-a412ed:
ping @rhvgoyal

docker-storage-setup.sh
```diff
@@ -706,9 +706,8 @@ scan_disks() {
   echo $new_disks
 }
 
-create_partition() {
-  local dev="$1" size
-
+create_partition_sfdisk(){
+local dev="$1" size
```
rhvgoyal: no spaces before local.

docker-storage-setup.sh
```diff
@@ -720,6 +719,21 @@ unit: sectors
 
 ${dev}1 : start=     2048, size=  ${size}, Id=8e
 EOF
+}
+
+create_partition_parted(){
+local dev="$1"
+parted $dev --script mklabel gpt mkpart primary 0% 100%
```
rhvgoyal: add 2 spaces before start of line.

rhvgoyal: sfdisk creates dos type partition table by default. I think we should do same for parted. At some point of time we want to switch to GPT but this is not the patch to do so. 

We might experience surprises when we move to GPT. So use `mklabel dos`

shishir-a412ed:
@rhvgoyal PTAL.

docker-storage-setup.sh
```diff
@@ -720,6 +719,21 @@ unit: sectors
 
 ${dev}1 : start=     2048, size=  ${size}, Id=8e
 EOF
+}
+
+create_partition_parted(){
+  local dev="$1"
+  parted $dev --script mklabel dos mkpart primary 0% 100%
```
rhvgoyal: it should be `mklabel msdos` and not `mklabel dos`

rhvgoyal: Where is it documented that parted accepts start and end value in %.  Can you point me to it

rhvgoyal: May be we should use `0` as start value which means start at 0 MB and use `-1s` as end value to tell parted to create partition till end of disk.

I think it will still do the right alignment and create partition  beginning at 2048 sector.

shishir-a412ed: I was following your comment for `mklabel dos` :)
https://github.com/projectatomic/docker-storage-setup/pull/191#discussion_r98042388

I found the `mkpart primary 0% 100%` syntax on stack overflow. I tested it and it works. We can also do 0 and -1, that should work too.

rhvgoyal: I see in man page that one can specify unit and % is one of the supported units. so may be parted is parsing 0% and 100% properly.

shishir-a412ed: I have changed `mklabel` from `dos` to `msdos` based on this documentation:
https://www.gnu.org/software/parted/manual/html_node/mklabel.html

docker-storage-setup.sh
```diff
@@ -720,6 +719,21 @@ unit: sectors
 
 ${dev}1 : start=     2048, size=  ${size}, Id=8e
 EOF
+}
+
+create_partition_parted(){
+  local dev="$1"
+  parted $dev --script mklabel msdos mkpart primary 0% 100%
```
rhvgoyal: sfdisk sets this partition type to lvm. For the sake of keeping it same,, lets do the same in parted too. Something like.

`parted $dev --script mklabel msdos mkpart primary 0% 100% set 1 lvm`

rhvgoyal: I forgot the keyword `on` at the end. So it should be.

arted $dev --script mklabel msdos mkpart primary 0% 100% set 1 lvm on

rhvgoyal: `parted $dev --script mklabel msdos mkpart primary 0% 100% set 1 lvm on`

rhvgoyal:
LGTM

rhvgoyal:
@rh-atomic-bot r+ 2da4cfa

PR #192, opened 2017-01-26
Add versioning to docker-storage-setup.

shishir-a412ed:

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -1148,6 +1148,11 @@ reset_storage() {
   rm -f ${STORAGE_OUT_FILE}
 }
 
+print_version(){
+  local version="1.0.0"
+  echo "$version"
+}
+
 usage() {
   cat <<-FOE
     Usage: $1 [OPTIONS] INPUTFILE OUTPUTFILE
@@ -1155,7 +1160,8 @@ usage() {
     Grows the root filesystem and sets up storage for container runtimes
 
     Options:
-      --help    Print help message
+      --help    Print help message.
+      --version Print version.
       --reset   Reset your docker storage to init state. 
 FOE
 }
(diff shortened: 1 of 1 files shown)


shishir-a412ed:
ping @rhvgoyal

rhvgoyal:
@shishir-a412ed can you update this PR please.

docker-storage-setup.sh
```diff
@@ -1167,6 +1167,10 @@ reset_storage() {
   rm -f $DOCKER_STORAGE
 }
 
+print_version() {
+  echo "1.0.0"
```
rhvgoyal: Lets store version information in a variable and echo that variable.

shishir-a412ed:
@rhvgoyal PTAL.

rhvgoyal:
@shishir-a412ed So how about following.

MAJOR_VERSION=0
MINOR_VERSION=1
SUBLEVEL=0
EXTRA_VERSION=""

And final version will be
VERSION=${MAJOR_VERSION}.${MINOR_VERSION}.${SUBLEVEL}
[ -n "$EXTRA_VERSION" ] && VERSION="${VERSION}-$EXTRA_VERSION"

That is initially our version will look like.

0.1.0

Once we have something stable we will bump it to.

1.0.0

And after that during development, we can add extra version like -dev, or -rc1 etc.

1.0.0-dev
1.0.0-rc1

and then release a new version say
1.1.0

@rhatdan WDYT about above versioning scheme for docker-storage-setup.

rhatdan:
I am fine with this. Not crazy about the -dev -rc1 stuff since this can cause rpm package versioning to get screwed up.

In Rawhide if we shipped these, the rc1 would take precedence over the -dev.  Which might not be what you expect.  We could just bounce the sublevel.  But this is what docker does, and we end up truncating off the -dev field.

container-storage-setup-1.0.0-dev

contianer-storage-setup-1.0.0-rc1

rhvgoyal:
I think -dev and -rc1 will be used for keeping track of status upstream and ideally rpms should not be packaging it. 

Once we are ready to release something for packaging, then we can get rid of -dev and -rc1 and packagers can take that version and package it.

rhvgoyal:
@shishir-a412ed can you update this PR. We probably need global variables as I describe in my comments.

rhatdan:
We have gotten this from another pull request.

PR #195, opened 2017-02-02
Add options to docker-storage-setup to allow other container runtimes to use it 

rhatdan:
Basically this patch set adds an optional inputfile and outputfile to docker-storage-setup to allow users to overide /etc/sysconfig/docker-storage-setup and /etc/sysconfig/docker-storage.

Mainly tests to check if this will work.

Traditional checks will make sure docker still works, new checks will verify options passed in will work.


rhatdan:
@rhvgoyal @shishir-a412ed PTAL

This now works and passes the tests.

shishir-a412ed:
@rhatdan I think there is too much duplication of code. Every existing test is copied into a new test to handle new `$INFILE` and `$OUTFILE` feature. Why can't we use the existing test code, and just make two calls to it ?

e.g. Let's take an example of `006-test-overlay-cleanup.sh`.
We can have two calls in the end
a) test_reset_overlay 
b) test_reset_overlay $INFILE $OUTFILE

If `$INFILE` and `$OUTFILE` are passed, we use the passed in values. Otherwise we can set them to default `$INFILE: /etc/sysconfig/docker-storage-setup` and `$OUTFILE: /etc/sysconfig/docker-storage`.

WDYT?

Shishir

rhatdan:
We could do that, but our original thought was to keep these separate, the first tests be "docker" specific and thrown away if we eventually remove the docker centric code.

docker-storage-setup.1
```diff
@@ -1,10 +1,23 @@
 .TH "DOCKER-STORAGE-SETUP" "1" "NOVEMBER 2014" "Helper Script for Docker Storage Setup" ""
 .SH NAME
 .PP
-docker\-storage\-setup - Grows the root filesystem and sets up storage for container runtimes.
+docker\-storage\-setup - Tool for setting up storage for container runtimes.
 .SH SYNOPSIS
 .PP
 \f[B]docker-storage-setup\f[] [OPTIONS]
+\f[B]docker-storage-setup\f[] [OPTIONS] INPUTFILE OUTPUTFILE
```
rhvgoyal: As of now INPUTFILE and OUTPUTFILE are optional so require brackets around these?

docker-storage-setup.1
```diff
@@ -1,10 +1,23 @@
 .TH "DOCKER-STORAGE-SETUP" "1" "NOVEMBER 2014" "Helper Script for Docker Storage Setup" ""
 .SH NAME
 .PP
-docker\-storage\-setup - Grows the root filesystem and sets up storage for container runtimes.
+docker\-storage\-setup - Tool for setting up storage for container runtimes.
 .SH SYNOPSIS
 .PP
 \f[B]docker-storage-setup\f[] [OPTIONS]
```
rhvgoyal: This line needs to go away?

tests/libtest.sh
```diff
@@ -21,8 +21,8 @@ lv_exists() {
 }
 
 clean_config_files() {
-  rm -f /etc/sysconfig/docker-storage-setup
-  rm -f /etc/sysconfig/docker-storage
+  rm -f $1
+  rm -f $2
```
rhvgoyal: Is this function really useful anymore as a library funciton?

Given file names are being passed in by callers, it might be simpler if callers do "rm -f INPUTFILE OUTPUTFILE"

rhvgoyal:
It would be nice to differente between two kind of tests. Tests which are for docker only and are there for making sure that backward compatibility with docker is not broken.

Also new tests which are for generic container runtime setup. Right now its not obvious which test is which (except numbering part).

Will it make sense to prefix all old test with docker. say `001-docker-test-use-devs-to-create-thin-pool.sh`. That way, it will be clear which are docker backward compatibility tests.

rhvgoyal:
We can also probably add a section `NOTES` in tests/README and mention that all tests between 0 and 100 are there for compatibility with docker. Anything beyond 100, are generic tests for container runtime storage.

tests/101-test-use-devs-to-create-thin-pool.sh
```diff
@@ -0,0 +1,45 @@
+source $SRCDIR/libtest.sh
+
+# Test DEVS= directive. Returns 0 on success and 1 on failure.
+test_devs() {
+  local devs=$TEST_DEVS
+  local test_status=1
+  local testname=`basename "$0"`
+  local vg_name="dss-test-foo"
```
rhvgoyal: These probably should be `css-test-foo`. dss is short form for docker-storage-setup.

tests/101-test-use-devs-to-create-thin-pool.sh
```diff
@@ -0,0 +1,45 @@
...
+  local INFILE=storage-setup
+  local OUTFILE=storage
```
rhvgoyal: So this will create files in current working directory of the test? We already create a directory called temp/ in cwd. How about putting these files there?

That way these will not spoil the code dir if I do tests/run-tests.sh

rhvgoyal: Also how about calling these `container-storage-setup` and `container-storage` respectively.

tests/101-test-use-devs-to-create-thin-pool.sh
```diff
@@ -0,0 +1,45 @@
...
+    echo "ERROR: $testname: $DSSBIN $INFILE $OUTFILE Failed." >> $LOGS
+    cleanup $vg_name "$devs" "$INFILE" "$OUTFILE"
+    return $test_status
+ fi
+
+  # Make sure volume group $VG got created
+  if vg_exists "$vg_name"; then
+    test_status=0
+  else
+    echo "ERROR: $testname: $DSSBIN $INFILE $OUTFILE Failed $vg_name was not created." >> $LOGS
```
rhvgoyal: Should `Failed` be `failed` and then next sentence starts.

rhvgoyal: rather `failed.`

tests/run-tests.sh
```diff
@@ -153,5 +157,5 @@ check_config_files
 setup_workdir
 setup_dss_binary
 check_block_devs "$DEVS"
-run_tests
+run_tests $@
```
rhvgoyal: This is a good idea. Can we push this patch as a separate PR. This is useful even without rest of the stuff.

docker-storage-setup.1
rhatdan: No I think the convention is to leave both uses of the command.  You can run it with no files in the first use case and in the second you need both.

docker-storage-setup.1
rhatdan: No I want it as a legitimate way to run the command.

docker-storage-setup.1
rhatdan: The problem with making them both optional is they are not.  IE If I specify INPUTFILE, you  have to specify OUTPUTFILE.

tests/libtest.sh
rhatdan: Sure I can remove it and replace all calls with that.

tests/101-test-use-devs-to-create-thin-pool.sh
rhatdan: We have not changed the name yet. I was figuring to wait on those until we change the name.

tests/run-tests.sh
rhatdan: https://github.com/projectatomic/docker-storage-setup/pull/196

tests/101-test-use-devs-to-create-thin-pool.sh
rhatdan: Ok I am lower casing all Failed in tests.

tests/101-test-use-devs-to-create-thin-pool.sh
rhatdan: Fixed

tests/101-test-use-devs-to-create-thin-pool.sh
rhvgoyal: Also need a full stop after failed.

tests/101-test-use-devs-to-create-thin-pool.sh
rhvgoyal: ok

docker-storage-setup.1
rhvgoyal: Ok, then we need a new line between two. Otherwise this is how man page shows it.

`docker-storage-setup [OPTIONS] docker-storage-setup [OPTIONS] INPUTFILE OUTPUTFILE`

tests/run-tests.sh
shishir-a412ed: Since we are opening a separate PR #196 for this, I think we should remove this commit from this PR.

tests/libtest.sh
rhatdan: Fixed

tests/run-tests.sh
rhatdan: Sure merge the other one, and I will remove it from here.

docker-storage-setup.sh
```diff
@@ -35,6 +35,39 @@ STORAGE_IN_FILE="/etc/sysconfig/docker-storage-setup"
...
+
+    Options:
+      --help    Print help message
+      --reset   Reset your docker storage to init state. 
+FOE
+}
+
+OPTS=`getopt -o h -l reset -l help -- "$@"`
+eval set -- "$OPTS"
+reset=0
```
rhvgoyal: All global variables are in capital letters in this script. lets continue to follow that convention.

docker-storage-setup.sh
```diff
@@ -35,6 +35,39 @@ STORAGE_IN_FILE="/etc/sysconfig/docker-storage-setup"
 STORAGE_OUT_FILE="/etc/sysconfig/docker-storage"
 STORAGE_DRIVERS="devicemapper overlay overlay2"
 
+usage() {
```
rhvgoyal: Shouldn't all this code go after comment `Main Script`. I mean that's where most the executable code starts. Things above that are supposed to be helper functions and variables.

rhvgoyal: Looks like that there is code above Main script too. I probably did not pay much attention. Can we fix it.

docker-storage-setup.sh
rhatdan: Fixed

docker-storage-setup.sh
rhatdan: We would need to refactor the code into a series of functions.  The problem now is that everything is not a function.

We need these global variables set before they are used.

tests/libtest.sh
```diff
@@ -86,7 +81,7 @@ cleanup() {
   # cases it has been observed that udev rule kept the device
   # busy.
   udevadm settle
-  clean_config_files $INFILE $OUTFILE
+  rm -f $INFILE $OUTFILE
```
rhvgoyal: This is library  and I feel it is better to pass in names of infile and outfile to clean instead of relying on global variables $INFILE and $OUTFILE.

rhvgoyal:
So you have dropped new tests from this PR?

tests/README
```diff
@@ -7,6 +7,12 @@ Configure a disk for test
   This tells test infrastructure to use disk /dev/sdb for tests. This disk
   should be unused.
 
+NOTES:
+All tests between 0 and 100 are there for compatibility with docker. These
+tests cover backwards compatability.  Going forward we would prefer docker as
+well as other container runtimes specify the input and ouptut files.  Anything
+beyond 100, are generic tests for container runtime storage.
+
```
rhatdan: Fixed

rhvgoyal:
I think they need to be part of this PR so that new code can be tested.

rhvgoyal:
Oh, they are there. I missed that patch.

tests/libtest.sh
```diff
@@ -71,6 +71,13 @@ wipe_signatures() {
 cleanup() {
   local vg_name=$1
   local devs=$2
+  local INFILE=/etc/sysconfig/docker-storage-setup
+  local OUTFILE=/etc/sysconfig/docker-storage
```
rhvgoyal: INFILE and OUTFILE are local variables and should be small cap. That's the convention we are following.

tests/libtest.sh
rhatdan: @rhvgoyal it is passed in.

rhvgoyal: ok, these uppercase variable names confused me. Again, can we keep lowercase for local variables and uppercase for gloabl variables.

tests/libtest.sh
rhatdan: Fixed

docker-storage-setup.sh
```diff
@@ -1200,6 +1175,32 @@ if [ -e /usr/lib/docker-storage-setup/docker-storage-setup ]; then
   source /usr/lib/docker-storage-setup/docker-storage-setup
 fi
 
+# Main Script
+OPTS=`getopt -o h -l reset -l help -- "$@"`
```
rhvgoyal: This patch is supposed to be adding tests. All this code was part of separate patch. Looks like you mixed up something.

rhatdan: Fixed

tests/006-test-overlay-cleanup.sh
```diff
@@ -4,8 +4,10 @@ source $SRCDIR/libtest.sh
 test_reset_overlay() {
   local test_status=0
   local testname=`basename "$0"`
+  local INFILE=/etc/sysconfig/docker-storage-setup
+  local OUTFILE=/etc/sysconfig/docker-storage
```
rhvgoyal: lowercase for local variables.

tests/libtest.sh
rhvgoyal: this code has come back?

tests/libtest.sh
```diff
@@ -79,7 +86,7 @@ cleanup() {
   # cases it has been observed that udev rule kept the device
   # busy.
   udevadm settle
-  clean_config_files
+  clean_config_files $INFILE $OUTFILE
```
rhvgoyal: I think you have lost some changes and gone back to old code. So it needs to be redone.

tests/libtest.sh
rhatdan: Not in the patches I just pushed.  All changes for infile and outfile lower case.

rhvgoyal: We had got rid of clean_config_file() function. It has come back. what's going on.

rhvgoyal:
Oh, you created additional patch on top to remove clean_config_files().

rhvgoyal:
@rhatdan checks are failing.

tests/101-test-use-devs-to-create-thin-pool.sh
```diff
@@ -0,0 +1,45 @@
+source $SRCDIR/libtest.sh
+
+# Test DEVS= directive. Returns 0 on success and 1 on failure.
+test_devs() {
+  local devs=$TEST_DEVS
+  local test_status=1
+  local testname=`basename "$0"`
+  local vg_name="dss-test-foo"
+  local infile=${WORKDIR}/docker-storage-setup
+  local outfile=storage
```
rhvgoyal: These have changed again. We had agreed upon container-storage-setup and container-storage respectively and you had made changes. You have lost those changes.

rhvgoyal: `WORKDIR` is not exported. Does this really work?

tests/103-test-override-signature-wipes-existing-signatures.sh
```diff
@@ -0,0 +1,52 @@
...
+    echo "ERROR: $testname: $DSSBIN failed." >> $LOGS
+    cleanup $vg_name "$devs" "$infile" "$outfile"
+    return $test_status
+  fi
+
+  # Make sure volume group $VG got created.
+  if vg_exists "$vg_name"; then
+    test_status=0
+  else
+    echo "ERROR: $testname: $DSSBIN $infile $outfile failed $vg_name was not created." >> $LOGS
```
rhvgoyal: `.` after failed.

tests/104-test-non-absolute-disk-name-support.sh
```diff
@@ -0,0 +1,51 @@
...
+    echo "ERROR: $testname: $DSSBIN failed." >> $LOGS
+    cleanup $vg_name "$TEST_DEVS" "$infile" "$outfile"
+    return $test_status
+  fi
+
+  # Make sure volume group $VG got created.
+  if vg_exists "$vg_name"; then
+    test_status=0
+  else
+    echo "ERROR: $testname: $DSSBIN $infile $outfile failed $vg_name was not created." >> $LOGS
```
rhvgoyal: `.` after failed.

tests/105-test-devmapper-cleanup.sh
```diff
@@ -0,0 +1,53 @@
...
+    return $test_status
+ fi
+
+  $DSSBIN --reset $infile $outfile >> $LOGS 2>&1
+  # Test failed.
+  if [ $? -eq 0 ]; then
+     if [ ! -e $outfile ]; then
+	  test_status=0
+     else
+	  echo "ERROR: $testname: $DSSBIN --reset $infile $outfile failed $outfile still exists." >> $LOGS
```
rhvgoyal: ditto

tests/106-test-overlay-cleanup.sh
```diff
@@ -0,0 +1,42 @@
...
+ fi
+
+ $DSSBIN --reset $infile $outfile >> $LOGS 2>&1
+ if [ $? -ne 0 ]; then
+    # Test failed.
+    echo "ERROR: $testname: $DSSBIN --reset failed." >> $LOGS
+    test_status=1
+ elif [ -e $outfile ]; then
+    # Test failed.
+    echo "ERROR: $testname: $DSSBIN --reset $infile $outfile failed to remove $outfile." >> $LOGS
```
rhvgoyal: ditto

tests/108-test-overlay2-setup-cleanup.sh
```diff
@@ -0,0 +1,48 @@
...
+
+ $DSSBIN --reset $infile $outfile >> $LOGS 2>&1
+ if [ $? -ne 0 ]; then
+    # Test failed.
+    test_status=1
+    echo "ERROR: $testname: $DSSBIN --reset $infile $outfile failed." >> $LOGS
+ elif [ -e $outfile ]; then
+    # Test failed.
+    test_status=1
+    echo "ERROR: $testname: $DSSBIN --reset $infile $outfile failed $outfile still exists." >> $LOGS
```
rhvgoyal: ditto. Please change rest of the instances too, if any.

tests/101-test-use-devs-to-create-thin-pool.sh
rhatdan: Yes, I believe so. Children inherit their parents environment.

```
# !/bin/sh
WORKDIR="dan"
sh -c "echo $WORKDIR"
```

tests/103-test-override-signature-wipes-existing-signatures.sh
rhatdan: fixed.  I seem to have brought back some older tests.

rhvgoyal:
Thanks. LGTM

rhvgoyal:
@rh-atomic-bot r+ 546ef5a

rhvgoyal:
This breaks start of docker. We have modified docker-storage-setup.service to make use of inputfile and outputfile but did not modify default /etc/sysconfig/docker-storage-setup to have CONTAINER_THINPOOL=docker-pool.

That means docker-storage-setup fails by default after doing "make install".

rhatdan:
Are you commenting on the wrong pull request?

rhvgoyal:
Nope. Just telling that with this PR, we should have also modified docker-storage-setup-override.conf to set CONTAINER_THINPOOL=docker-pool.

Or we should not have modified docker-storage-setup.service. 

Because after this PR, if you do "make install" and try to start docker-storage-setup.service, it fails saying CONTAINER_THINPOOL needs to be defined.

IOW, this PR broke the tree.

PR #197, opened 2017-02-08
Rename docker-storage-setup script to container-storage-setup

rhatdan:
This is a follow on to the existing patch set, to actually do the rename.  Just want to run it through CI to see if it has any issues.  

Lets concentrate on getting the initial patches merged before looking at this one.


rhatdan:
@rhvgoyal This should be the final one...

container-storage-setup.sh
```diff
@@ -23,7 +23,7 @@ set -e
 
 # This section reads the config file $INPUTFILE
 # Read man page for a description of currently supported options:
-# 'man docker-storage-setup'
+# 'man container-storage-setup'
 
 POOL_LV_NAME="docker-pool"
```
rhvgoyal: This probably should be renamed to container-thin-pool? But this change will be tricky and we will have to make sure that it is not broken over upgrade. So this should kick in only for new thin pool set up.

Also for the case of default docker,  we should continue to use name docker-pool.

rhvgoyal: In fact this change probably should be the first one and should be done in a separate PR and then we can look at rest of the changes.

rhatdan: https://github.com/projectatomic/docker-storage-setup/pull/198

rhvgoyal:
@rhatdan can you please update this.

rhatdan:
I was waiting for the other patch to be merged. https://github.com/projectatomic/docker-storage-setup/pull/198

rhatdan:
@rhvgoyal PTAL

rhatdan:
@shishir-a412ed PTAL

rhvgoyal:
Name of the patch and commit description is wrong. It belongs to other PR which is already merged.

rhatdan:
Fixed commit message.

tests/001-test-use-devs-to-create-thin-pool.sh
```diff
@@ -5,9 +5,9 @@ test_devs() {
   local devs=$TEST_DEVS
   local test_status=1
   local testname=`basename "$0"`
-  local vg_name="dss-test-foo"
+  local vg_name="css-test-foo"
```
rhvgoyal: I see that there are two kind of changes in this PR. One which are necessary and the one which are not. For example, this one is not necessary. Can we take all un-necessary changes and put in a separate PR and push that one first.

.redhat-ci.yml
```diff
@@ -21,7 +21,7 @@ tests:
   - rm -f /etc/sysconfig/docker-storage
   - if findmnt /dev/vdb; then umount /dev/vdb; fi
   - wipefs -a /dev/vdb
-  - echo "DEVS=/dev/vdb" >> tests/dss-test-config
+  - echo "DEVS=/dev/vdb" >> tests/css-test-config
```
rhvgoyal: This can be part of first PR.

Makefile
```diff
@@ -1,25 +1,29 @@
...
 .PHONY: clean
 clean:
 	-rm -rf *~ \#* .#*
 
 .PHONY: install
-install:
-
-	install -D -m 755 ${DOCKER}-storage-setup.sh ${BINDIR}/${DOCKER}-storage-setup
+install-docker:
 	install -D -m 644 ${DOCKER}-storage-setup.service ${UNITDIR}/${DOCKER}-storage-setup.service
```
rhvgoyal: I had not noticed that we install unit file as well. Does it override the unit file installed by docker package on this system?

Makefile
```diff
@@ -1,25 +1,29 @@
...
 
 .PHONY: clean
 clean:
 	-rm -rf *~ \#* .#*
 
 .PHONY: install
-install:
-
-	install -D -m 755 ${DOCKER}-storage-setup.sh ${BINDIR}/${DOCKER}-storage-setup
+install-docker:
```
rhvgoyal: What is this install_docker target and why is it required.

Makefile
```diff
@@ -1,25 +1,29 @@
...
-	install -D -m 755 dss-child-read-write.sh ${DSSLIBDIR}/dss-child-read-write
-	install -D -m 644 ${DOCKER}-storage-setup.1 ${MANDIR}/man1/${DOCKER}-storage-setup.1
+
+install:
+	install -D -m 755 ${CONTAINER}-storage-setup.sh ${BINDIR}/${CONTAINER}-storage-setup
+	(cd ${BINDIR}; ln -sf ${CONTAINER}-storage-setup ${DOCKER}-storage-setup)
+	install -D -m 644 ${CONTAINER}-storage-setup.conf ${CSSLIBDIR}/${CONTAINER}-storage-setup
+	install -D -m 755 libcss.sh ${CSSLIBDIR}/libcss.sh
+	install -D -m 755 css-child-read-write.sh ${CSSLIBDIR}/css-child-read-write
+	install -D -m 644 ${CONTAINER}-storage-setup.1 ${MANDIR}/man1/${CONTAINER}-storage-setup.1
```
rhvgoyal: Should we install a man page for docker-storage-setup which says that dokcer-storage-setup has been renamed to container-storage-setup and do `man container-storage-setup` instead.

docker-storage-setup-override.conf
```diff
@@ -1,4 +0,0 @@
-# Edit this file to override any configuration options specified in
-# /usr/lib/docker-storage-setup/docker-storage-setup.
-#
-# For more details refer to "man docker-storage-setup"
```
rhvgoyal: So this file will now move in docker and ocid packages respectively?

docker-storage-setup.service
```diff
@@ -5,7 +5,7 @@ Before=docker.service
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/docker-storage-setup /etc/sysconfig/docker-storage-setup /etc/sysconfig/docker-storage
+ExecStart=/usr/bin/container-storage-setup /etc/sysconfig/docker-storage-setup /etc/sysconfig/docker-storage
 EnvironmentFile=-/etc/sysconfig/docker-storage-setup
```
rhvgoyal: Why do we need to retain this file now? Even if we retain, should we really install it during "make install" time? I feel that installing this unit file feels little odd.

rhvgoyal: Should we stop installing this file and call this container-storage-setup.service just for an example?

tests/run-tests.sh
```diff
@@ -48,7 +48,7 @@ check_config_files() {
   fi
 }
 
-setup_dss_binary() {
+setup_css_binary() {
   # One can setup environment variable DOCKER_STORAGE_SETUP to override
   # which binary is used for tests.
   if [ -n "$DOCKER_STORAGE_SETUP" ];then
```
rhvgoyal: So you are retaining this for backward compatibility of running tests? I guess not a bad idea. We should probably also parse CONTAINER_STORAGE_SETUP and ask users to set that instead.

Or we could start emiting warning that DOCKER_STORAGE_SETUP environment variable has been deprecated. Use CONTAINER_STORAGE_SETUP instead.

tests/113-test-container-root-volume-cleanup.sh
```diff
@@ -1,20 +1,20 @@
 source $SRCDIR/libtest.sh
 
-# Test "docker-storage-setup --reset" for CONTAINER_ROOT_LV_NAME=
```
rhvgoyal: Can we keep all the changes related to tests either in a separate PR or in a separate patch. I wan to make sure core changes are part of a separate commit.

tests/112-test-use-devs-to-create-container-root-volume.sh
```diff
@@ -27,12 +27,12 @@ CONTAINER_ROOT_LV_MOUNT_PATH=$root_lv_mount_path
...
 EOF
 
- # Run docker-storage-setup
- $DSSBIN $infile $outfile >> $LOGS 2>&1
+ # Run container-storage-setup
+ $CSSBIN $infile $outfile >> $LOGS 2>&1
 
  # Test failed.
  if [ $? -ne 0 ]; then
-    echo "ERROR: $testname: $DSSBIN failed." >> $LOGS
```
rhvgoyal: All this naming of DSSBIN to CSSBIN can really be part of a separate PR in preparation to renaming to container-storage-setup. This will reduce size of PR.

tests/112-test-use-devs-to-create-container-root-volume.sh
```diff
@@ -6,13 +6,13 @@ test_container_root_volume() {
   local devs=$TEST_DEVS
   local test_status=1
   local testname=`basename "$0"`
-  local vg_name="dss-test-foo"
+  local vg_name="css-test-foo"
```
rhvgoyal: renaming of vg_name to css-test-foo can be in that other PR for preparation of renaming to container-storage-setup

Makefile
rhatdan: No, I don't believe so.  I believe @lsm5 is either not using the Makefile or is overriding the docker.service file elsewhere.  If he is using it he will need to move it docker.spec.

Makefile
rhatdan: I just left it because we have an example of how docker.service would use container-storage-setup.

Makefile
rhvgoyal: I mean "make install" will override docker.service already installed on the system? That does not sound right, does it? "make install" should not get into business of installing systemd unit files I think and it should be left to rpm packaging?

tests/001-test-use-devs-to-create-thin-pool.sh
rhatdan: So you want me to just change file names in first pass and then internal references in second/third patch?

Makefile
rhatdan: I meant to install a man page for docker-storage-setup that just brings up container-storage-setup.

docker-storage-setup-override.conf
rhatdan: Yes that is the idea.  But we would just ship /etc/sysconfig/docker-storage-setup and /etc/sysconfig/ocid-storage-setup. Perhaps with a /usr/share/ocid-storage-setup file available for users to play with.

docker-storage-setup.service
rhatdan: I don't think we should call it container-storage-setup.service.  I would rather remove it all together or use it as an example of how you could use a config file.

tests/113-test-container-root-volume-cleanup.sh
rhatdan: Well if you want to tests the patches separately the tests will fail.

Makefile
rhvgoyal: That's fine too.

tests/112-test-use-devs-to-create-container-root-volume.sh
rhatdan: Sure, if that is what you want.  I still think we should change all files on disk in one patch and then change internal names as a separate patches.

rhvgoyal:
Ok, I think this PR is too big. Lets us try to break it down in to smaller PRs and keep actual renaming as the last PR.

I think first step is that most of the changes related to tests/ dir can be done first in a separate PR and that should bring down the size of the patch. Then we can look again and see what other piece can be committed first.

rhvgoyal:
@rhatdan can you please resolve conflicts and push again.

rhatdan:
@rhvgoyal PTAL tests are passing.  Although we probably should merge the make install patch first.

rhvgoyal:
@rhatdan can you please rebase this on top of latest master.

rhvgoyal:
make install gives following message.

"install: cannot stat 'docker-storage-setup-override.conf': No such file or directory"

I see you have renamed  docker-storage-setup-override.conf to container-storage-setup-override.conf and that's the issue.

If you are not expecting to install a default /etc/sysconfig/container-storage-setup, why not leave it as it is and not rename this file.

Makefile
```diff
@@ -19,12 +20,14 @@ install-docker: install-core
...
 
 .PHONY: install-core
 install-core:
-	install -D -m 755 ${DOCKER}-storage-setup.sh ${BINDIR}/${DOCKER}-storage-setup
-	install -D -m 644 ${DOCKER}-storage-setup.conf ${DSSLIBDIR}/${DOCKER}-storage-setup
-	install -D -m 755 libdss.sh ${DSSLIBDIR}/libdss.sh
-	install -D -m 755 dss-child-read-write.sh ${DSSLIBDIR}/dss-child-read-write
-	install -D -m 644 ${DOCKER}-storage-setup.1 ${MANDIR}/man1/${DOCKER}-storage-setup.1
+	install -D -m 755 ${CONTAINER}-storage-setup.sh ${BINDIR}/${CONTAINER}-storage-setup
+	(cd ${BINDIR}; ln -sf ${CONTAINER}-storage-setup ${DOCKER}-storage-setup)
```
rhvgoyal: Should we do this soft link creation logic as part of install-docker target. This is required for backward compatibility and all that stuff we seem to be putting in install-docker target?

That means of docker rpm will take care of creating this soft link and that seem to make sense.

rhvgoyal: Also you are not installing new man page for docker-storage-setup. That should probably go in install-docker target as well?

rhatdan: No, I was thinking more of the upgrade case.  Where container-storage-setup gets updated but docker package does not.  These are basically aliases of the old package, while the other content is core to the docker package.

rhvgoyal: That sounds like a real corner case where docker package is old while container-storage-setup is newer.

.travis.yml
```diff
@@ -8,7 +8,7 @@ addons:
       - autotools-dev
 script:
   # For now, just a syntax check.
-  - bash -n docker-storage-setup.sh
+  - bash -n container-storage-setup.sh
```
rhvgoyal: I am curious, who uses this .travis.yml file? Some CI somewhere?

README.md
```diff
@@ -1,27 +1,27 @@
...
 [Docker Engine Storage docs](https://docs.docker.com/engine/userguide/storagedriver/)
 
-`docker-storage-setup` is a script to configure the devicemapper or
+`container-storage-setup` is a script to configure the devicemapper or
 overlayfs backends, part of the
 [Project Atomic](http://www.projectatomic.io/) suite of container
 projects.  It is run via a systemd service
-`docker-storage-setup.service` before the Docker daemon, ensuring
-storage is provisioned according to configuration in
+`container-storage-setup.service` before the Container runtime daemons, ensuring
```
rhvgoyal: There is no file container-storage-setup.service

container-storage-setup-override.conf
```diff
@@ -0,0 +1,4 @@
+# Edit this file to override any configuration options specified in
+# /usr/lib/container-storage-setup/container-storage-setup.
+#
+# For more details refer to "man container-storage-setup"
```
rhvgoyal: As discussed, lets not rename this file.

container-storage-setup.1
```diff
@@ -0,0 +1,218 @@
...
+
+.SH "SEE ALSO"
+.BR atomic "(1)"
+
+.SH HISTORY
+
+.PP
+November 2014, originally compiled by Joe Brockmeier <jzb@redhat.com>
+based on comments in Andy Grimm's <agrimm@redhat.com> script.
+.SH AUTHORS
```
rhvgoyal: You have done enough changes to this man page that you probably should add your name to it now as well.

container-storage-setup.sh
```diff
@@ -0,0 +1,1286 @@
+#!/bin/bash
```
rhvgoyal: Why this is showing as new file. It should have shown as simple rename. Like libcss.sh.

.travis.yml
rhatdan: Yes I think this is what triggers the tests when we push updates.

tests/101-test-use-devs-to-create-thin-pool.sh
```diff
@@ -22,7 +22,7 @@ VG=$vg_name
 CONTAINER_THINPOOL=container-thinpool
 EOF
 
- # Run docker-storage-setup
+ # Run container-storage-setup
  $DSSBIN $infile $outfile >> $LOGS 2>&1
```
rhvgoyal: Not converting DSSBIN to CSSBIN?

tests/001-test-use-devs-to-create-thin-pool.sh
```diff
@@ -7,7 +7,7 @@ test_devs() {
   local testname=`basename "$0"`
   local vg_name="dss-test-foo"
```
rhvgoyal: not converting dss-test-foo to css-test-foo?

tests/run-tests.sh
```diff
@@ -49,19 +49,22 @@ check_config_files() {
...
+  # One can setup environment variable CONTAINER_STORAGE_SETUP to override
   # which binary is used for tests.
-  if [ -n "$DOCKER_STORAGE_SETUP" ];then
-    if [ ! -f "$DOCKER_STORAGE_SETUP" ];then
+  if [ -z "$CONTAINER_STORAGE_SETUP" -a -n "$DOCKER_STORAGE_SETUP" ];then
+      CONTAINER_STORAGE_SETUP=$DOCKER_STORAGE_SETUP
+  fi
+  if [ -n "$CONTAINER_STORAGE_SETUP" ];then
+    if [ ! -f "$CONTAINER_STORAGE_SETUP" ];then
       echo "Error: Executable $DOCKER_STORAGE_SETUP does not exist"
```
rhvgoyal: You mean `Error: Executable $CONTAINER_STORAGE_SETUP does not exist` ?

.travis.yml
rhvgoyal: That's .redhat-ci.yml I think. This feeds in some other CI.

rhvgoyal:
git pull still shows that docker-storage-setup.sh is new file and not the renamed one (like other files).

create mode 100755 container-storage-setup.sh

rhatdan:
@shishir-a412ed PTAL at last patch in this series.  We are no longer using systemd unit files, and only mounting the file system via container-storage-setup.

rhvgoyal:
mount_extra_volume() is called only when volume is being created. Will not be called after reboot. So we need to make sure it is called on every invocation of docker-storage-setup.service.

rhvgoyal:
Also this particular change can be a separate PR I think. We should also make sure reset logic will cleanup previously created .mount file and put a comment in there that why do we have that code still in there.

tests/001-test-use-devs-to-create-thin-pool.sh
rhatdan: you told me not to until the next one.

tests/101-test-use-devs-to-create-thin-pool.sh
rhatdan: you told me not to until this gets merged.

tests/run-tests.sh
rhatdan: Fixed

rhatdan:
Reset logic still cleans up unit file.

rhatdan:
We decided we wanted this change so that container-storage-setup no longer creates the mount unit files.

rhatdan:
@rhvgoyal Back to this one now.

Makefile
```diff
@@ -4,27 +4,31 @@ PREFIX ?= $(DESTDIR)/usr
...
 	-rm -rf *~ \#* .#*
 
 .PHONY: install
-install: install-docker
+install: install-docker install-core
 
 .PHONY: install-docker
-install-docker: install-core
-	install -D -m 644 ${DOCKER}-storage-setup.service ${UNITDIR}/${DOCKER}-storage-setup.service
+install-docker:
```
rhvgoyal: So install-docker should not install core as well? What does that have anything to do with rename.

rhvgoyal: I think install-docker should install core as well. This target alone is not useful.

Makefile
```diff
@@ -4,27 +4,31 @@ PREFIX ?= $(DESTDIR)/usr
...
+	install -D -m 644 docker-storage-setup.service ${UNITDIR}/${DOCKER}-storage-setup.service
```
rhvgoyal: So we are fixing that docker-latest bug here?

Makefile
```diff
@@ -4,27 +4,31 @@ PREFIX ?= $(DESTDIR)/usr
...
+	install -D -m 644 docker-storage-setup.1 ${MANDIR}/man1/${DOCKER}-storage-setup.1
 
 .PHONY: install-core
 install-core:
-	install -D -m 755 ${DOCKER}-storage-setup.sh ${BINDIR}/${DOCKER}-storage-setup
-	install -D -m 644 ${DOCKER}-storage-setup.conf ${DSSLIBDIR}/${DOCKER}-storage-setup
-	install -D -m 755 libdss.sh ${DSSLIBDIR}/libdss.sh
-	install -D -m 755 dss-child-read-write.sh ${DSSLIBDIR}/dss-child-read-write
-	install -D -m 644 ${DOCKER}-storage-setup.1 ${MANDIR}/man1/${DOCKER}-storage-setup.1
+	install -D -m 755 container-storage-setup.sh ${BINDIR}/container-storage-setup
```
rhvgoyal: so we don't need equivalent of docker-latest, container-lastest?

container-storage-setup.sh
```diff
@@ -499,9 +499,6 @@ lvm_pool_exists() {
   local lvname lv lvsize
   local thinpool_name=$1
 
-  if [ -z "$thinpool_name" ]; then
-      Fatal "Thin pool name must be specified."
-  fi
```
rhvgoyal: Why are you getting rid of this code during rename?

container-storage-setup.spec
```diff
@@ -1,12 +1,12 @@
-%define dsslibdir %{_prefix}/lib/docker-storage-setup
+%define dsslibdir %{_prefix}/lib/container-storage-setup
```
rhvgoyal: should it be csslibdir?

container-storage-setup.spec
```diff
@@ -26,21 +26,11 @@ as the root logical volume and partition table.
...
 
-%post
-%systemd_post %{name}.service
-
-%preun
-%systemd_preun %{name}.service
-
-%postun
-%systemd_postun %{name}.service
-
```
rhvgoyal: Why are we removing above?

container-storage-setup.spec
```diff
@@ -26,21 +26,11 @@ as the root logical volume and partition table.
...
 %files
-%{_unitdir}/docker-storage-setup.service
```
rhvgoyal: Why are we removing this. this package should own container-storage-setup.service?

container-storage-setup.spec
```diff
@@ -26,21 +26,11 @@ as the root logical volume and partition table.
...
+%{_bindir}/container-storage-setup
 %{_bindir}/docker-storage-setup
-%config(noreplace) %{_sysconfdir}/sysconfig/docker-storage-setup
```
rhvgoyal: same here. This package should own /etc/sysconfig/docker-storage-setup

rhvgoyal:
@rhatdan Please address the review comments and push patches again.

Makefile
rhatdan: I thought I put this in the original patch. The idea was to allow docker package to just using the install-docker portion without installing all of the core packages, if it wanted to take them.

Makefile
rhatdan: If a user overrides DOCKER it will blow up.  There is only one docker-storage-setup.service in our pool,  If you wanted to install 
make DOCKER=docker-latest install

This would blow up.

Makefile
rhatdan: Nope since we are no longer hard coding content, docker-latest.service would use

container-storage-setup /etc/sysconfig/docker-latest-storage-setup /etc/sysconfig/docker-latest-storage

container-storage-setup.sh
rhatdan: Bug Fixed.

container-storage-setup.spec
rhatdan: Not in this patch set, next one.  You told me not to do dss->css, to keep the noise down.

container-storage-setup.spec
rhatdan: Because container-storage-setup does not include services.  Those need to move to docker package.

container-storage-setup.spec
rhatdan: We are shipping a container-storage-setup package it does not include any docker services. Those will ship in the docker package.

container-storage-setup.spec
rhatdan: Same as above.  That is not a container-storage-setup package.

rhatdan:
@rhvgoyal Fixed or answered.  If you want the sed -i 's/dss/css/g' patch I can add that.

Makefile
rhvgoyal: Given we are planning to package it separately, I think docker package will have to do all this manually anyway and not rely on these targets.  Even if docker were to use it, they would like to
get core bits as well as comaptibility bits. Can't see how compatibility bits alone are helpful to anybody.

rhvgoyal:
I am not worried too much about dss->css conversion. Lets do it in separate PR once this one is merged.

rhvgoyal:
LGTM.

rhvgoyal:
@rh-atomic-bot r+ 46a4606

PR #198, opened 2017-02-10
Remove default thinpool name, user must specify it

rhatdan:
Add new field CONTAINER_THINPOOL, which is required when using devicemapper.
We still will default to docker-pool if the user does not specify anything
for inputfile, but if user specifies input file, and they are using
devicemapper, they must specify CONTAINER_THINPOOL.

### DIFF
--- a/docker-storage-setup.sh
+++ b/docker-storage-setup.sh
@@ -25,7 +25,6 @@ set -e
 # Read man page for a description of currently supported options:
 # 'man docker-storage-setup'
 
-POOL_LV_NAME="docker-pool"
 DOCKER_ROOT_LV_NAME="docker-root-lv"
 DOCKER_ROOT_DIR="/var/lib/docker"
 DOCKER_METADATA_DIR="/var/lib/docker"
@@ -183,7 +182,7 @@ get_devicemapper_config_options() {
   # docker expects device mapper device and not lvm device. Do the conversion.
   eval $( lvs --nameprefixes --noheadings -o lv_name,kernel_major,kernel_minor $VG | while read line; do
     eval $line
-    if [ "$LVM2_LV_NAME" = "$POOL_LV_NAME" ]; then
+    if [ "$LVM2_LV_NAME" = "$CONTAINER_THINPOOL" ]; then
       echo POOL_DEVICE_PATH=/dev/mapper/$( cat /sys/dev/block/${LVM2_LV_KERNEL_MAJOR}:${LVM2_LV_KERNEL_MINOR}/dm/name )
     fi
   done )
@@ -321,6 +320,10 @@ check_min_data_size_condition() {
 }
 
 create_lvm_thin_pool () {
+  if [ "$STORAGE_DRIVER" == "devicemapper" ] && [ -z "$CONTAINER_THINPOOL" ];then
+     Fatal "CONTAINER_THINPOOL must be defined for the devicemapper storage driver."
+  fi
+
   if [ -z "$DEVS_ABS" ] && [ -z "$VG_EXISTS" ]; then
     Fatal "Specified volume group $VG does not exist, and no devices were specified"
   fi
@@ -354,7 +357,7 @@ create_lvm_thin_pool () {
     DATA_SIZE_ARG="-L $DATA_SIZE"
   fi
 
-  lvcreate -y --type thin-pool --zero n $CHUNK_SIZE_ARG --poolmetadatasize ${META_SIZE}s $DATA_SIZE_ARG -n $POOL_LV_NAME $VG
+  lvcreate -y --type thin-pool --zero n $CHUNK_SIZE_ARG --poolmetadatasize ${META_SIZE}s $DATA_SIZE_ARG -n $CONTAINER_THINPOOL $VG
 }
 
 get_configured_thin_pool() {
@@ -422,9 +425,10 @@ reset_extra_volume () {
 }
 
 reset_lvm_thin_pool () {
-  if lvm_pool_exists; then
-      lvchange -an $VG/${POOL_LV_NAME}
-      lvremove $VG/${POOL_LV_NAME}
+  local thinpool_name=${CONTAINER_THINPOOL:-docker-pool}
+  if lvm_pool_exists $thinpool_name; then
+      lvchange -an $VG/${thinpool_name}
+      lvremove $VG/${thinpool_name}
   fi
 }
 
@@ -433,13 +437,14 @@ setup_lvm_thin_pool () {
   # Check if a thin pool is already configured in /etc/sysconfig/docker-storage.
   # If yes, wait for that thin pool to come up.
   tpool=`get_configured_thin_pool`
+  local thinpool_name=${CONTAINER_THINPOOL:-docker-pool}
 
   if [ -n "$tpool" ]; then
-     local escaped_pool_lv_name=`echo $POOL_LV_NAME | sed 's/-/--/g'`
+     local escaped_pool_lv_name=`echo $thinpool_name | sed 's/-/--/g'`
      Info "Found an already configured thin pool $tpool in ${STORAGE_OUT_FILE}"
 
      # dss generated thin pool device name starts with /dev/mapper/ and
-     # ends with POOL_LV_NAME
+     # ends with $thinpool_name
      if [[ "$tpool" != /dev/mapper/*${escaped_pool_lv_name} ]];then
        Fatal "Thin pool ${tpool} does not seem to be managed by docker-storage-setup. Exiting."
      fi
@@ -457,7 +462,7 @@ setup_lvm_thin_pool () {
     VG_EXISTS=1
   fi
 
-  if ! lvm_pool_exists; then
+  if ! lvm_pool_exists $thinpool_name; then
     check_docker_storage_metadata
     create_lvm_thin_pool
     write_storage_config_file
@@ -472,15 +477,16 @@ setup_lvm_thin_pool () {
 
   # Enable or disable automatic pool extension
   if [ "$AUTO_EXTEND_POOL" == "yes" ];then
-    enable_auto_pool_extension ${VG} ${POOL_LV_NAME}
+    enable_auto_pool_extension ${VG} ${thinpool_name}
   else
-    disable_auto_pool_extension ${VG} ${POOL_LV_NAME}
+    disable_auto_pool_extension ${VG} ${thinpool_name}
   fi
 }
 
 lvm_pool_exists() {
   local lv_data
   local lvname lv lvsize
+  local thinpool_name=$1
 
   lv_data=$( lvs --noheadings -o lv_name,lv_attr --separator , $VG | sed -e 's/^ *//')
   SAVEDIFS=$IFS
@@ -488,7 +494,7 @@ lvm_pool_exists() {
   IFS=,
   read lvname lvattr <<< "$lv"
     # pool logical volume has "t" as first character in its attributes
-    if [ "$lvname" == "$POOL_LV_NAME" ] && [[ $lvattr == t* ]]; then
+    if [ "$lvname" == "$thinpool_name" ] && [[ $lvattr == t* ]]; then
       IFS=$SAVEDIFS
       return 0
     fi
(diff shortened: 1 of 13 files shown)


docker-storage-setup.sh
```diff
@@ -990,6 +989,10 @@ setup_docker_root_lv_fs() {
 }
 
 check_storage_options(){
+  if [ "$STORAGE_DRIVER" == "devicemapper" ] && [ -n "$CONTAINER_THINPOOL" ];then
```
rhvgoyal: You mean `[ -z  $CONTAINER_THINPOOL ]` ?

docker-storage-setup.sh
```diff
@@ -1191,7 +1194,10 @@ if [ $# -eq 2 ]; then
     STORAGE_IN_FILE=$1
     STORAGE_OUT_FILE=$2
 else
-    if [ $# -gt 0 ]; then
+    if [ $# -eq 0 ]; then
+	# Default for the docker use case
+	CONTAINER_THINPOOL="docker-pool"
```
rhvgoyal: What happens if docker user specifies CONTAINER_THINPOOL. I think we should still honor that?

So this probably should be 

` [  - z "$CONTAINER_THINPOOL ] && CONTAINER_THINPOOL="docker-pool" ?

rhvgoyal:
Can we add a test for CONTAINER_THINPOOL, so that if user does not specify it, dss fails with appropriate message.

rhatdan:
@rh-atomic-bot test

rhatdan:
Ok tests are passing and the new test seems to work.

@rhvgoyal PTAL

rhatdan:
@rhvgoyal Ping...

tests/110-test-fail-if-no-thinpool.sh
```diff
@@ -0,0 +1,36 @@
...
+      test_status=0
+  else
+      echo "ERROR: $testname: $DSSBIN Succeeded. Should have failed with CONTAINER_THINPOOL specified" >> $LOGS
+  fi
+  cleanup $vg_name "$DEVS" "$infile" "$outfile"
+  return $test_status
+}
+
+# Make sure command fails if no CONTAINER_THINPOOL is specified
+test_fail-if-no-thinpool
```
rhvgoyal: Typically all function names have underscore as separator. This seems to have both underscore and hyphen as separators. Would be good to have consistent usage of underscore.

rhvgoyal: Can we rename test file to `110-test-fail-if-no-container-thinpool.sh`

tests/110-test-fail-if-no-thinpool.sh
```diff
@@ -0,0 +1,36 @@
+source $SRCDIR/libtest.sh
+
+test_fail-if-no-thinpool() {
+  local devs dev
+  local devlinks devlink
```
rhvgoyal: Not being used.

tests/110-test-fail-if-no-thinpool.sh
```diff
@@ -0,0 +1,36 @@
...
+  cat << EOF > $infile
+DEVS="$devs"
+VG=$vg_name
+EOF
+  # Run docker-storage-setup
+  $DSSBIN $infile $outfile >> $LOGS 2>&1
+
+  # Test failed.
+  if [ $? -ne 0 ]; then
+      test_status=0
```
rhvgoyal: I think we should also grep for error message and make sure failure happened due to CONTAINER_THINPOOL not being specified and not because of some other reason.

rhatdan: We don't currently check error output since we are redirecting to the logs.  I guess we could output to a temp log file and then cat that output into the logs after checking?

rhatdan: Fixed

rhvgoyal:
@rhatdan Tests are failing.

rhatdan:
Ok looks like cat command was destroying return code.  I have updated to save the return code.

docker-storage-setup.sh
```diff
@@ -990,6 +989,10 @@ setup_docker_root_lv_fs() {
 }
 
 check_storage_options(){
+  if [ "$STORAGE_DRIVER" == "devicemapper" ] && [ -z "$CONTAINER_THINPOOL" ];then
+     Fatal "CONTAINER_THINPOOL must be defined for the devicemapper storage driver."
+  fi
+
```
rhvgoyal: One last thing. Can we move this check in create_lvm_thin_pool(). And reason being that we have shipped code where we can create a thin pool without passing CONTAINER_THINPOOL. I don't want these people to be broken over upgrade. 

So this check will kick in only if we are creating a new thin pool. If thin pool is already there, then this check should not kick in.

rhatdan: This code will not kick in if the user has not specified options to container-storage-setup.  In that code path docker-pool is always set.

rhatdan:
@rh-atomic-bot retest

rhatdan:
bot retest

docker-storage-setup.sh
rhvgoyal: Right. So we have two backward compatibility cases.

- No options specified. This is docker case and it will not be broken
- Options specified and we will create docker-pool. Some fedora package might have pulled this code in. These are the users I am concerned about.

rhvgoyal:
@rhatdan checks are failing.

rhatdan:
@rhvgoyal Finally passes.

docker-storage-setup.sh
```diff
@@ -435,11 +438,11 @@ setup_lvm_thin_pool () {
   tpool=`get_configured_thin_pool`
 
   if [ -n "$tpool" ]; then
-     local escaped_pool_lv_name=`echo $POOL_LV_NAME | sed 's/-/--/g'`
+     local escaped_pool_lv_name=`echo $CONTAINER_THINPOOL | sed 's/-/--/g'`
```
rhvgoyal: I think this will be a problem if a thin pool with name docker-root-lv was already there? In this case CONTAINER_THINPOOL will be nil.

rhatdan: If $argv1 is not specifyied CONTAINER_THINPOOL will be docker-root-lv.  Only way this would be a problem would be if the user specified a service file without CONTAINER_THINPOOL specified.

rhatdan: I have changed the code again to just check if the STORAGE_IN_FILE is /etc/sysconfig/docker-storage-setup then it will default CONTAINER_THINPOOL to docker-pool if not set.

rhvgoyal: did you get rid of checks which did not allow passing in only inputfile. If user specifies only inputfile and not outputfile, it should error out.

rhvgoyal: Also, instead of hardcoding based on /etc/sysconfig/docker-storage-setup as inputfile, can we use pool name as docker-lv when user has not specified CONTAINER_THINPOOL and we are not creating new thing pool.

IOW, if we are creating new thin pool we make sure COINTAINER_THINPOOL is specified, otherwise if CONTAINER_THINPOOL is not specified, then we fall back to using docker-pool.

rhatdan: I figured to allow it to still default to docker-storage if they only specify input.  Simplifies the code and  no huge benefit to the error code.  I guess not allowing them to specify more then 2 is no longer covered, but I don't see that is all that useful.

rhatdan: Isn't docker-lv something that could be overriden by the user?

tests/001-test-use-devs-to-create-thin-pool.sh
```diff
@@ -17,6 +17,7 @@ test_devs() {
   cat << EOF > /etc/sysconfig/docker-storage-setup
 DEVS="$devs"
 VG=$vg_name
+CONTAINER_THINPOOL=docker-pool
```
rhvgoyal: All the tests below 100 are docker legacy tests and should work as it is. these are good tests to make sure docker continues to work even if CONTAINER_THINPOOL is not specified.

So I don't think we should be modifying tests below 100. Only new tests should specify CONTAINER_THINPOOL.

docker-storage-setup.sh
```diff
@@ -1190,11 +1196,6 @@ done
 if [ $# -eq 2 ]; then
     STORAGE_IN_FILE=$1
     STORAGE_OUT_FILE=$2
-else
-    if [ $# -gt 0 ]; then
-	usage $(basename $0) >&2
-	exit 1
-    fi
 fi
```
rhvgoyal: Why are you taking this check away. We want to make sure that either user need to specify both input and output file or does not specify any.

rhatdan: Ok I will add it back.

tests/001-test-use-devs-to-create-thin-pool.sh
rhatdan: We agreed yesterday to only support docker-pool on systems that had it precreated.  These tests all fail since they are creating new thin-pools.  The only other way to make these pass is to hard code it in the case that the user does not specify a any arguments, or specified docker-storage-setup as arg1, which is what you rejected.

rhvgoyal: We had agreed that pool name will continue to be docker-pool if user did not pass in any input/output files. This is legacy docker mode. So if we do that, these tests should not fail.

rhvgoyal: I had objected to third mode that is if input file name is "/etc/sysconfig/docker-storage-setup" then pool name will be docker-pool.

rhatdan: Then the fixes we did yesterday will not take effect except in the weird case where they specify the a argv1 and argv2 that does not have a CONTAINER_THINPOOL defined, but they previously created a docker-pool thinpool.

rhvgoyal: So here are my thoughts.
- If user did not pass in inputfile/outputfile and CONTAINER_THINPOOL is not set, pool name will be dokcker-pool. 
- If user did not pass in inputfile/outputfile and CONTAINER_THINPOOL is set, pool name will be as specified by CONTAINER_THINPOOL.
- If user did specify inputfile/outputfile but no CONTAINER_THINPOOL, this should error out if we are about to create new thin pool.
- If user did specify inputfile/output but no CONTAINER_THINPOOL, this should fall back to docker-pool for the case of upgrade and reset. This will ensure that users who used inputfile/outputfile and still got pool name as docker-pool continue to work and are not broken.

rhatdan: I disagree with 3/4,  If you specify a input/output file you should be forced to do it correctly.  But your the boss, so I have updated the pull request to match this.

rhatdan: I think we are still going to have a problem going forward.  If @lsm5 updates to use container-storage-setup inside of docker.service and it specifies correctly CONTAINER_THINPOOL=container-pool, then on upgrade this is going to cause us issues.

rhatdan: @lsm5 this is not a problem as long as CONTAINER_THINPOOL=docker-pool.  So when we update docker.service to using container-storage-setup with inputfile/outputfile, we just need to make sure we add 
In docker-storage-setup file we ship.

docker-storage-setup.sh
```diff
@@ -422,9 +425,10 @@ reset_extra_volume () {
 }
 
 reset_lvm_thin_pool () {
+  local thinpool_name=${CONTAINER_THINPOOL:-docker-pool}
   if lvm_pool_exists; then
```
rhvgoyal: --reset is broken. you need to pass in name of thin pool to lvm_pool_exists.

if lvm_pool_exists $thinpool_name; then
fi

rhvgoyal: I am wondering why automated tests did not catch this.

rhvgoyal: oh, test 005 only checks that file /etc/sysconfig/docker-storage got removed. And does not check if thin pool docker-pool got removed or not. That's why automated test did not catch this.

rhvgoyal:
LGTM. Thanks

rhvgoyal:
@rh-atomic-bot r+ cb2e726

PR #200, opened 2017-02-21
This patch set should catch the errors we missed when docker-storage-setup --reset failed.

rhatdan:


rhvgoyal:
@rhatdan tests are failing.

rhvgoyal:
docker-storage-setup --reset, does not clean VG or underlying PVs. That's done by atomic tool. So you will have to modify test to look for docker-pool thin pool lv and make sure that's gone away.

rhatdan:
@rh-atomic-bot Sounds good, What are the lv* commands to check this?

rhvgoyal:
@rhatdan we have a helper function called lv_exists(). you can ues that.

rhatdan:
@rhvgoyal I wrote my own, before I saw your note.  I am now using lv_exists.

tests/005-test-devmapper-cleanup.sh
```diff
@@ -31,13 +31,19 @@ EOF
   $DSSBIN --reset >> $LOGS 2>&1
   # Test failed.
   if [ $? -eq 0 ]; then
-     if [ ! -e /etc/sysconfig/docker-storage ]; then
-          test_status=0
+     if [ -e /etc/sysconfig/docker-storage ]; then
+	 echo "ERROR: $testname: $DSSBIN failed. /etc/sysconfig/docker-storage still exists." >> $LOGS
      else
-	  echo "ERROR: $testname: $DSSBIN failed. /etc/sysconfig/docker-storage still exists." >> $LOGS
+	 if lvm_exists $vg_name "docker-pool"; then
```
rhvgoyal: This should be lv_exists and not lvm_exists?

rhatdan: Why didn't the test blow up?

rhvgoyal: Looks like bash does not exit if command not found. I see following on stdout and test still passes.

/tests/005-test-devmapper-cleanup.sh: line 38: lvm_exists: command not found

rhvgoyal:
LGTM

I really would like to reduce the nesting of if statements. I think there should be a better way to organize this code. But that's for another PR.

rhvgoyal:
@rh-atomic-bot r+ f97543a

rhatdan:
Perhaps we should add
set -e 

For all test scripts so that they crash on bugs like this.

rhvgoyal:
I tried adding "set -e" to the test and it did not seem to help.

rhatdan:
Ok so since this was in an if statement that was expected to fail, it fails and sets the state to success.
I don't think we could catch this.

PR #201, opened 2017-02-21
Allow installing docker-storage-setup without docker.service

rhatdan:
We give users the abiltiy to install-min which will just install
docker-storage-setup without installing docker.service of
/etc/sysconfig/docker-storage-service file


rhvgoyal:
So who needs to install dokcer-storage-setup.service by default? I had agreed that rpms should drop a docker-storage-setup.service file, if need be.

Do you have a use case in mind?

rhvgoyal:
Re-reading it again. So you have basically divided it in two targets. install-min just installs core bits and install-docker installs service file and empty /etc/sysconfig/docker-storage-setup too. 

And once we move to container-storage-setup, most of the users will use install-min and install-docker will be there for legacy purposes?

I guess that's fine.

rhvgoyal:
How about `install-core` instead of `install-min`?

rhatdan:
Now uses install-core

rhatdan:
Yes install-docker is just their for anyone who might have used docker-storage-setup directly from master and relied on it installing the docker.service.  We will have @lsm5 start to install-core in the container-storage-setup package and will have to add docker.service and /etc/sysconfig/docker-storage-setup files in the docker package.

rhvgoyal:
LGTM

rhvgoyal:
@rh-atomic-bot r+ b226185

PR #202, opened 2017-02-23
Stop using service file for mounting extra volumes

rhatdan:
Have container-storage-setup mount it, itself.


rhvgoyal:
Dan, would be good to make commit logs give more description that why are we preferring this method over a mount unit file.

docker-storage-setup.sh
```diff
@@ -964,6 +946,9 @@ setup_extra_lv_fs() {
     return 1
   fi
   if extra_volume_exists $CONTAINER_ROOT_LV_NAME; then
+    if ! mount_extra_volume $CONTAINER_ROOT_LV_NAME $RESOLVED_MOUNT_DIR_PATH; then
+      Fatal "Failed to mount volume $CONTAINER_ROOT_LV_NAME on $RESOLVED_MOUNT_DIR_PATH"
+    fi
```
rhvgoyal: same change needs to be done in setup_docker_root_lv_fs() as well. That path is used to setup extra volume for docker root lv.

rhatdan:
@rhvgoyal PTAL

rhvgoyal:
@rhatdan So after upgrade, both systemd and our script will try to mount an existing extra volume. That's like a race condition and we might even see a double mount?

rhvgoyal:
I think we will have to check if there is existing mount unit file, stop it and disable it, remove it and then make sure mount point is mounted. 

IOW, mount unit file need to be cleaned up during first run after upgrade so that we don't have both docker-storage-setup and systemd mounting the volume.

rhatdan:
Our code should handle the fact that it is already mounted.   And the systemd.mount should be complete before we mount it.  If we want to do what you want we would need to do it every time we mount, which is ok. But it will need to happen forever.

rhatdan:
Actually I don't think there would be a double mount, I would figure that systemd would realize it is already mounted as well. We could try this by creating a target mount and stop it, mount the file system and then start the unit file.

rhvgoyal:
I think it is a race which can still happen. (Though hard to reproduce).

We are doing this. 

if `is_not_mounted; then
    mount_volume;
fi

Now it can happen that after call to if_not_mounted() finished, systemd does the mounting. And we will do another mounting on top.

So I think it makes sense to cleanup mount unit file always if it exists. Little extra code which we have to always run, but atleast we get rid of race condition.

rhatdan:
@rhvgoyal PTAL, I added a call to always remove it.  Theoretically you could still get a double mount, but this is probably the best we can do.

docker-storage-setup.sh
```diff
@@ -395,6 +395,21 @@ systemd_escaped_filename () {
...
+
+#
+# In the past we created a systemd mount target file, we no longer
+# use it, but if one pre-existed we still need to handle it.
+#
+remove_systemd_mount_target () {
+  local mp=$1
+  local filename=$(systemd_escaped_filename $mp)
+  if [ -f /etc/systemd/system/$filename ];then
+    systemctl disable $filename >/dev/null 2>&1
```
rhvgoyal: lets call `systemctl stop $filename` after disabling it. That unmounts any mount done by systemd.

rhvgoyal:
@rhatdan What race can still happen theoretically?

rhatdan:
@rhvgoyal Fixed.

rhvgoyal:
LGTM

rhvgoyal:
@rh-atomic-bot r+ 138a145

rhvgoyal:
@rhatdan this PR breaks tests on rhel. Old findmnt utility does not support --mountpoint option.

rhatdan:
Does --target?

PR #203, opened 2017-02-23
Change all internal refereces for dss to css

rhatdan:
This is just a cleanup to refer to container-storage-setup (css) rather
then docker-storage-setup (dss)

### DIFF
--- a/container-storage-setup.sh
+++ b/container-storage-setup.sh
@@ -34,8 +34,8 @@ STORAGE_IN_FILE="/etc/sysconfig/docker-storage-setup"
 STORAGE_OUT_FILE="/etc/sysconfig/docker-storage"
 STORAGE_DRIVERS="devicemapper overlay overlay2"
 
-PIPE1=/run/dss-fifo1
-PIPE2=/run/dss-fifo2
+PIPE1=/run/css-fifo1
+PIPE2=/run/css-fifo2
 TEMPDIR=$(mktemp --tmpdir -d)
 
 # DEVS can have device names without absolute path. Convert these to absolute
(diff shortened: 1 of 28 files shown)


container-storage-setup.sh
```diff
@@ -34,8 +34,8 @@ STORAGE_IN_FILE="/etc/sysconfig/docker-storage-setup"
 STORAGE_OUT_FILE="/etc/sysconfig/docker-storage"
 STORAGE_DRIVERS="devicemapper overlay overlay2"
 
-PIPE1=/run/dss-fifo1
-PIPE2=/run/dss-fifo2
+PIPE1=/run/css-fifo1
+PIPE2=/run/css-fifo2
```
rhvgoyal: This will create issues with multiple css instances running, right? We probably should introduce some locking so that only one instance of container-storage-setup is running at a time.

rhvgoyal: This improvement should be a separate PR though.

rhatdan: Can we just add the pid of the running container so that they are different?

rhvgoyal: You mean create a new pair of pipes for each running instance? I think child needs to know in advance the name of pipe to wait on.

rhvgoyal: Ok, we are passing the name of named fifo pipe to child. So yes, using pid of running process should be fine, I guess.  This is better than locking.

rhvgoyal:
LGTM

rhvgoyal:
@rh-atomic-bot r+ 14fac9e

PR #206, opened 2017-02-24
Rewrite the README to fully explain what container-storage-setup is

rhatdan:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,27 +1,75 @@
+# Container Storage Setup
+
+## Tool for setting up container runtimes storage
+
+`container-storage-setup` is part of the [Project Atomic](http://www.projectatomic.io/) suite of container projects, formerly known as docker-storage-setup.
+
+
+A crucial aspect to container runtimes is the concept of the copy-on-write (COW) layered filesystems.  The [Docker Engine Storage docs](https://docs.docker.com/engine/userguide/storagedriver/) site explains how the docker daemon uses uses COW file systems
+
+`container-storage-setup` is a script to configure COW File systems like devicemapper and overlayfs.   It is usually run via a systemd service.  For example `docker-storage-setup.service`, runs `container-storage-setup` before the docker.service script starts the docker daemon.
+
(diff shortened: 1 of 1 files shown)


rhvgoyal:
I am not sure about the idea of copying all the info from man page or config file in this page. Just providing
links should be enough. Otherwise we will have now 3 places to update for any documentation change.

We already have a link to man page in the doc and that seems to do a reasonable job.

rhatdan:
I would like to experiment with writing the man page as a markdown, and then some how have it in the README and the man page.

The problem with the README, now is it does not describe enough to make it interesting.

rhatdan:
Is there a way to link to the man page from the README, that would also be cool.

rhvgoyal:
May be we should put in a simple example also there to quickly demonstrate what it can do. That should make it little more interesting.

rhvgoyal:
@rhatdan we already have a link to man page in README at the bottom.

rhatdan:
Perhaps if it was written in Markdown, that would be better.

Take a look at:
https://github.com/docker/docker/blob/master/man/docker-run.1.md

rhatdan:
But Anyways it is probably better to discuss what the README should look like in a google doc.

rhvgoyal:
How would man utility parse a man page written in md. Does it already do it?

rhatdan:
The is a tool go-md2man that converts from markdown to nroff format, you would do this during make install

rhvgoyal:
So then we will have to create a dependency on this go-md2man tool. I am not sure which rpm/package provides that tool.

rhvgoyal:
How about keeping it simple for now. That is enhance existing README to explain what this project is all about with a simple example. And then provide a link to man page.

And converting man page to md can be a separate conversation in a separate PR?

rhatdan:
Sure.  go-md2man is shipped as an rpm in Fedora/Centos and RHEL.  It is needed for building docker and some other packages, but I am fine with removing it.  I sent you a link to the google doc to quickly finish the README.

rhvgoyal:
@rh-atomic-bot r+ 79462e9

PR #207, opened 2017-02-24
Always default docker.service to NOT use input file

rhatdan:
If someone uses the INPUTFILE/OUTPUTFILE in a service on an existing
install, on update there is a chance things will break.  Since most likely
/etc/sysconfig/docker-storage-setup will not have


rhvgoyal:
LGTM

rhvgoyal:
@rh-atomic-bot r+ a017297

PR #208, opened 2017-02-24
Look for default config file and helper exec file in source dir first

rhvgoyal:
We put all defaults in container-storage-setup.conf and look for this
file in /usr/lib/container-storage-setup. Instead first look for this
file in source dir where container-storage-setup is present and if not
found look in /usr/lib/

It solves one main problem that is one does not have to do "make install"
to test the code. Everything needed is available in source code itself.

It also helps if during code development, if one introduces a new default
variable, then without doing "make install" one can test their code.

During migration from docker-storage-setup to container-storage-setup also
it continues to work. Otherwise without this change, script can't find
container-storage-setup.conf in /usr/lib/container-storage-setup/ and
breaks down.

### DIFF
--- a/container-storage-setup.sh
+++ b/container-storage-setup.sh
@@ -118,12 +118,16 @@ should_enable_deferred_deletion() {
 platform_supports_deferred_deletion() {
         local deferred_deletion_supported=1
         trap cleanup_pipes EXIT
-        if [ ! -x "/usr/lib/container-storage-setup/css-child-read-write" ];then
+        local child_exec="$SRCDIR/css-child-read-write.sh"
+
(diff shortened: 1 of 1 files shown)


rhvgoyal:
@rhatdan PTAL

rhvgoyal:
I am not sure why tests are not kicking in.

rhvgoyal:
@jlebon it took almost 3 mins for tests to kick in. Is it expected.

rhatdan:
LGTM

rhvgoyal:
@rh-atomic-bot r+ 6bd7557

PR #209, opened 2017-02-24
Use findmnt option --target instead of --mountpoint

rhvgoyal:
Old findmnt utility (on rhel/centos) does not support new option --mountpoint
and tests fail there.

### DIFF
--- a/container-storage-setup.sh
+++ b/container-storage-setup.sh
@@ -903,7 +903,7 @@ extra_lv_mountpoint() {
   local mounts
   local lv_name=$1
   local mount_dir=$2
-  mounts=$(findmnt -n -o TARGET --source /dev/$VG/$lv_name --mountpoint $mount_dir | head -1)
+  mounts=$(findmnt -n -o TARGET --source /dev/$VG/$lv_name | grep $mount_dir)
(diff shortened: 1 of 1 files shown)


rhvgoyal:
@rhatdan PTAL

rhatdan:
LGTM

rhvgoyal:
@rhatdan I think --target has some problems on rhel. It lists a mount point even if --target does not match. So we can't rely on --mountpoint or --target options in rhel.

rhvgoyal:
For example, 

[root@vm6-rhel7 ~]# findmnt --source /dev/mapper/rhel-home --target /home1 -n 
/home  /dev/mapper/rhel-home xfs    rw,relatime,seclabel,attr2,inode64,noquota

Above --target is home1 which does not match home still findmnt lists the mountpoint. In fedora that does not happen. So it is a bug in findmnt I think.

rhvgoyal:
I guess that must have been the reason we did not use --target to begin with in original implementation.

rhatdan:
LGTM

rhvgoyal:
@rh-atomic-bot r+ 337ae8b

PR #213, opened 2017-02-27
rpmlint does not like us using /usr/lib.

rhatdan:
Fedora review is finding some small issues.  This pull request is an attempt to fix these.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ PREFIX ?= $(DESTDIR)/usr
 BINDIR ?= $(PREFIX)/bin
 MANDIR ?= $(PREFIX)/share/man
 UNITDIR ?= $(PREFIX)/lib/systemd/system
-CSSLIBDIR ?= $(PREFIX)/lib/container-storage-setup
+CSSLIBDIR ?= $(PREFIX)/share/container-storage-setup
 SYSCONFDIR ?= $(DESTDIR)/etc/sysconfig
 
 .PHONY: clean
@@ -36,6 +36,9 @@ install-core:
 dist: container-storage-setup.spec
 	spectool -g container-storage-setup.spec
 
+srpm: dist
+	rpmbuild --define "_sourcedir `pwd`" --define "_specdir `pwd`" \
+	--define "_rpmdir `pwd`" --define "_srcrpmdir `pwd`" -bs container-storage-setup.spec
 rpm: dist
 	rpmbuild --define "_sourcedir `pwd`" --define "_specdir `pwd`" \
 	--define "_rpmdir `pwd`" --define "_srcrpmdir `pwd`" -ba container-storage-setup.spec
--- a/container-storage-setup.spec
+++ b/container-storage-setup.spec
@@ -1,6 +1,6 @@
 %global project container-storage-setup
 %global git0 https://github.com/projectatomic/%{repo}
-%define csslibdir %{_prefix}/lib/container-storage-setup
+%define csslibdir %{_prefix}/share/container-storage-setup
 %global commit         79462e9565053fb1e0d87c336e6d980f0a56c41e
 %global shortcommit    %(c=%{commit}; echo ${c:0:7})
 %global repo %{project}
--- a/container-storage-setup.spec
+++ b/container-storage-setup.spec
@@ -15,10 +15,7 @@ URL:            http://github.com/projectatomic/container-storage-setup/
 Source0: %{git0}/archive/%{commit}/%{repo}-%{shortcommit}.tar.gz
 BuildArch: noarch
 
-BuildRequires:  pkgconfig(systemd)
-
 Requires:       lvm2
-Requires:       systemd-units
 Requires:       xfsprogs 
 
 %description
--- a/container-storage-setup.spec
+++ b/container-storage-setup.spec
@@ -1,6 +1,6 @@
 %global project container-storage-setup
 %global git0 https://github.com/projectatomic/%{repo}
-%define csslibdir %{_prefix}/share/container-storage-setup
+%global csslibdir %{_prefix}/share/container-storage-setup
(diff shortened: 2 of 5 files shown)


cgwalters:
I think rpmlint is happier with `/usr/libexec/` FWIW.  (But in any case that rpmlint check is really wrong, since it only matters for multilib, which we *definitely* don't care about for this)

rhatdan:
@lsm5 PTAL

container-storage-setup.spec
```diff
@@ -15,10 +15,7 @@ URL:            http://github.com/projectatomic/container-storage-setup/
 Source0: %{git0}/archive/%{commit}/%{repo}-%{shortcommit}.tar.gz
 BuildArch: noarch
 
-BuildRequires:  pkgconfig(systemd)
```
rhvgoyal: We still have systemctl commands. So we atleast have install time/run time depedency on systemd. (If not build time dependency).

container-storage-setup.spec
```diff
@@ -15,10 +15,7 @@ URL:            http://github.com/projectatomic/container-storage-setup/
...
-
 Requires:       lvm2
-Requires:       systemd-units
```
rhvgoyal: I am assuming that atleast we need to keep "Requires"  on systemd.

rhvgoyal:
So what is rpmlint complaining about? dracut, systemd all store package shipped information in /usr/lib/

also libcss.sh is definitely a helper library.

Makefile
```diff
@@ -4,7 +4,7 @@ PREFIX ?= $(DESTDIR)/usr
 BINDIR ?= $(PREFIX)/bin
 MANDIR ?= $(PREFIX)/share/man
 UNITDIR ?= $(PREFIX)/lib/systemd/system
-CSSLIBDIR ?= $(PREFIX)/lib/container-storage-setup
+CSSLIBDIR ?= $(PREFIX)/share/container-storage-setup
```
lsm5: just wanted to mention, %{_libdir} is /usr/lib64 on x86_64 systems in case you'd like to put things there. It also evaluates to /usr/lib on x86 systems. So we can just use %{_libdir} if you're not really keen on /usr/share

container-storage-setup.spec
```diff
@@ -1,6 +1,6 @@
 %global project container-storage-setup
 %global git0 https://github.com/projectatomic/%{repo}
-%define csslibdir %{_prefix}/lib/container-storage-setup
+%global csslibdir %{_prefix}/share/container-storage-setup
```
lsm5: I think we can continue keeping this at %{_prefix}/lib. Changing %define to %global was what would make one of the complains go away.

lsm5: complaints*

container-storage-setup.spec
lsm5: I see that the current package doesn't need to start/restart any unitfiles on install/upgrade. So, we can skip the systemd requirement.

rhatdan: No need for systemd.  We don't ship a service file.

Makefile
rhatdan: library is usually used for storing executables, which I think the rpmlint is complaining about. Since we are just shipping scripts, then putting them into /usr/share seems fine.  I think systemd just ignores the error.

container-storage-setup.spec
rhatdan: We should not have any systemctl commands executed in the spec?

rhatdan: We don't ship unit files so no service needed to cycle.

rhatdan:
I would say libcss.sh is a helper script rather then a library but that is semantics.

lsm5:
@rhatdan my comments have been addressed, I'll leave the rest to @rhvgoyal and you :)

lsm5:
ohh I forgot the thing about a LICENSE file, great if you can include it, but no problem if you don't want to

container-storage-setup.spec
rhvgoyal: @rhatdan how do we make sure systemctl is installed on the system? container-storage-setup uses that.

rhvgoyal: systemd package provides that. So we should have install time depency on "systemd"?

lsm5:
@rhatdan LICENSE should go in a separate line like so:

`%license LICENSE`

Rest looks good

container-storage-setup.spec
lsm5: @rhvgoyal systemd is a core package which will always be installed on fedora and other distros that we care about. Looks like even where unitfiles are started via scriptlets, systemd is only needed as a buildtime dep. See https://fedoraproject.org/wiki/Packaging:Scriptlets?rd=Packaging:ScriptletSnippets#Systemd . So skipping it should be totally ok.

lsm5: systemd will always be pre-existing* on the distro is what I meant to say.

rhatdan:
I have moved the LICENSE to a separate pull request.  Lets get this merged so we can approve the package and get it into Rawhide.

container-storage-setup.spec
rhvgoyal: @lsm5 Ok, I will buy it for now. If somebody takes this spec and tries to use it on a non-systemd distro, i think they would run into issues.

rhatdan: We probably should fix the script to not blow up if /usr/sbin/systemctl does not exist. Most likely if it does not exist, we don't need to run that code.  Since we are only attempting to cleanup some thing that was done for systemd...

rhvgoyal:
LGTM

rhvgoyal:
@rh-atomic-bot r+ cee5aa7

rhatdan:
@lsm5 Can we get the bugzilla approved now, and I will add LICENSE early next week if/when we get the ACKS?

PR #217, opened 2017-03-01
Change STORAGE_DRIVERS to STORAGE_DRIVER_LIST to prevent conflicts

rhatdan:
We use STORAGE_DRIVER passed in by the user, and STORAGE_DRIVERS intenally,
this change makes it more difficult to accidently break contaner-storage-setup.


rhatdan:
One other idea would be to prefix all STORAGE_* with an _.

_STORAGE_DRIVERS
_STORAGE_IN_FILE
_STORAGE_OUT_FILE

rhatdan:
This would make https://github.com/projectatomic/container-storage-setup/pull/217 less likely to happen

rhvgoyal:
@rhatdan can you please have a look at #218 and see how does that look.

rhvgoyal:
This is not needed any more with the merge of PR #220

PR #218, opened 2017-03-01
Set internal gloabls after sourcing input config file

rhvgoyal:
This will make sure that config file can't set any global
variables of container-storage-setup which are internal and
change behavior unexpectedly.

### DIFF
--- a/container-storage-setup.sh
+++ b/container-storage-setup.sh
@@ -21,40 +21,45 @@
 
 set -e
 
-# container-storage-setup version information
-CSS_MAJOR_VERSION="0"
-CSS_MINOR_VERSION="1"
-CSS_SUBLEVEL="0"
-CSS_EXTRA_VERSION=""
+set_overridable_globals() {
+  DOCKER_ROOT_VOLUME_SIZE=40%FREE
+}
 
-CSS_VERSION="${CSS_MAJOR_VERSION}.${CSS_MINOR_VERSION}.${CSS_SUBLEVEL}"
-[ -n "$CSS_EXTRA_VERSION" ] && CSS_VERSION="${CSS_VERSION}-${CSS_EXTRA_VERSION}"
+set_internal_globals() {
+  # container-storage-setup version information
+  CSS_MAJOR_VERSION="0"
+  CSS_MINOR_VERSION="1"
+  CSS_SUBLEVEL="0"
+  CSS_EXTRA_VERSION=""
 
-# This section reads the config file $INPUTFILE
-# Read man page for a description of currently supported options:
-# 'man container-storage-setup'
+  CSS_VERSION="${CSS_MAJOR_VERSION}.${CSS_MINOR_VERSION}.${CSS_SUBLEVEL}"
+  [ -n "$CSS_EXTRA_VERSION" ] && CSS_VERSION="${CSS_VERSION}-${CSS_EXTRA_VERSION}"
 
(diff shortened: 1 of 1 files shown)


rhvgoyal:
@rhatdan PTAL

container-storage-setup.sh
```diff
@@ -21,40 +21,42 @@
...
+  # 'man container-storage-setup'
 
-DOCKER_ROOT_LV_NAME="docker-root-lv"
-DOCKER_ROOT_DIR="/var/lib/docker"
-DOCKER_METADATA_DIR="/var/lib/docker"
-DOCKER_ROOT_VOLUME_SIZE=40%FREE
+  DOCKER_ROOT_LV_NAME="docker-root-lv"
+  DOCKER_ROOT_DIR="/var/lib/docker"
+  DOCKER_METADATA_DIR="/var/lib/docker"
+  DOCKER_ROOT_VOLUME_SIZE=40%FREE
```
rhatdan: Can't these be overriden by the user?
@shishir-a412ed PTAL

container-storage-setup.sh
```diff
@@ -21,40 +21,42 @@
...
 
-STORAGE_IN_FILE="/etc/sysconfig/docker-storage-setup"
-STORAGE_OUT_FILE="/etc/sysconfig/docker-storage"
-STORAGE_DRIVERS="devicemapper overlay overlay2"
+  STORAGE_IN_FILE="/etc/sysconfig/docker-storage-setup"
```
rhatdan: These are set above and need to be set during getopt?

container-storage-setup.sh
```diff
@@ -1199,8 +1201,34 @@ elif [ -e /usr/share/container-storage-setup/container-storage-setup ]; then
...
+# that file.
+if [ $# -eq 2 ]; then
+  [ -e "$1" ] && source "$1"
+else
+  if [ -e "/etc/sysconfig/docker-storage-setup" ]; then
+    source /etc/sysconfig/docker-storage-setup
+  fi
+fi
+
+set_internal_gloabls
```
rhatdan: typo?

rhvgoyal: yes

container-storage-setup.sh
rhvgoyal: DOCKER_ROOT_VOLUME_SIZE can be. I think I should divide them in two parts. One which can be overridden and once which are internal. In fact, I should probably prefix internal ones with CSS_FOO.

container-storage-setup.sh
rhvgoyal: Not sure what you are referring to.

rhatdan: I looked at your code again, and it is ok.  I thought you were setting this after you set
STORAGE_IN_FILE=$1
STORAGE_OUT_FILE=$2

container-storage-setup.sh
rhatdan: Yes this is why I was adding the _ to indicate internal (In Python anyways.)

container-storage-setup.sh
rhvgoyal: I am doing getopt twice now. internal gloabls are set after first getopt. That getopt is done just to figure out which config file should be sourced.

I source the file and after that I do getopt again and that time override STORAGE_IN_FILE and STORAGE_OUT_FILE

rhvgoyal:
This is not needed any more after merge of PR #220

PR #220, opened 2017-03-01
Prefix internal global variables with underscore

rhvgoyal:
This will help differentiate between variables which are internal
and should not be overridden using container-storage-setup file
and the one which are overriddable.


rhvgoyal:
@rhatdan PTAL

rhatdan:
LGTM

rhvgoyal:
@rh-atomic-bot r+ b2786ad

PR #223, opened 2017-03-02
Do docker metadata checks only in docker compatibility mode

rhvgoyal:
We don't want to be checking for existing /var/lib/docker/metadata/ dir
for cases where inputfile and outputfile has been specified.


rhvgoyal:
@rhatdan PTAL

rhatdan:
Should we wait for @shishir-a412ed patch to just check if STORAGE_CONFIG="docker"?

That way user could specify /etc/sysconfig/docker-storage-setup with STORAGE_CONFIG="docker"

rhatdan:
Fine LGTM

rhvgoyal:
@rh-atomic-bot r+ 16334a1

PR #224, opened 2017-03-02
Use DOCKER_STORAGE_OPTIONS in output only in compatibility mode

rhvgoyal:
In generic mode, it will output STORAGE_OPTIONS. At some point of time
we should probably provide another config knob to make this prefix
configurable so that docker can make use of new generic mode and
not be limited to only compatibility mode.

### DIFF
--- a/container-storage-setup.sh
+++ b/container-storage-setup.sh
@@ -1227,7 +1227,7 @@ case $# in
 	;;
 esac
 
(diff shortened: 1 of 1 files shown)


rhvgoyal:
@rhatdan PTAL

rhatdan:
@rh-atomic-bot r+ 25fc27a

PR #225, opened 2017-03-03
Move the check_storage_options to right after the INPUT_FILE is sourced

rhatdan:
We don't want to do any setup before we verify that the users input
was correct.


rhatdan:
@shishir-a412ed @rhvgoyal PTAL

This should help cleaup @shishir-a412ed patch if the tests all pass.

rhvgoyal:
LGTM

rhvgoyal:
@rh-atomic-bot r+ 138f2ad

PR #226, opened 2017-03-03
Info message when STORAGE_DRIVER is missing.

shishir-a412ed:

### DIFF
--- a/container-storage-setup.sh
+++ b/container-storage-setup.sh
@@ -1047,7 +1047,7 @@ setup_storage() {
   local current_driver
 
   if [ "$STORAGE_DRIVER" == "" ];then
-    Info "No storage driver specified. Specify one using STORAGE_DRIVER option."
+    Info "STORAGE_DRIVER not set, no storage will be configured. You must specify STORAGE_DRIVER if you want to configure storage."
     exit 0
   fi
 


shishir-a412ed:
ping @rhvgoyal @rhatdan

container-storage-setup.sh
```diff
@@ -1047,8 +1047,7 @@ setup_storage() {
   local current_driver
 
   if [ "$STORAGE_DRIVER" == "" ];then
```
rhatdan: Can you just change this line to 
if [ -z "$STORAGE_DRIVER ];

rhvgoyal:
Actually, this is intentional. We support passing `STORAGE_DRIVER=""` so that if somebody wants to test with loop devices intentionally, docker-storage-setup can exit without setting up any storage.

rhvgoyal:
So you might want to improve message but exiting with code 0 is intentional.

rhatdan:
What should the message say?  

STORAGE_DRIVER not set, no storage will be configured. You must specify STORAGE_DRIVER if you want c-s-s to configure storage."

rhvgoyal:
How about getting rid of word c-s-s? rest looks fine.

`STORAGE_DRIVER not set, no storage will be configured. You must specify STORAGE_DRIVER if you want to configure storage."`

shishir-a412ed:
@rhvgoyal PTAL.

rhatdan:
@rh-atomic-bot r+ a548498

PR #227, opened 2017-03-03
Do not default to name docker-pool if not in docker compatibility mode

rhvgoyal:
By default name docker-pool should be used only for compatibility mode.
In new mode, user need to specify name of thin pool.

We had committed some code where name of thin pool could be default
to docker-pool even in new mode. But that code was only in upstream
tree for a very short duration. We should not have done that.

But can't carry code to support backward compatibility for that. It
becomes too messy.

Now onwards backward compatibility will be supported for docker
compatibility mode and backward compatibility for rest of the
stuff will be provided once we reach version 1.0.0. Till then
things are in a state of flux as lot of new things are being
added.

Moved CONTAINER_THINPOOL check earlier to make sure in non-compat
mode we check early that a CONTAINER_THINPOOL name has been 
passed in.

### DIFF
--- a/container-storage-setup.sh
+++ b/container-storage-setup.sh
@@ -334,10 +334,6 @@ check_min_data_size_condition() {
 }
 
 create_lvm_thin_pool () {
-  if [ "$STORAGE_DRIVER" == "devicemapper" ] && [ -z "$CONTAINER_THINPOOL" ];then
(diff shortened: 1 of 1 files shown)


rhvgoyal:
@rhatdan PTAL

rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+ b86a92b

PR #233, opened 2017-03-10
Suggest usage of /var/lib/containers/storage

giuseppe:
system containers install under /var/lib/containers/atomic and they
need to use the same file system as /ostree/repo so that hard links
from the OSTree repository can be used.  Suggest usage of
/var/lib/containers/storage to avoid this conflict.

Closes: https://github.com/projectatomic/container-storage-setup/issues/231

Change done with:

git ls-files | xargs sed -i -e"s|/var/lib/containers|/var/lib/containers/storage|"

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

### DIFF
--- a/container-storage-setup.1
+++ b/container-storage-setup.1
@@ -136,9 +136,9 @@ CONTAINER_ROOT_LV_MOUNT_PATH:
      Creates a logical volume named CONTAINER_ROOT_LV_NAME and mounts
      it at the specified path. By default no new logical volume will
      be created. For example:
-     \f[B]CONTAINER_ROOT_LV_MOUNT_PATH=/var/lib/containers\f[]
+     \f[B]CONTAINER_ROOT_LV_MOUNT_PATH=/var/lib/containers/container-runtime\f[]
      would carve out a logical volume, format it with an XFS filesystem
-     and mount it on /var/lib/containers.
+     and mount it on /var/lib/containers/container-runtime.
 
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

rhvgoyal:
What all does /var/lib/containers supposed to contain? So two container runtime are going to share this directory? say `system containers` and `crio-containers`?

IOW, multiple container runtimes are going to share /var/lib/containers/? If that's the case, shouldn't we be creating a directory for each container runtime so that runtime has exclusive ownership of that.

Say /var/lib/containers/crio  /var/lib/containers/system-containers /var/lib/containers/foo

giuseppe:
@rhvgoyal yes I agree with that, and we need another level of indirection under `/var/lib/containers`.  We probably didn't imagine there would be different kind of containers supported by atomic, so system containers got the `/var/lib/containers/atomic` namespace.  It would mess up everything though if other kind of containers are stored there as well

rhvgoyal:
Right. a container run time should get exclusive ownership of a directory under which it can setup things as it wants.

So to me we should probably change example to  `/var/lib/containers/container-runtime`

giuseppe:
@rhvgoyal yes that is better, thanks.  I've pushed a new version ⬆️

rhatdan:
@rhvgoyal Works for me.

rhvgoyal:
LGTM

rhvgoyal:
@rh-atomic-bot r+ 1ccbd55

PR #234, opened 2017-03-27
Handle partition names like "loop0p1".

shishir-a412ed:
Signed-off-by: Marius Vollmer <mvollmer@redhat.com>

### DIFF
--- a/container-storage-setup.sh
+++ b/container-storage-setup.sh
@@ -770,7 +770,7 @@ create_disk_partitions() {
     # signatures on disk or signatures should be wiped. Don't care
     # about any signatures found in the middle of disk after creating
     # partition and wipe signatures if any are found.
-    if ! wipefs -a ${dev}1; then
+    if ! wipefs -f -a ${dev}1; then
       Fatal "Failed to wipe signatures on device ${dev}1"
     fi
     pvcreate ${dev}1
--- a/container-storage-setup.sh
+++ b/container-storage-setup.sh
@@ -614,13 +614,24 @@ vg_exists() {
 }
 
 is_block_dev_partition() {
-  local bdev=$1
+  local bdev=$1 devparent
 
   if ! disktype=$(lsblk -n --nodeps --output type ${bdev}); then
     Fatal "Failed to run lsblk on device $bdev"
   fi
 
   if [ "$disktype" == "part" ];then
+    return 0
+  fi
+
+  # For loop device partitions, lsblk reports type as "loop" and not "part".
+  # So check if device has a parent in the tree and if it does, there are high
+  # chances it is partition (except the case of lvm volumes)
+  if ! devparent=$(lsblk -npls -o NAME ${bdev}|tail -n +2); then
+    Fatal "Failed to run lsblk on device $bdev"
+  fi
+
+  if [ -n "$devparent" ];then
     return 0
   fi
 
@@ -692,10 +703,9 @@ scan_disks() {
   local new_disks=""
 
   for dev in $disk_list; do
-    local basename=$(basename $dev)
-    local p
+    local part=$(dev_query_first_child $dev)
 
-    if is_dev_part_of_vg ${dev}1 $vg; then
+    if [ -n "$part" ] && is_dev_part_of_vg ${part} $vg; then
       Info "Device ${dev} is already partitioned and is part of volume group $VG"
       continue
     fi
@@ -708,8 +718,7 @@ scan_disks() {
     fi
 
     # If device does not have partitions, it is a new disk requiring processing.
-    p=$(awk "\$4 ~ /${basename}./ {print \$4}" /proc/partitions)
-    if [[ -z "$p" ]]; then
+    if [[ -z "$part" ]]; then
       new_disks="$dev $new_disks"
       continue
     fi
@@ -731,7 +740,7 @@ create_partition_sfdisk(){
     cat <<EOF | sfdisk $dev
 unit: sectors
 
-${dev}1 : start=     2048, size=  ${size}, Id=8e
+start=     2048, size=  ${size}, Id=8e
 EOF
 }
 
@@ -741,7 +750,7 @@ create_partition_parted(){
 }
 
 create_partition() {
-  local dev="$1"
+  local dev="$1" part
 
   if [ -x "/usr/sbin/parted" ]; then
     create_partition_parted $dev
@@ -755,26 +764,33 @@ create_partition() {
     Fatal "udevadm settle after partition creation failed. Exiting."
   fi
 
-  if ! wait_for_dev ${dev}1; then
-    Fatal "Partition device ${dev}1 is not available"
+  part=$(dev_query_first_child $dev)
+
+  if ! wait_for_dev ${part}; then
+    Fatal "Partition device ${part} is not available"
   fi
 }
 
+dev_query_first_child() {
+  lsblk -npl -o NAME "$1" | tail -n +2 | head -1
+}
+
 create_disk_partitions() {
-  local devs="$1"
+  local devs="$1" part
 
   for dev in $devs; do
     create_partition $dev
+    part=$(dev_query_first_child $dev)
 
     # By now we have ownership of disk and we have checked there are no
     # signatures on disk or signatures should be wiped. Don't care
     # about any signatures found in the middle of disk after creating
     # partition and wipe signatures if any are found.
-    if ! wipefs -f -a ${dev}1; then
-      Fatal "Failed to wipe signatures on device ${dev}1"
+    if ! wipefs -f -a ${part}; then
+      Fatal "Failed to wipe signatures on device ${part}"
     fi
-    pvcreate ${dev}1
-    _PVS="$_PVS ${dev}1"
+    pvcreate ${part}
+    _PVS="$_PVS ${part}"
   done
 }
 
--- a/tests/015-test-fail-loop-to-create-thin-pool.sh
+++ b/tests/015-test-fail-loop-to-create-thin-pool.sh
@@ -1,0 +1,62 @@
+source $SRCDIR/libtest.sh
+
+test_fail_if_loop_partition_passed() {
+  local lbdevice tmpfile
+  local test_status=1
+  local testname=`basename "$0"`
+  local vg_name="css-test-foo"
+  local tmplog=${WORKDIR}/tmplog
+  local errmsg="Partition specification unsupported at this time."
+
+  # Error out if any pre-existing volume group vg named css-test-foo
+  if vg_exists "$vg_name"; then
+    echo "ERROR: $testname: Volume group $vg_name already exists." >> $LOGS
+    return $test_status
+  fi
+
+  # Create loopback device.
+   tmpfile=$(mktemp /tmp/c-s-s.XXXXXX)
+   truncate --size=6G $tmpfile
+   lbdevice=$(losetup -f)
+   losetup --partscan $lbdevice $tmpfile
+   if ! create_partition $lbdevice;then
+      echo "ERROR: Failed partitioning $lbdevice"
+      cleanup_loop_device "$tmpfile" "$lbdevice"
+      return $test_status
+   fi
+
+  cat << EOF > /etc/sysconfig/docker-storage-setup
+DEVS="${lbdevice}p1"
+VG=$vg_name
+EOF
+
+  # Run container-storage-setup
+  $CSSBIN > $tmplog 2>&1
+  rc=$?
+  cat $tmplog >> $LOGS 2>&1
+
+  # Test failed.
+  if [ $rc -ne 0 ]; then
+      if grep --no-messages -q "$errmsg" $tmplog; then
+	  test_status=0
+      else
+	  echo "ERROR: $testname: $CSSBIN Failed for a reason other then \"$errmsg\"" >> $LOGS
+      fi
+  else
+      echo "ERROR: $testname: $CSSBIN Succeeded. Should have failed since ${lbdevice}p1 is a loop device partition." >> $LOGS
+  fi
+  cleanup $vg_name "$lbdevice"
+  cleanup_loop_device "$tmpfile" "$lbdevice"
+  return $test_status
+}
+
+cleanup_loop_device() {
+  local tmpfile=$1
+  local lbdevice=$2
+  losetup -d $lbdevice
+  rm $tmpfile > /dev/null 2>&1
+}
+
+# Make sure command fails if loop device partition /dev/loop0p1 is passed.
+test_fail_if_loop_partition_passed
+
--- a/tests/115-test-fail-loop-to-create-thin-pool.sh
+++ b/tests/115-test-fail-loop-to-create-thin-pool.sh
@@ -1,0 +1,65 @@
+source $SRCDIR/libtest.sh
+
+test_fail_if_loop_partition_passed() {
+  local lbdevice tmpfile
+  local test_status=1
+  local testname=`basename "$0"`
+  local vg_name="css-test-foo"
+  local infile=${WORKDIR}/container-storage-setup
+  local outfile=${WORKDIR}/container-storage
+  local tmplog=${WORKDIR}/tmplog
+  local errmsg="Partition specification unsupported at this time."
+
+  # Error out if any pre-existing volume group vg named css-test-foo
+  if vg_exists "$vg_name"; then
+    echo "ERROR: $testname: Volume group $vg_name already exists." >> $LOGS
+    return $test_status
+  fi
+
+  # Create loopback device.
+   tmpfile=$(mktemp /tmp/c-s-s.XXXXXX)
+   truncate --size=6G $tmpfile
+   lbdevice=$(losetup -f)
+   losetup --partscan $lbdevice $tmpfile
+   if ! create_partition $lbdevice;then
+      echo "ERROR: Failed partitioning $lbdevice"
+      cleanup_loop_device "$tmpfile" "$lbdevice"
+      return $test_status
+   fi
+
+  cat << EOF > $infile
+DEVS="${lbdevice}p1"
+VG=$vg_name
+CONTAINER_THINPOOL=container-thinpool
+EOF
+
+  # Run container-storage-setup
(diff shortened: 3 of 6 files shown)


shishir-a412ed:
@rhvgoyal The tests are passing locally on my system.

http://pastebin.test.redhat.com/468919
Need to look why it is failing on CI ?

@jlebon Any idea ?

Shishir

shishir-a412ed:
````
Running test ./tests/001-test-use-devs-to-create-thin-pool.sh
INFO: Device node /dev/vdb1 exists.
wipefs: error: Checking: probing initialization failed: No such file or directory
ERROR: Failed to wipe signatures on device Checking that no-one is using this disk right now ... OK
````

Seems like the device `/dev/vdb` is already partitioned. Any idea why `/dev/vdb1` already exists ?

/cc @jlebon 

Shishir

File: container-storage-setup.sh [hunk 1]
```diff
@@ -731,40 +729,47 @@ create_partition_parted(){
...
-  local dev="$1"
+  local dev="$1" part
 
   if [ -x "/usr/sbin/parted" ]; then
     create_partition_parted $dev
   else
     create_partition_sfdisk $dev
   fi
 
+  part=$(query_first_child $dev)
```
mvollmer: If this succeeds, then the partition exists and there is no point in calling `udevadm settle` or `wait_for_dev` below.  If you are worried that you need to wait for the partition block device to come into existence after calling `sfdisk`, then you should probably call `query_first_child` in a loop until it returns something non-empty.

However, the device for the partition does exist after calling `sfdisk` so there is no need to wait for it to appear.  What does happen, though, is that udevd asynchronously calls the BLKRRPART ioctl some time after the partition has been created, and because of the way BLKRRPART works, there is a brief period of time where the block device for the partition disappears.  If you want to protect against that, I recommend wrapping the whole `sfdisk; pvcreate; vgextend` operation in `flock -s $dev`.  The shared lock signals to udevd not to do the BLKRRPART ioctl.

mvollmer: Also, if you think `udev settle` is needed, then you should call `query_first_child` after calling it, not before.

container-storage-setup.sh
```diff
@@ -731,40 +729,47 @@ create_partition_parted(){
...
+  lsblk -npl -o NAME "$1" | tail -n +2 | head -1
 }
 
 create_disk_partitions() {
-  local devs="$1"
+  local devs="$1" part
 
   for dev in $devs; do
-    create_partition $dev
+    part=$(create_partition $dev)
```
mvollmer: This will not work when `sfdisk` is used, I think.  `sfdisk` produces a lot of extra output on stdout.

mvollmer:
What about removing all partition creation code from container-storage-setup?

container-storage-setup.sh
mvollmer: We can probably drop support for `sfdisk`, no?

mvollmer:
```
wipefs: error: Checking: probing initialization failed: No such file or directory
ERROR: Failed to wipe signatures on device Checking that no-one is using this disk right now ... OK
```

This might be the issue I pointed out about `sfdisk` outputting a lot on stdout.  I think ${part} ends up being all of `Checking that no-one is using this disk right now ... OK....` and wipefs tries to use that as the device name.

container-storage-setup.sh
shishir-a412ed: We cannot drop support for `sfdisk`, that would break backward compatibility. Users who don't have `parted` will fall back to `sfdisk` for partitioning and we need to support that.

shishir-a412ed:
@mvollmer Thanks for pointing it out. `sfdisk` output might be causing problems. Let me try to remove `parted` from my machine, then `container-storage-setup` would fallback to `sfdisk` and I can reproduce the issue.

shishir-a412ed:
`What about removing all partition creation code from container-storage-setup?`

@mvollmer I will open a separate PR for that.

container-storage-setup.sh
shishir-a412ed: Okay makes sense. The `BLKRRPART` ioctl is a separate issue, I will drop that from the scope of this PR. @rhvgoyal Any specific reason you introduced `udevadm settle` and `wait_for_dev` ? Since if `create_partition_sfdisk` or `create_partition_parted` is successful, the device for the partition will exist. So we can just call `query_first_child()` and extract the partition name.

shishir-a412ed:
@rhvgoyal PTAL.
I have rebase and resolved the conflicts against `upstream/master`, not sure why CI is not running through.

````
[smahajan@localhost docker-storage-setup]$ git branch
  docker_root_volume_test2
  dss_copy
  error_storage_driver
* handle_partition_names
  master
[smahajan@localhost docker-storage-setup]$ git pull --rebase upstream master
From github.com:projectatomic/docker-storage-setup
 * branch            master     -> FETCH_HEAD
Current branch handle_partition_names is up to date.
````

Shishir

File: container-storage-setup.sh [hunk 3]
```diff
@@ -755,26 +756,33 @@ create_partition() {
...
```
rhvgoyal: how about querying partition after creation. Instead of returning it from create_partition. 

Anyway, create_partition() itself does not know better as it also is querying the partition name after creating it.

Downside is that we will query twice. But it removes the need of capturing the stdout of create_partition() and need of supressing sfdisk output. I like that better.

container-storage-setup.sh
```diff
@@ -755,26 +756,33 @@ create_partition() {
...
+  part=$(query_first_child $dev)
+
+  if ! wait_for_dev ${part}; then
+    Fatal "Partition device ${part} is not available"
   fi
+  echo $part
+}
+
+query_first_child() {
+  lsblk -npl -o NAME "$1" | tail -n +2 | head -1
```
rhvgoyal: How about doing 
`lsblk -npl -o NAME "$1" | head -n 2 | tail -1`

container-storage-setup.sh
```diff
@@ -755,26 +756,33 @@ create_partition() {
...
-    Fatal "Partition device ${dev}1 is not available"
+  part=$(query_first_child $dev)
+
+  if ! wait_for_dev ${part}; then
+    Fatal "Partition device ${part} is not available"
   fi
+  echo $part
+}
+
+query_first_child() {
```
rhvgoyal: lets call this dev_query_first_child()

container-storage-setup.sh
```diff
@@ -721,18 +719,21 @@ scan_disks() {
...
-  local dev="$1" size
+  local dev="$1" size output
   # Use a single partition of a whole device
   # TODO:
   #   * Consider gpt, or unpartitioned volumes
   #   * Error handling when partition(s) already exist
   #   * Deal with loop/nbd device names. See growpart code
   size=$(( $( awk "\$4 ~ /"$( basename $dev )"/ { print \$3 }" /proc/partitions ) * 2 - 2048 ))
-    cat <<EOF | sfdisk $dev
+  if ! output=$(cat <<EOF | sfdisk $dev
```
rhvgoyal: we can get rid of capturing output in a variable and supressing it once you stop returning partition name from create_partition()

container-storage-setup.sh
shishir-a412ed: Makes sense. Fixed.

container-storage-setup.sh
shishir-a412ed: Fixed.

container-storage-setup.sh
shishir-a412ed: Fixed.

shishir-a412ed:
@rhvgoyal PTAL.

container-storage-setup.sh
```diff
@@ -755,26 +753,33 @@ create_partition() {
...
-    Fatal "Partition device ${dev}1 is not available"
+  part=$(dev_query_first_child $dev)
+
+  if ! wait_for_dev ${part}; then
+    Fatal "Partition device ${part} is not available"
   fi
 }
 
+dev_query_first_child() {
+  lsblk -npl -o NAME "$1" | tail -n +2 | head -1
```
rhvgoyal: I had given a comment on this to simplify it. can you please take care of that as well.

shishir-a412ed: @rhvgoyal https://github.com/projectatomic/container-storage-setup/pull/234#discussion_r109312810

container-storage-setup.sh
rhvgoyal: I was thinking that it is easier to understand. If we are querying first child, then take first two lines of output. That will have parent and first child. 

tail -n +2, will return you all children. So I think it is simpler to do just first two lines?

shishir-a412ed: Both of them seems exactly the same to me :). `tail -n +2` will return everything except the device and then you take the first partition out of it. e.g.
````
/dev/vdb
/dev/vdb1
/dev/vdb2
/dev/vdb3.
````
`tail` command will return everything except the device (`/dev/vdb`) and then you pick up the first partition using `head` which is `/dev/vdb1`. If your format seems simpler to you, I will change it.

**Comment by shishir-a412ed** [comment]
@rhvgoyal PTAL.

**Comment by shishir-a412ed** [comment]
@rhvgoyal Actually the tests are failing on this new format
https://github.com/projectatomic/container-storage-setup/pull/234#discussion_r109229653
in `dev_query_first_child()` so we need to go back to the original command which was working.

File: container-storage-setup.sh (hunk 4 above)
Comment (rhvgoyal) [inline container-storage-setup.sh]: ok, we can leave it as it is. I think I was much more familiar with head -2. For tail -n +2, I had to read the man page and that's one reason I thought it was simpler. 

No big deal. I will test it.

**Comment by rhvgoyal** [comment]
@shishir-a412ed one more minor nit. Try passing in /dev/loop0p1 as input and see what happens.

`DEVS=/dev/loop0p1`

lsblk does not report it as partitioned device. So we might have to do something additional. May be for device type loop, we can check if partition name ends with p*. Or try to do something by scanning /proc/partitions.

**Comment by shishir-a412ed** [comment]
@rhvgoyal Right. the `type` of a loop block device partition is `loop` and not `part`. As an experiment I partitioned a loop block device (`/dev/loop0`) and a regular VM Disk (`/dev/vdb`) on my system using `parted`.

````
sh-4.3# lsblk
NAME            MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
sr0              11:0    1 1024M  0 rom  
vda             252:0    0   40G  0 disk 
├─vda1          252:1    0  500M  0 part /boot
└─vda2          252:2    0 39.5G  0 part 
  ├─fedora-root 253:0    0 37.5G  0 lvm  /
  └─fedora-swap 253:1    0    2G  0 lvm  [SWAP]
vdb             252:16   0   20G  0 disk 
└─vdb1          252:17   0   20G  0 part 
loop0             7:0    0    5G  0 loop 
└─loop0p1       259:0    0    5G  0 loop 
````

As you can see `/dev/vdb1` has type `part` and `/dev/loop0p1` has type `loop`.
Right now there is already a check in `is_block_dev_partition()` to check if the block device was already partitioned or not.
````
  if [ "$disktype" == "part" ];then
    return 0
  fi
 ````
We can add a similar check to handle loop block devices partition. Something like
````
If [ "$disktype == "loop" && (Partition name ends with p*) ];then
   return 0
fi
````

**Comment by shishir-a412ed** [comment]
@rhvgoyal PTAL.

File: container-storage-setup.sh [hunk 8]
```diff
@@ -624,6 +624,15 @@ is_block_dev_partition() {
     return 0
   fi
 
+  # This will check if a loop/nvme device is passed and is already partitioned.
```
rhvgoyal: Lets mention something like.

` For loop device partitions, lsblk reports type as "loop" and not "part". So check if device has a parent in the tree and if it does, there are high chances it is partition (except the case of lvm volumes)"

container-storage-setup.sh
```diff
@@ -624,6 +624,15 @@ is_block_dev_partition() {
...
+  if ! devicePartition=$(lsblk -npls -o NAME ${bdev}|tail -n +2); then
```
rhvgoyal: how about calling it "devparent". And also make it `local` variable to funciton.

container-storage-setup.sh
shishir-a412ed: Fixed.

container-storage-setup.sh
shishir-a412ed: Fixed.

rhvgoyal:
Give me few minutes to finish review before you push new patch, please.

shishir-a412ed:
@rhvgoyal ack.

rhvgoyal:
@shishir-a412ed seeing some strange behavior. Try this.

- Create loop device
- Create partition
- Delete partition
- Use DEVS=/dev/loop0 and run container-storage-setup

I get following.

[root@vm7-f25 rhvgoyal-docker-storage-setup]# ./container-storage-setup.sh 
INFO: Device node /dev/loop0p1 exists.
wipefs: /dev/loop0p1: ignoring nested "dos" partition table on non-whole disk device
wipefs: /dev/loop0p1: ignoring nested "PMBR" partition table on non-whole disk device
wipefs: Use the --force option to force erase.
WARNING: dos signature detected on /dev/loop0p1 at offset 510. Wipe it? [y/n]:

rhvgoyal:
Can we also add two tests.

- One test to make sure this works with loop devices
- Second test to make sure passing in partition loop device fails.

shishir-a412ed:
@rhvgoyal PTAL.

container-storage-setup.sh
```diff
@@ -755,26 +764,33 @@ create_partition() {
...
     create_partition $dev
+    part=$(dev_query_first_child $dev)
 
     # By now we have ownership of disk and we have checked there are no
     # signatures on disk or signatures should be wiped. Don't care
     # about any signatures found in the middle of disk after creating
     # partition and wipe signatures if any are found.
-    if ! wipefs -a ${dev}1; then
-      Fatal "Failed to wipe signatures on device ${dev}1"
+    if ! wipefs -f -a ${part}; then
```
rhvgoyal: Lets keep this in a separate patch.

shishir-a412ed: Fixed.

container-storage-setup.sh
```diff
@@ -770,7 +770,7 @@ create_disk_partitions() {
     # signatures on disk or signatures should be wiped. Don't care
     # about any signatures found in the middle of disk after creating
     # partition and wipe signatures if any are found.
-    if ! wipefs -a ${dev}1; then
+    if ! wipefs -f -a ${dev}1; then
```
rhvgoyal: We need to improve changelog. How about following.

`After partition creation, we wipe any signatures on the partition using wipefs. For some reason, for loop devices it was detecting nested partition table signature and wipefs will skip it by default and instead ask user whether to clean it or not. 

Use option -f instead to not ask user the question and delete it alaways.
`

shishir-a412ed: s/alaways/always.

shishir-a412ed: Fixed.

shishir-a412ed:
@rhvgoyal PTAL.

tests/015-test-fail-loop-to-create-thin-pool.sh
```diff
@@ -0,0 +1,60 @@
...
+    return $test_status
+  fi
+
+  # Create loopback device.
+   tmpfile=$(mktemp /tmp/c-s-s.XXXXXX)
+   truncate --size=6G $tmpfile
+   lbdevice=$(losetup -f)
+   losetup --partscan $lbdevice $tmpfile
+   if ! create_partition $lbdevice;then
+      echo "ERROR: Failed partitioning $lbdevice"
```
rhvgoyal: If we fail to partition the device, we should fail test and return with status 1 here, isn't it?

tests/115-test-fail-loop-to-create-thin-pool.sh
```diff
@@ -0,0 +1,63 @@
...
+    return $test_status
+  fi
+
+  # Create loopback device.
+   tmpfile=$(mktemp /tmp/c-s-s.XXXXXX)
+   truncate --size=6G $tmpfile
+   lbdevice=$(losetup -f)
+   losetup --partscan $lbdevice $tmpfile
+   if ! create_partition $lbdevice;then
+      echo "ERROR: Failed partitioning $lbdevice"
```
rhvgoyal: Same here, if we fail to partition device, we should fail test and return 1 here.

tests/115-test-fail-loop-to-create-thin-pool.sh
```diff
@@ -0,0 +1,63 @@
...
+  cleanup $vg_name "$lbdevice" "$infile" "$outfile"
+  cleanup_loop_device "$tmpfile" "$lbdevice"
+  return $test_status
+}
+
+cleanup_loop_device() {
+  local tmpfile=$1
+  local lbdevice=$2
+  rm $tmpfile > /dev/null 2>&1
+  losetup -d $lbdevice
```
rhvgoyal: This applies to all the four places you have used this in tests.

rhvgoyal:
@shishir-a412ed Patches look more or less good to me. Gave some comments just now. Can you please take care of these and then I should be able to merge patches.

tests/115-test-fail-loop-to-create-thin-pool.sh
shishir-a412ed: Fixed.

tests/115-test-fail-loop-to-create-thin-pool.sh
shishir-a412ed: Fixed.

tests/015-test-fail-loop-to-create-thin-pool.sh
shishir-a412ed: Fixed.

shishir-a412ed:
@rhvgoyal PTAL.

rhvgoyal:
LGTM. Thanks.

rhvgoyal:
@rh-atomic-bot r+ 7ce2091

rhvgoyal:
@rh-atomic-bot r+ 7ce2091

PR #237, opened 2017-03-29
Miscellaneous cleanup 2

rhvgoyal:
Few more patches for cleanup. These primarily pass arguments as parameters to bunch of functions instead of relying on global variables. This make is it easy to reuse these functions.


rhvgoyal:
@rhatdan PTAL

rhatdan:
:+1:

rhatdan:
@rh-atomic-bot r+ 608d339

PR #240, opened 2017-04-11
Fix: 002-test-reject-disk-with-lvm-signature.sh

shishir-a412ed:


shishir-a412ed:
Right now when I run test `002-test-reject-disk-with-lvm-signature.sh` on upstream/master branch I see the following errors in the logs.

````
Running test ./002-test-reject-disk-with-lvm-signature.sh
  Physical volume "/dev/vdb" successfully created.
ERROR: Found LVM2_member signature on device /dev/vdb at offset 0x218. Wipe signatures using wipefs or use WIPE_SIGNATURES=true and retry.
  Volume group "css-test-foo" not found
  Cannot process volume group css-test-foo
  Please enter a physical volume path
  Run `pvremove --help' for more information.
Error: /dev/vdb: unrecognised disk label
/dev/vdb: 8 bytes were erased at offset 0x00000218 (LVM2_member): 4c 56 4d 32 20 30 30 31
````

This PR fixes that problem.

ping @rhvgoyal 
PS: I will open a separate PR for test `102-test-reject-disk-with-lvm-signature.sh` which has same issue.

**Comment by rhvgoyal** [comment]
@shishir-a412ed can you rebase this PR on top of latest tree.

**Comment by shishir-a412ed** [comment]
@rhvgoyal PTAL.

File: tests/002-test-reject-disk-with-lvm-signature.sh [hunk 1]
```diff
@@ -25,17 +27,33 @@ EOF
...
+          test_status=0
+      else
+          echo "ERROR: $testname: $CSSBIN Failed for a reason other then \"$errmsg\"" >> $LOGS
+      fi
+  else
+      echo "ERROR: $testname: $CSSBIN Succeeded. Should have failed since LVM2_member signature exists on devices $devs" >> $LOGS
+  fi
 
-  cleanup $vg_name "$devs"
+  cleanup_pvs $devs
```
rhvgoyal: We should continue to call cleanup(), right? In case there is a bug in tool and it has done something with device and added it to volume group etc, we need to clean all that state.

So, IMO, we need to call both cleanup() and cleanup_pvs().

tests/002-test-reject-disk-with-lvm-signature.sh
```diff
@@ -25,17 +27,33 @@ EOF
...
   return $test_status
 }
 
+cleanup_pvs() {
```
rhvgoyal: cleanup_pvs_sig()

rhvgoyal:
@shishir-a412ed Can you please take care of minor comments.

tests/002-test-reject-disk-with-lvm-signature.sh
shishir-a412ed: For this particular test, we fail right away since there are signatures on the block device.
So ideally no volume group (VG) should exist. If I call both `cleanup()` and `cleanup_pvs()`.

1) I will see same errors in the logs like `Cannot process volume group css-test-foo` since volume group is not present.
2) There will be duplication of code in `cleanup()` and `cleanup_pvs()` e.g both are calling `wipe_signatures`.

How I implemented was: `cleanup_pvs()` is a condensed version of `cleanup()` which is sufficient just for this test.

Shishir

tests/002-test-reject-disk-with-lvm-signature.sh
shishir-a412ed: Fixed.

tests/002-test-reject-disk-with-lvm-signature.sh
rhvgoyal: @shishir-a412ed What if code is buggy and it does not recognize the signature on disk and continues to create partition and add it to volume group?

shishir-a412ed:
@rhvgoyal PTAL.

tests/002-test-reject-disk-with-lvm-signature.sh
shishir-a412ed: @rhvgoyal Then we have a much bigger problem than the test :)
Let me come to your desk.

rhvgoyal:
I think we can modify remove_pvs() to remove pvs both from partition and disk if pv exists. So we can implement a function pv_exists() and if pv exists, remove it. And then do this check both on partition as well device.

Same thing we can do for volume group. check for vg_exists() and if volume group exists, remove it. That will get rid of error.

shishir-a412ed:
@rhvgoyal PTAL.

tests/libtest.sh
```diff
@@ -50,6 +50,19 @@ vg_exists() {
   return 1
 }
 
+# Tests if the physical volume pv_name exists
+pv_exists() {
```
rhvgoyal: @shishir-a412ed I just implemented pv_exists() in libcss.sh. How about that implementation?

shishir-a412ed: Fixed.

shishir-a412ed:
@rhvgoyal PTAL.

rhatdan:
@rhvgoyal We should either close these pull requests or rework them or merge them.

PR #241, opened 2017-04-17
Use profile_dir setting used by LVM2 during command execution

thescriptbender:
Use profile_dir setting used by LVM2 during command execution
    
"lvm dumpconfig" is currently used to determine profile_dir location.
But if profile_dir is not explicitly defined in /etc/lvm/lvm.conf, then
the default value used by lvm will not be listed.
Hence use "lvm dumpconfig --type full" instead.
    
Signed-off-by: Carson Blake Smith <smith.carson19@gmail.com>

### DIFF
--- a/container-storage-setup.sh
+++ b/container-storage-setup.sh
@@ -910,7 +910,7 @@ enable_auto_pool_extension() {
   local profileDir
   local tmpFile=`mktemp -p /run -t tmp.XXXXX`
 
-  profileDir=$(lvm dumpconfig | grep "profile_dir" | cut -d "=" -f2 | sed 's/"//g')
+  profileDir=$(lvm dumpconfig --type full | grep "profile_dir" | cut -d "=" -f2 | sed 's/"//g')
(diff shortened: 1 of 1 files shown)


rhvgoyal:
Can you make commit message a little cleaner. Something like.

`Currently we use "lvm dumpconfing" to determine profile_dir location. But if profile_dir is not mentioned in /etc/lvm/lvm.conf, then default value used by lvm will not be listed. Hence use "lvm dumpconfig --type full" instead`

rhvgoyal:
Also please put a "Signed-off-by" message in the commit.

thescriptbender:
Sure. Although mine was clear, I spent a lot of text over-justifying.

Will go for something like your concise message, with a little cleanup.
Also amending with my signoff as well.

rhvgoyal:
LGTM

rhvgoyal:
@rh-atomic-bot r+ 	b847fe3

PR #243, opened 2017-04-18
Remove Volume Group and disks when a configuration is being removed

rhvgoyal:
container-storage-setup remove command should also remove the volume group and disks if container-storage-setup created volume group. This patch series does some cleanup and adds that functionality.

### DIFF
--- a/container-storage-setup.sh
+++ b/container-storage-setup.sh
@@ -876,6 +876,38 @@ create_disk_partitions() {
   done
 }
 
+remove_partition() {
+  local dev="$1"
+
+  if [ -x "/usr/sbin/parted" ]; then
+    parted "$dev" rm 1 >/dev/null
+  else
+    sfdisk --delete "$dev" 1 >/dev/null
+  fi
+}
+
+# Remove disk pvs and partitions. This is called in reset storage path.
+# If partition or pv does not exist, it will still return success. Error
+# will be returned only if pv or partition exists and removal fails.
+remove_disk_pvs_parts() {
+  local devs="$1" part
+
+  for dev in $devs; do
+    part=$(dev_query_first_child $dev)
+    [ -z "$part" ] && continue
+
+    if ! remove_pv_if_exists $part; then
+      Error "Failed to remove physical volume label on device $part"
+      return 1
+    fi
+
+    if ! remove_partition $dev; then
+      Error "Failed to remove partition on device $dev"
+      return 1
+    fi
+  done
+}
+
 create_extend_volume_group() {
   if [ -z "$_VG_EXISTS" ]; then
     vgcreate $VG $_PVS
@@ -1790,7 +1822,7 @@ reset_extra_volume() {
 
 # Remove command processing
 reset_storage() {
-  local resolved_path
+  local resolved_path dev
 
   # Populate $_RESOLVED_MOUNT_DIR_PATH
   if [ -n "$_M_CONTAINER_ROOT_LV_MOUNT_PATH" ];then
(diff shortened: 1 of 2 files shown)


rhvgoyal:
@rhatdan @shishir-a412ed PTAL

rhvgoyal:
@jlebon tests can't seem to run.

container-storage-setup.sh
```diff
@@ -888,6 +876,38 @@ create_disk_partitions() {
...
+  if [ -x "/usr/sbin/parted" ]; then
+    parted "$dev" rm 1
+  else
+    sfdisk --delete "$dev" 1
+  fi
+}
+
+# Remove disk pvs and partitions. This is called in reset storage path.
+# If partition or pv does not exist, it will still return success. Error
+# willl be returned only if pv or partition existst and removal fails.
```
rhatdan: +# willl be returned only if pv or partition existst and removal fails.
SPELLING =========================^

container-storage-setup.sh
```diff
@@ -888,6 +876,38 @@ create_disk_partitions() {
...
+remove_disk_pvs_parts() {
+  local devs="$1" part
+
+  for dev in $devs; do
+    part=$(dev_query_first_child $dev)
+    [ -z "$part" ] && continue
+
+    if ! remove_pv_if_exists $part; then
+      Error "Failed to remove physical volume label on device $part"
+      return 1
```
rhatdan: Should we fail right away or continue removing others and only error out at the end?

container-storage-setup.sh
```diff
@@ -1589,14 +1609,21 @@ activate_extra_lv_fs() {
...
   local metafile_path="$_CONFIG_DIR/$_CONFIG_NAME/$_METAFILE_NAME"
+  local status_file="$_CONFIG_DIR/$_CONFIG_NAME/$_STATUSFILE_NAME"
+  local curr_status
 
   [ ! -d "$_CONFIG_DIR/$_CONFIG_NAME" ] && Fatal "Storage configuration $_CONFIG_NAME does not exist"
 
   [ ! -e "$metafile_path" ] && Fatal "Storage configuration $_CONFIG_NAME metadata does not exist"
   source "$metafile_path"
 
+  curr_status=`cat $status_file`
```
rhatdan: We should probably start using 

``` curr_status=$(cat $status_file)```

I think that is the preferred way to handle this now.

Comment (rhatdan) [inline container-storage-setup.sh]: What happens if status_file does not exist?

File: container-storage-setup.sh [hunk 4]
```diff
@@ -1814,6 +1841,19 @@ reset_storage() {
...
 
+  # If we created a volume group, remove volume group.
+  if [ "$_M_VG_CREATED" == "1" ];then
+    if ! remove_vg_if_exists "$_M_VG"; then
+      Error "Failed to remove volume group $_M_VG"
+      return 1
+    fi
+
+    # Cleanup any disks we added to volume group.
+    if ! remove_disk_pvs_parts "$_M_DEVS_RESOLVED";then
```
rhatdan: Do errors get logged at the lower level here?

container-storage-setup.sh
rhvgoyal: will fix.

container-storage-setup.sh
rhvgoyal: I think we should fail right away. User can try again after fixing the issue.

rhvgoyal: And we can change it later if there is a good use case. Right now model is that on first failure bail out. We probably should implement a switch say --force. That switch should do whatever cleanup is possible. 

`dmsetup remove -f foo`

And with that switch it will make sense to continue to remove pvs from other disks.

rhvgoyal: We can implement that switch later. As that also leaves resources behind in bad state.

container-storage-setup.sh
rhvgoyal: good point. I should handle exit code of cat. So that if status file does not exist, we should error out early.

container-storage-setup.sh
rhvgoyal: No, right now I don't have a separate log file where errors can go. For now I have sent everything to /dev/null.

So where should the errors from lvm commands go. Or should I just redirect stdout to /dev/null and let errors go to stderr?

container-storage-setup.sh
rhatdan: SGTM

container-storage-setup.sh
rhatdan: I would let errors go to stderr.  Otherwise you are going to get a failure and have no idea why.

rhvgoyal: "parted" sends messages like this on stderr while removing partition.

"Information: You may need to update /etc/fstab."

This is ugly.

rhvgoyal:
@rhatdan took care of your comments. PTAL.

- Fixed typos
- Handled the case of status file missing
- supressed stdout but not stderr

rhatdan:
@rh-atomic-bot r+ d84cd48

PR #244, opened 2017-06-01
Do not hardcode 'parted' path

xdrus:
sfdisk failed on some disk, so I would prefer to use parted. However, it is not always installed to /usr/sbin (for example Amazon Linus install it /sbin).

### DIFF
--- a/container-storage-setup.sh
+++ b/container-storage-setup.sh
@@ -78,6 +78,9 @@ _STORAGE_OPTIONS="STORAGE_OPTIONS"
 # Keeps track of if we created a volume group or not.
 _VG_CREATED=
 
+# Do we have parted
+_PARTED=$(which parted || echo "")
+
 get_docker_version() {
   local version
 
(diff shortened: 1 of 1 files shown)


rhatdan:
bot, add author to whitelist

rhatdan:
LGTM
@rhvgoyal PTAL

container-storage-setup.sh
```diff
@@ -78,6 +78,9 @@ _STORAGE_OPTIONS="STORAGE_OPTIONS"
 # Keeps track of if we created a volume group or not.
 _VG_CREATED=
 
+# Do we have parted
+_PARTED=$(which parted || echo "")
```
rhvgoyal: There is no need to create a global variable for this information. Its not re-used outside a single function. How about creating a helper function say `has_parted` and use that to decide whether to use parted for sfdisk.

if has_parted; then
   use_parted
else
  use_sfdisk

jhaynes: `command` is a good option: `if command -v parted > /dev/null 2>&1 ; then ...`

rhvgoyal:
Did not get any response to my comments. Closing it. Once somebody cares enough, reopen with updated patch.

PR #247, opened 2017-07-05
Release version 0.6.0

rhvgoyal:
Bumped version from 0.3.0 to 0.6.0. I realized that in the past we already
released some versions like v0.5. So to avoid conflict with previous versions
bump it to a higher number which has never been released.


rhvgoyal:
@rhatdan PTAL

PR #248, opened 2017-07-06
add MAX_DATA_SIZE option

jason-liew:
Some sever have a large disk like 2T, but only need at most 200G for docker to store images and containers. 
And then create logical volume  with rest free space as host path volume to be used with  -v option in docker run command .
Add new MAX_DATA_SIZE option may be simpler than calculating DATA_SIZE for every SERVER.

### DIFF
--- a/container-storage-setup.conf
+++ b/container-storage-setup.conf
@@ -59,6 +59,20 @@ DATA_SIZE=40%FREE
 #
 MIN_DATA_SIZE=2G
 
+# MAX_DATA_SIZE specifies the maximum size of data volume otherwise pool
+# creation fails.
+#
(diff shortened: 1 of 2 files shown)


rhatdan:
bot, add author to whitelist

rhatdan:
LGTM
@rhvgoyal PTAL

rhvgoyal:
I don't think this is required. 

- First of all, we don't consume all the free space. We use 40% of free space and leave rest for others to use.
- If you are worried that disk is too big and even 40% is a lot, then you can reduce that percentage by reducing DATA_SIZE=20%.

So DATA_SIZE is already doing what you need. Just that its in terms of percentage and not an absolute number.

PR #253, opened 2017-08-16
Add a knob POOL_META_SIZE

rhvgoyal:
Right now thin pool metadata size is automatically determined. That is .1%
of free space in volume group.

A user has asked for a knob to make it configurable. They want to allocate
a larger metadata volume to start with. Looks like they run out of free space
in metadata volume and VG is full and there is not enough space to grow it
later.

Given their worklaod, they can allocate a bigger metadata volume using
this knob at the beginning.


rhvgoyal:
@rhatdan PTAL

rhatdan:
@rh-atomic-bot r+ 0754316

PR #262, opened 2017-11-02
A try at try-overlay

cgwalters:


container-storage-setup.sh
```diff
@@ -437,6 +437,46 @@ check_docker_storage_metadata() {
...
+             fi
+             ;;
+       ext4) # I'm just going to assume this is OK, but we should probably try to check it?
+             ;;
+       *)
+             # Punt and assume things are OK
+             ;;
+    esac
+    if test ${have_ftype} = yes; then
+        return 0
```
cgwalters: Need a kernel check here too, will do that after we've agreed on this approach.

rhatdan:
@cgwalters We are not planning on using this other then for Docker.  Not planning on using it for CRI-O.

cgwalters:
What *is* the plan for CRI-O?  Just `overlay` by default?

rhatdan:
Overlay by default and have admin configure the setup if he wants to modify it.  The main reason for docker-storage-setup originally was to setup docker to use devicemapper on loopback devices so it would work out of the box.   We did not have a nice default at that time like Overlay. We added some additional features, but when we looked at making this work for other tools like CRI-O, Buildah, kpod, skopeo things started to fall apart.    So @rhvgoyal and I decided in the end it was better to document how to setup alternative storage.

cgwalters:
One deep issue of course is that AH starts `docker.service` by default.  I regret that now...but we're somewhat locked into it.

rhatdan:
Yes I would love to remove all container runtimes from AH, and allow users to pick and choose which CR they want to use.  Customers I talked to, really want AH to be shrunk, where they are talking about the number of packages in the default install.

container-storage-setup.sh
```diff
@@ -1005,10 +1011,9 @@ is_valid_storage_driver() {
   # Empty driver is valid. That means user does not want us to setup any
   # storage.
   [ -z "$driver" ] && return 0
-
-  for d in $_STORAGE_DRIVERS;do
-    [ "$driver" == "$d" ] && return 0
-  done
+  # Must be an overlay variant or devicemapper
+  is_overlay_driver "$driver" && return 0
+  test "$driver" == "devicemapper" && return 0
```
rhvgoyal: I think there is no need to change this. Adding another driver does not change code. In fact older code is simpler and does not care what's in the list.

container-storage-setup.1
```diff
@@ -88,8 +88,11 @@ functionality to work properly.
 STORAGE_DRIVER:
       Specify a storage driver to be used with container runtime.
       Default: "devicemapper".
-      Valid values are overlay, overlay2 and "".
-      "" tells container-storage-setup to not perform any storage setup.
+      Valid values are overlay, try-overlay, overlay2 and "".  The
```
rhvgoyal: How about introducing a new option instead. Say `STORAGE_DRIVER_AUTO=1`. If this is set, then storage driver is determined automatically by the script. And it could be anything. I mean down the line we might want to configure a new driver say "xfs reflink" first and then fall back to overlay and then fall back to devicemapper. This option will continue to work even when we introduce new drivers.

rhvgoyal: STORAGE_DRIVER_AUTO basically means determine storage driver automatically without user having to specify a fixed value.

cgwalters: How about `s/try-overlay/auto/`?

rhvgoyal: STORAGE_DRIVER=auto sounds good.

container-storage-setup.sh
```diff
@@ -437,6 +437,46 @@ check_docker_storage_metadata() {
   exit 1
 }
 
+# See whether we can use overlay backend
+can_do_overlay() {
+    if test -d $_DOCKER_METADATA_DIR; then
+        mountroot=$_DOCKER_METADATA_DIR
+    else
+        mountroot=/var
```
rhvgoyal: This function can be useful both in compatibility mode as well as non-compatibility mode. So anything which ties it to compatibility mode should be outside the function. I think it is better to not check  `_DOCKER_METADATA_DIR` here. Instead caller should pass the mountroot argument to function.

container-storage-setup.sh
```diff
@@ -437,6 +437,46 @@ check_docker_storage_metadata() {
...
+    fstype=$(stat -f -c '%T' $mountroot)
+    have_ftype=yes
+    case "$fstype" in
+        # For xfs, see https://bugzilla.redhat.com/show_bug.cgi?id=1288162#c8
+        xfs) if test "$(xfs_info $mountroot | grep -o 'ftype=[01]')" = "ftype=0"; then
+                 have_ftype=no
+                 Info "XFS filesystem at ${mountroot} has ftype=0, cannot use overlay backend; consider separate volume (or OS reprovision)"
+             fi
+             ;;
+       ext4) # I'm just going to assume this is OK, but we should probably try to check it?
```
rhvgoyal: I think ext4 and others case can be merged into one. No need to special case ext4, as of now.

container-storage-setup.sh
rhvgoyal: I guess any kernel version check will have to be specific to distribution and distribution should carry that patch and not upstream project?

container-storage-setup.sh
```diff
@@ -437,6 +437,46 @@ check_docker_storage_metadata() {
...
+    esac
+    if test ${have_ftype} = yes; then
+        return 0
+    else
+        return 1
+    fi
+}
+
+# Call can_do_overlay() and set STORAGE_DRIVER
+consume_try_overlay() {
```
rhvgoyal: how about function name `choose_storage_driver()` instead.

rhvgoyal: Or `select_storage_driver()`

container-storage-setup.sh
```diff
@@ -1336,6 +1376,8 @@ setup_storage_compat() {
   if [ "$STORAGE_DRIVER" == "devicemapper" ]; then
     setup_lvm_thin_pool_compat
   else
+      # Handle the try-overlay case here, otherwise just write the config
+      consume_try_overlay
```
rhvgoyal: I think this call should come before the devicemapper check? First we need to select a storage driver and then get into the code which sets it up.

container-storage-setup.sh
```diff
@@ -2315,8 +2357,10 @@ setup_storage() {
 
   # Set up lvm thin pool LV.
   if [ "$STORAGE_DRIVER" == "devicemapper" ]; then
-    setup_lvm_thin_pool
-  elif is_overlay_driver "$STORAGE_DRIVER";then
+      setup_lvm_thin_pool
+  else
+      # Handle the try-overlay case here, otherwise just write the config
+      consume_try_overlay
```
rhvgoyal: This also should probably be called before the if block starts.

rhvgoyal:
So if STORAGE_DRIVER=auto, we can't allow users to specify any driver specific options. If user does not know what driver will be selected, how can they specific driver specific options. So we will have to put logic which checks if there are any driver specific options in /etc/sysconfig/docker-storage-setup and error out if that's the case.

container-storage-setup.sh
cgwalters: I did that in the `fixup!` - try reviewing via the full "Files changed" tab.

container-storage-setup.sh
cgwalters: Every other case though needs to distinguish between them, it seems simpler to have another `is_overlay_driver` call than to have a list only used by one caller right?

cgwalters:
> So if STORAGE_DRIVER=auto, we can't allow users to specify any driver specific options.

Or perhaps instead assume if the admin specifies devmapper-specific options, use that by default?

container-storage-setup.sh
rhvgoyal: we already have a list and if list traversal is easy, I really can't see why should not we do this in this case. In other cases using `is_overlay_driver` is just fine.

rhvgoyal:
Our current default is `devicemapper` as specificed in /usr/share/container-storage-setup/container-storage-setup`. I am assuming that now you are going to change this to `STORAGE_DRIVER=auto` instead. That is a user by default will get `STORAGE_DRIVER=auto`, until and unless user has overridden it.

I am not a big fan of assuming user wants devicemapper. What will we do if user has specified conflicting options. Say some for devicemapper and some for overlay.

cgwalters:
Yeah, though `auto` will (should) only take effect for new installs.  But what we do need to think about is the case where an admin is writing to the config via e.g. cloud-init/ansible/etc., and hasn't specified `STORAGE_DRIVER`.

I'd say we *have* to handle this case correctly; right now the default on CentOS AH for example is:

```
# Edit this file to override any configuration options specified in
# /usr/share/container-storage-setup/container-storage-setup.
#
# For more details refer to "man container-storage-setup"
```

If the user changes that to e.g.
`CONTAINER_THINPOOL=my-awesome-thinpool` and that's it, we need to auto-select `devicemapper` right?

As far as specifying options for both devmapper and overlay, I'd personally be fine if we just picked arbitrarily...

cgwalters:
OK, pushed a total rework of this ⬆️.  I haven't tested it, but see whether the high level looks better to you?

container-storage-setup.1
```diff
@@ -88,8 +88,13 @@ functionality to work properly.
 STORAGE_DRIVER:
       Specify a storage driver to be used with container runtime.
       Default: "devicemapper".
-      Valid values are overlay, overlay2 and "".
-      "" tells container-storage-setup to not perform any storage setup.
+      Valid values are devicemapper, overlay, auto, and "".  The empty string ""
+      is used when you want to ensure container-storage-setup does not automatically
+      change configuration (for example, you want to do it manually).  The "auto" option
+      may be used when c-s-s is started by default (for example, Atomic Host);
```
rhvgoyal: Can we specify that `auto` option means that css will determine storage driver automatically. choosing overlay2 first  is only current behavior and it can change down the line. Don't want users to rely on this.

container-storage-setup.sh
```diff
@@ -1002,13 +1046,9 @@ get_current_storage_options() {
...
 
-  # Empty driver is valid. That means user does not want us to setup any
-  # storage.
-  [ -z "$driver" ] && return 0
-
-  for d in $_STORAGE_DRIVERS;do
-    [ "$driver" == "$d" ] && return 0
-  done
+  case "$driver" in
+      ""|auto|devicemapper|overlay|overlay2) return 0;;
```
rhvgoyal: I reallly don't understand this. defining all valid storage drivers in a single global variable makes so much of sense. Rest of the code can make use of it.

Why to move it inside a function. What do we gain.

cgwalters: Before, the only user of that variable was `is_valid_storage_driver` - every *other* function needs to take action depending on the type by using `case` - so we're now more consistent.

container-storage-setup.sh
```diff
@@ -49,6 +49,7 @@ _STATUSFILE_NAME="status"
 # Read man page for a description of currently supported options:
 # 'man container-storage-setup'
 
+_DEFAULT_CONTAINER_THINPOOL=docker-pool
```
rhvgoyal: Not sure why do we need it. Is it just a cleanup/fix for some other issue.

cgwalters: See the use of it in `select_auto_storage_driver`

cgwalters:
One final step I didn't take in this PR is to change the default to `auto`.  That's obviously a small additional patch; I just wanted to separate the infrastructure from the default change.  But the goal is to make that switch, so review with that in mind.

container-storage-setup.sh
```diff
@@ -431,6 +431,50 @@ check_docker_storage_metadata() {
...
+}
+
+# If STORAGE_DRIVER=auto, pick a backend, and update STORAGE_DRIVER
+select_auto_storage_driver() {
+    local mountroot=$1
+    if [ "$STORAGE_DRIVER" == "auto" ]; then
+        # Is a CONTAINER_THINPOOL option set to a non-default value, or
+        # the "VG" option is set?
+        # Then choose "devicemapper".
+        if [ "${CONTAINER_THINPOOL:-}" != "${_DEFAULT_CONTAINER_THINPOOL}" ] ||
```
rhvgoyal: Oh using here.

Hmm.., so what happens if user specifies `CONTAINER_THINPOOL=docker-pool` in /etc/sysconfig/docker-storaege-setup. 

So how about this. 

Lets not change our STORAGE_DRIVER default and let that be devicemapper. And ask user to specify STORAGE_DRIVER=auto.

That way distributions can ship their default config file with STORAGE_DRIVER=auto. And this will take affect only on new installations. 

And in simplest form, we will ignore all driver specific options (instead of guessing like above). Down the line we could harden the code to error out.

cgwalters:
OK, pushed fixups ⬆️, and tested - things seem to work well for me, now we just need to change `STORAGE_DRIVER=auto` in our downstreams.

cgwalters:
https://src.fedoraproject.org/rpms/docker/pull-request/3

Makefile
```diff
@@ -19,7 +19,6 @@ install-docker:
 	install -D -m 644 docker-storage-setup.service ${UNITDIR}/${DOCKER}-storage-setup.service
 	if [ ! -f ${SYSCONFDIR}/${DOCKER}-storage-setup ]; then \
 		install -D -m 644 docker-storage-setup-override.conf ${SYSCONFDIR}/${DOCKER}-storage-setup; \
-		echo "CONTAINER_THINPOOL=docker-pool" >> ${SYSCONFDIR}/${DOCKER}-storage-setup; \
```
rhvgoyal: If we are not changing upstream default, there is no need to change this. right? In fact I think it probably is a good idea to also add `STORAGE_DRIVER=devicemapper` to make default explicit. Downstream should overwrite this file to change upstream default.

container-storage-setup.1
```diff
@@ -86,10 +86,16 @@ functionality to work properly.
...
+      Specify a storage driver to be used with container runtime; the default
+      value will chosen by your OS vendor.
+      Known values are: devicemapper, overlay, auto, and "".  The empty string ""
+      is used when you want to ensure container-storage-setup does not automatically
+      change configuration (for example, you want to do it manually).  The "auto" option
+      may be used when c-s-s is started by default (for example, Atomic Host);
+      when specified, c-s-s will choose a backend.  Currently, this will be "overlay" if
+      supported by the kernel/filesystem at the target mount point, and if not, then
+      "devicemapper" will be used.
+      Note the "overlay2" option is an alias for "overlay".
```
rhvgoyal: In docker, overlay2 and overlay are two different drivers (and overlay2 is not an alias for overlay).

container-storage-setup.conf
```diff
@@ -1,7 +1,7 @@
 # Specify storage driver one wants to use with container runtimes.
-# Default is devicemapper.
-# Other possible options are overlay, overlay2 and "". Empty string means do
-# not do any storage setup.
+# The default is `devicemapper`.
+# Other possible options are "auto", "overlay" as
+# well as the empty string "".  For more information, see `man container-storage-setup`.
```
rhvgoyal: overlay and overlay2 are two different drivers in docker. So not sure why should we remove overlay2 from list.

container-storage-setup.sh
```diff
@@ -49,6 +49,8 @@ _STATUSFILE_NAME="status"
 # Read man page for a description of currently supported options:
 # 'man container-storage-setup'
 
+# This variable used to be set by default in the config file
+_DEFAULT_CONTAINER_THINPOOL="docker-pool"
```
rhvgoyal: This is used only in compatibility mode with docker. So will make sense to call it `_DEFAULT_DOCKER_THINPOOL`

container-storage-setup.sh
```diff
@@ -2501,7 +2551,10 @@ done
 # Check subcommands
 case $# in
   0)
-    CONTAINER_THINPOOL=docker-pool
+    # A lot of code references this currently; it used to be
+    # written in the config file by default as well.  Set it to the default
+    # until we can rework the code to do "check if set, otherwise use default".
+    CONTAINER_THINPOOL=${_DEFAULT_CONTAINER_THINPOOL}
```
rhvgoyal: Current logic is that we start with default and then source /etc/sysconfig/docker-storage-setup and if user has specified a different name, that new name will take effect. So this comment is not accurate. May be its not needed at all.

rhvgoyal:
Right now I see 3 patches. With two patches with same subject. And there are no changelogs in top two patches. Can we get patches broken down properly and with proper changelogs. We have discussed a lot about our rationale of why we are doing certain thing. It will be good of changelogs are verbose.

cgwalters:
The latter two are `fixup!` - they get auto-squashed by homu on merge.

cgwalters:
The idea behind using `fixup!` is you can easily see what changed between last review, without littering the final merge with "fix comments" type patches.

container-storage-setup.1
cgwalters: Oh man...yes.  I had thought for some reason we were canonicalizing now but in fact you're right - that would have been a major screwup!

container-storage-setup.sh
cgwalters: My patch no longer writes it into the default config file, but the lvm path still relies on there being a default, right?

Makefile
rhvgoyal: Lets leave it untouched for now. I can't see how it is related to STORAGE_DRIVER=auto

cgwalters:
📞 ?

tests/016-test-auto-setup-cleanup.sh
```diff
@@ -0,0 +1,34 @@
+source $SRCDIR/libtest.sh
+
```
rhvgoyal: We will need a equivalent test for non docker/compatibility mode as well. Because this option is not specific to docker only. It should work with command line too.

rhvgoyal:
I was waiting for you to update patch. Did not realize you already updated it. I will test it tomorrow and look little more closely. In the mean time please add another test case for non-compatibility mode.

tests/016-test-auto-setup-cleanup.sh
cgwalters: Feels a bit weird to be setting up storage "directly"¹ but choose `STORAGE_DRIVER=auto` right?  It feels like it's fine if those admins have `STORAGE_DRIVER=overlay2` in their config files...in the (IMO fairly unlikely now) event that we add a new driver, that'd be a pretty major event; not something to just have happen automatically right?

¹ Calling this "non-compatibility mode" feels weird, how about "direct mode"?

rhvgoyal: Sure, we can call it `direct` mode.  If STORAGE_DRIVER=auto makes sense for docker, then it makes sense for direct mode also why not. Admin is asking tools to choose a driver based on what kernel is supporting. I mean how two cases are different.

rhvgoyal: Let me take that back. I think in current form STORAGE_DRIVER=auto might not make much sense for direct mode. And reason being that it primarily is doing ftype check. Which should be done on the filesystem which will be used as upper layer by container run time. And in direct mode, tool has no idea which filesystem/dir container run time is planning to use for upper dirs.

For compatibility mode it *sort* of works because we have hardcoded location of docker dir /var/lib/docker. That will be broken too if docker is started with option `-g`.

So lets atleast block being able to specify STORAGE_DRIVER=auto in direct mode for now.

container-storage-setup.1
```diff
@@ -86,10 +86,15 @@ functionality to work properly.
...
-      Valid values are overlay, overlay2 and "".
-      "" tells container-storage-setup to not perform any storage setup.
+      Specify a storage driver to be used with container runtime; the default
+      value will chosen by your OS vendor.
+      Known values are: devicemapper, overlay2, overlay, auto, and "".  The empty string ""
+      is used when you want to ensure container-storage-setup does not automatically
+      change configuration (for example, you want to do it manually).  The "auto" option
+      may be used when c-s-s is started by default (for example, Atomic Host);
+      when specified, c-s-s will choose a backend.  Currently, this will be "overlay2" if
+      supported by the kernel/filesystem at the target mount point, and if not, then
```
rhvgoyal: What is `target mount point`. I understand it. But normal users don't. How about this. We define another knob say `CONTAINER_METADATA_PATH`, which specifies the directory where container metadata will be stored. And STORAGE_DRIVER=auto will make ftype checks on filesystem which has `CONTAINER_METADATA_PATH`.

For compatibility mode, CONTAINER_METADATA_PATH can default to /var/lib/docker. For direct mode, user will have to specify it otherwise we fail. So for case of crio, user will specify `/var/lib/containers/storage` or something like that.

rhvgoyal: It also has the benefit that STORAGE_DRIVER=auto will be properly defined both for compatibility mode and direct mode.

container-storage-setup.sh
```diff
@@ -1321,6 +1353,15 @@ setup_storage_compat() {
    Fatal "Storage is already configured with ${current_driver} driver. Can't configure it with ${STORAGE_DRIVER} driver. To override, remove ${_STORAGE_OUT_FILE} and retry."
   fi
 
+  # First expand `auto` if in use
+  local mountroot
+  if test -d ${_DOCKER_METADATA_DIR}; then
```
rhvgoyal: In fact this test probably should be if we are in compatibility mode or not. If we are in compatibility mode, then default to /var/lib/docker otherwise default to /var.

For direct mode, may be we can default to /var for now. That will work for CRI-O as well. And once somebody needs to change the path, we can introduce knob CONTAINER_METADATA_PATH.

container-storage-setup.sh
```diff
@@ -2072,13 +2115,11 @@ list_config() {
...
 
-  if [ "$_M_STORAGE_DRIVER" == "" ]; then
-    return 0
-  elif [ "$_M_STORAGE_DRIVER" == "overlay" ]  || [ "$_M_STORAGE_DRIVER" == "overlay2" ];then
-    list_overlay_params
-  else
-    list_devicemapper_params
-  fi
+  case "$_M_STORAGE_DRIVER" in
+      ""|auto) return 0;;
```
rhvgoyal: I don't think we can keep _M_STORAGE_DRIVER=auto. Because we use this value to determine how to activate storage. If we store "auto", then nothing will be activated and direct mode will be broken.

In direct mode we will have to translate it to either overlay2 or devicemapper and save that value.

container-storage-setup.sh
```diff
@@ -2297,11 +2338,17 @@ setup_lvm_thin_pool () {
...
 
+# See also setup_storage_compat()
 setup_storage() {
   if ! is_valid_storage_driver $STORAGE_DRIVER;then
     Fatal "Invalid storage driver: ${STORAGE_DRIVER}."
   fi
 
+  # First expand `auto` if in use
+  if test "${STORAGE_DRIVER}" = "auto"; then
+      Fatal "Cannot use STORAGE_DRIVER=auto in non-compatibility mode"
```
rhvgoyal: Again, I think this should be allowed. We can check ftype by default on /var.

rhvgoyal:
And last, I am still concerned that we have not defined the semantics of what happens if user specifies STORAGE_DRIVER=auto, and there are other config options present in the file.

Current semantics seem to be that existing options will be used if option belongs to selected driver. That means a user will be to specify options belonging to all drivers because user does not know which driver will be selected. 

For example, Specifying CONTAINER_THINPOOL is mandatory now. So a config file will look.

STORAGE_DRIVER=auto
CONTAINER_THINPOOL=foo

And we might end up using "overlay2".

This looks pretty odd and hacky to me.

rhvgoyal:
We probably are making this thing more complicated, then it really has to be. Thinking purely from upstream point of view, how about if we switch to overlay2 as default graph driver. At run time we do bunch of checks to make sure overlay is supported.

- /var/ supports ftype=1 (for xfs)
- Create a temporary overlay2 mount point with multiple lower directories to make sure kernel supports multiple lower dirs.

And if that works, overlayfs will most likely work. And we don't have to introduce STORAGE_DRIVER=auto. 

Now possible issues with this are.

- After an upgrade, default storage driver will change to overlay2 (from devicemapper) and that can be a surprise to people. This will not break existing storage configurations but if people reset storage and restart docker, then they will be using overlay2 instead.  Is that a problem?

We could put some logic to detect if devicemapper specific options are present in /etc/sysconfig/docker-storage-setup, and error out saying default storage driver is overlay2 and CONTAINER_THINPOOL does not make sense and force user to make some changes.

- It will take affect on older kernels too. (Don't know what can we do about it).

But overall, this seems much simpler than going into the mechanics of supporting STORAGE_DRIVER=auto.

rhvgoyal:
@rhatdan WDYT?

rhvgoyal:
One problem with dynamic checks is it will require some temp directories to be created. May be we can write something in /tmp/ or /var/tmp/ and test if overlay mount is supported or not.

rhvgoyal:
That way upstream does not have to deal with kernel version checks. And if something does not work, then user will have to change storage driver. Well current default also does not guarantee that it works on all machines. It requires extra space and a volume group and many cloud images don't have that by default.

In fact overlay2 as default probably will work fine on many cloud images which don't have LVM and does not require extra disk/space out of box, which is kind of nice.

rhatdan:
Makes sense to me.  I would prefer that you do the checking for overlay on /run rather then /tmp, to prevent potential attacks from users.

cgwalters:
Hmm, maybe it actually would just work to change the default.  Existing installs will be fine.  The only case where it'll break is using old install media (e.g. RHEL 7.1 ISO) to kickstart a 7.4 tree.  But IMO those people will just have to deal with it.

So...OK, on one hand we need to get this right, on the other hand we don't have a ton of time to pick a choice and start testing/documenting it.  I can take a shot at rewriting this patch again, but if you have a strong opinion on how it's done, do you want to take it and I'll review?

rhvgoyal:
@rhatdan is /run writable for most the use cases? What about /run inside container?

cgwalters:
Obsoleted by https://github.com/projectatomic/container-storage-setup/pull/263

rhatdan:
/run would be writable when the process is run as root.  /run inside of a container would have the same chance of being read/only as /tmp or /var/tmp.

PR #266, opened 2018-04-04
Add support for GPT partition table for disks in size bigger than 2TB

rhvgoyal:
Right now we create MSDOS partition tables by default and that woks only till partition/disk size of 2TB. If somebody passes in a bigger disk, both sfdisk and parted fail and container storage setup fails. This patch series detects if disk is bigger than 2TB and if it is, then it switches to using gpt format partition table.

### DIFF
--- a/container-storage-setup.sh
+++ b/container-storage-setup.sh
@@ -38,6 +38,9 @@ _LOCKFILE="lock"
 _CONFIG_NAME=""
 _CONFIG_DIR="/var/lib/container-storage-setup/"
 
+# Partition type related
+_MAX_MBR_SIZE_BYTES="2199023255040"
+
 # Metadata related stuff
 _METADATA_VERSION=1
 _INFILE_NAME="infile"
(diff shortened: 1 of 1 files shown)


rhvgoyal:
cc @rhatdan @cgwalters

rhvgoyal:
@cgwalters Sure if Ignition takes over, that's not a problem. We should be able to write a test for this on loop device. Will do.

rhvgoyal:
@cgwalters @rhatdan I added the tests. PTAL.

rhatdan:
LGTM
Is this something we will need to modify the spec file to make sure that gpt is installed or is in the same package as the msdos version?

rhvgoyal:
Old sfdisk does not handle GPT. But old parted does. So we need to make sure "parted" is installed. 
I think we had a bug where we made docker/container-storage-setup requiring parted. Let me check and see if that change finally made it in or not.

rhvgoyal:
@rhatdan, In fedora and rhel container-storage-setup, I see it has "Requires: parted". So I think we should be good and this should not require any spec file changes.

cgwalters:
@rh-atomic-bot r+ c019f15

PR #269, opened 2018-05-02
Zero first 4MB of device being partitioned

rhvgoyal:
It is possible that disk user passed to us either does not have any
signatures in the beginning or signatures have been wiped but there are
valid lvm signatures present in the middle of disk. Especially at the
start of partition. So in the past there might have been partition on
disk and valid lvm signatures. But somebody deleted partition (and left
signatures in place) and passed the disk to a different VM. Now when this
new VM creates a partition (at same offset by default), then these
signatures immediately become visible to udev and lvm rules and they start
processing /dev/part1 as valid PV.

This is unintentional and it soon starts racing with container-storage-setup
logic of wiping partition signatures and creating new PV label etc. 

So before creating a partition, zero out first 4MB of disk. Hopefully this
will wipe out any PV label which is in first partition. This assumes that
first partition offset is with-in first 4MB.
 
### DIFF
--- a/container-storage-setup.sh
+++ b/container-storage-setup.sh
@@ -958,13 +958,27 @@ create_disk_partitions() {
   local devs="$1" part
 
   for dev in $devs; do
+    # wipefs /dev/disk does not wipe any lvm signatures which might be
+    # present on /dev/diskpart1. This signature will become visible to
+    # lvm udev rules and will kickstart volume creation as soon as partion
+    # is created and race with further partition commands like wipefs,
+    # pvcreate etc. So zero out first few MB of disk in an attempt to
+    # wipe any lvm signatures on first partition.
+    #
+    # By now we have ownership of disk and we have checked there are no
+    # signatures on disk or signatures have been wiped. Dont care about
+    # any signatures now on in the middle of disk.
+    Info "Writing zeros to first 4MB of device $dev"
+    if ! dd if=/dev/zero of=$dev bs=1M count=4; then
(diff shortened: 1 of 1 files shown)


rhvgoyal:
cc @rhatdan @cgwalters

rhatdan:
LGTM

rhatdan:
@fkluknav @lsm5 need new build of docker for RHEL to fix this issue.  No changes to docker required, just container-storage-setup.

container-storage-setup.sh
```diff
@@ -958,13 +958,27 @@ create_disk_partitions() {
...
+    # lvm udev rules and will kickstart volume creation as soon as partion
+    # is created and race with further partition commands like wipefs,
+    # pvcreate etc. So zero out first few MB of disk in an attempt to
+    # wipe any lvm signatures on first partition.
+    #
+    # By now we have ownership of disk and we have checked there are no
+    # signatures on disk or signatures have been wiped. Dont care about
+    # any signatures now on in the middle of disk.
+    Info "Writing zeros to first 4MB of device $dev"
+    if ! dd if=/dev/zero of=$dev bs=1M count=4; then
```
fche: (Is there a possible quoting problem with $dev?  Are we guaranteed only ever to have vanilla space-non-containing names?)

rhvgoyal: can device path container space? I assumed it can't and that's why did not bother to put double quotes.

PR #271, opened 2018-05-22
Cope with atomic upgrade overwriting /etc/sysconfig/docker-storage-setup

rhvgoyal:
Our current default driver is "devicemapper" and recently it was changed to "overlay2". Normally it works as over upgrades rpms don't overwrite /etc/sysconfig/docker-storage-setup. (It is not a config(noreplace) file). 

But atomic does not understand it and overwrites this file over upgrade. That means we are left with a scenario where a devicemapper thin pool is already configured but /etc/sysconfig/docker-storage-setup
has STORAGE_DRIVER=overlay2.

This is a corner case. Deal with it. During start, bring up thin pool if one is already configured. And during reset, continue to reset if there is a configured thin pool.

### DIFF
--- a/container-storage-setup.sh
+++ b/container-storage-setup.sh
@@ -179,6 +179,16 @@ extra_options_has_dm_fs() {
   return 1
 }
 
+# Given a dm device name in /dev/mapper/ dir
+# (ex. /dev/mapper/docker-vg--docker-pool), get associated volume group
+get_dmdev_vg() {
+  local dmdev=${1##"/dev/mapper/"}
+  local vg
+
+  vg=`dmsetup splitname $dmdev --noheadings | cut -d ":" -f1`
+  echo $vg
+}
+
 # Wait for a device for certain time interval. If device is found 0 is
 # returned otherwise 1.
 wait_for_dev() {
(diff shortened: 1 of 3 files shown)


rhvgoyal:
cc @cgwalters @dustymabe @rhatdan

cgwalters:
The unhappiness with how ostree handles this in the commit messages aside, this all looks good to me.

(Remember, terminology gets tricky here because today you can do e.g. `rpm-ostree override replace ./container-storage-setup.x86_64.rpm` - and that *still* has the same semantics around `%posttrans`.  It's RPM - but handled in a new way.  Among other advantages over `yum` for example, it's still fully transactional.  Hence, I tend to use the term "classic" or "yum-managed" systems, and "ostree"/rpm-ostree (or atomic) for the latter.)

rhatdan:
LGTM. but this is a lot of changes.

rhvgoyal:
@cgwalters I don't mind new way of doing things as long as it is well published what one can do. Problem is that we are used to do things rpm way and suddenly come to know that what we can do in rpm, is not compatible with rpm-ostree stuff. And suddenly we are hacking to code to work it around after users have reported issues.

To me this all should have been analyzed and addressed when we were switching driver to overlay2. I am not sure why did we let it go and did not think more about it.

PR #275, opened 2018-06-26
c-s-s: Support GROWPART=true on NVMe

### LINKED ISSUE #133: Fails to extend disk on AWS with NVMe
We dropped cloud-init which has disk resizing code, and it turns out Ignition doesn't do this, see https://github.com/coreos/init/blob/master/scripts/extend-filesystems

Now the next problem we have is that script doesn't handle LVM...in practice what we did with Fedora Atomic was that it was actually [container-storage-setup](https://github.com/projectatomic/container-storage-setup) that does this, I need to figure out why that's not working.

cgwalters:
By using sysfs, splitting device/partition is a lot more reliable.
This fixes using GROWPART=true on NVMe with Fedora Atomic Host,
as occurs in AWS with modern instance types.

Closes: https://github.com/openshift/os/issues/133

### DIFF
--- a/container-storage-setup.sh
+++ b/container-storage-setup.sh
@@ -727,8 +727,20 @@ grow_root_pvs() {
   # partitions on all disks, so as long as they match, growing the LV should
   # also work.
   for pv in $_ROOT_PVS; do
-    # Split device & partition.  Ick.
-    growpart $( echo $pv | sed -r 's/([^0-9]*)([0-9]+)/\1 \2/' ) || true
+    if ! test -b $pv; then
+        Error "Not a block device: $pv"
+    fi
+    local major_hex minor_hex major_minor
+    local devpath partition parent_path parent_device
+    major_hex=$(stat -c '%t' $pv)
+    minor_hex=$(stat -c '%T' $pv)
+    major_minor=$((0x${major_hex})):$((0x${minor_hex}))
+    devpath=$(realpath /sys/dev/block/$major_minor)
+    partition=$(cat $devpath/partition)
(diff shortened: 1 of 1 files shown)


container-storage-setup.sh
```diff
@@ -727,8 +727,18 @@ grow_root_pvs() {
...
-    # Split device & partition.  Ick.
-    growpart $( echo $pv | sed -r 's/([^0-9]*)([0-9]+)/\1 \2/' ) || true
+    if ! test -b $pv; then
+        Error "Not a block device: $pv"
+    fi
+    local major_hex=$(stat -c '%t' $pv)
+    local minor_hex=$(stat -c '%T' $pv)
+    local major_minor=$((0x${major_hex})):$((0x${minor_hex}))
+    local devpath=$(realpath /sys/dev/block/$major_minor)
+    local partition=$(cat $devpath/partition)
```
dm0-: Should the script abort if any of these commands fail, such as this one if the device is not a partition?  The `set -e` won't catch this since bash interprets the `local` statement as a successful simple command, so the exit status of the command substitution is ignored.

ashcrow:
\***bump**\*

dm0- (approved):
/lgtm

cgwalters:
(This repo is [papr](https://github.com/projectatomic/papr) and not prow, it's confusing)

@rh-atomic-bot r=dm0- d4994e6

cgwalters:
We should do a release+update with this: https://discussion.fedoraproject.org/t/partitions-not-expanding-on-nvme/399

PR #283, opened 2019-06-28
Fixes #282 use requiredby to make a hard dependency chain

### LINKED ISSUE #282: Failed mount does not lead to startup error
Hi,
I think the services created by docker-storage-setup need to enforce successful startup. In my case the /var/lib/docker mount had gone into a bad state ("mount: wrong fs type, bad option, bad superblock") but that did not keep docker from starting and using my /var volume directly.

I think the unit files need to be changed like this:
https://github.com/projectatomic/container-storage-setup/blob/master/container-storage-setup.sh#L1330
needs to be RequiredBy=

And https://github.com/projectatomic/container-storage-setup/blob/master/docker-storage-setup.service#L12
needs to also have RequiredBy=docker.service

That way an admin is sure to notice a problem with docker storage.

I'll create a PR.

Greetings
Klaas

Klaas-:
Fixes #282 use requiredby to make a hard dependency chain

### DIFF
--- a/container-storage-setup.sh
+++ b/container-storage-setup.sh
@@ -1327,7 +1327,7 @@ Type=xfs
 Options=pquota
 
 [Install]
-WantedBy=docker-storage-setup.service
+RequiredBy=docker-storage-setup.service
 EOF
   mv "${unit_file_path}.tmp" "$unit_file_path"
   systemctl daemon-reload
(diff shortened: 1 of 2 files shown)


rhatdan:
@rhvgoyal PTAL
As I recall we wanted to allow this to fail in order to allow Docker running with Devicemapper on loopback devices to work.  I think containers-storage-setup would fail in that situation and cause Docker to use loopback devices.  Now that we default to using Overlay, we could perhaps make the switch.

Klaas-:
So there is a situation where you expect the service to fail? Wouldn't it be better/easier to just not use the service ie when dm on loopback "systemctl disable docker-storage-setup"? or even fully remove the package and only install it when needed?

